PR #4, opened 2015-10-29
Remove route-breaking hrefs

jking90:
Leaving `href="#"` on the `<a>`s consistently redirected the user to the route `/`. I've removed them so that routing is no longer affected by clicking around on the calendar.

### DIFF
--- a/src/EventEndingRow.jsx
+++ b/src/EventEndingRow.jsx
@@ -85,7 +85,6 @@ let EventRow = React.createClass({
       ? (
         <a
           key={'sm_' + slot}
-          href='#'
           className={'rbc-show-more'}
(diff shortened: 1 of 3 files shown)


jquense:
Hey there, 

Thanks for the PR, I think I actually fixed this already [using preventDefault](https://github.com/jquense/react-big-calendar/blob/master/src/EventEndingRow.jsx#L98) just yesterday, can you confirm that the newest version resolves the issue?

jking90:
I was having the issue today when clicking on the the date numbers in the month view, using the latest published version.

jquense:
ahh sorry about that. fixed in https://github.com/jquense/react-big-calendar/commit/69b092d

PR #10, opened 2015-11-20
The Taskbar button labels can now be changed.

rafsoaken:
Use the 'messages' props for the <Calendar /> component to pass the labels
as described in the PropTypes of <Calendar />


jquense:
thanks!

PR #25, opened 2016-01-08
properties types fixed

Sigura:
if messages is 
```javascript
<BigCalendar
  events={events}
  messages={{
    previous: <i className="fa fa-chevron-left" />,
    next: <i className="fa fa-chevron-right" />
  }}
/>
```
then warning:
`Warning: Failed propType: Invalid prop 'messages.previous' of type 'object' supplied to 'Calendar', expected 'string'. Check the render method of 'Uncontrolled(Calendar)'.`

There is fix


antoinerousseau:
why did you commit the `lib/` folder?!

antoinerousseau:
Also, you can use `PropTypes.node` which includes `string` and `element`

https://facebook.github.io/react/docs/reusable-components.html

> `React.PropTypes.node`:
> Anything that can be rendered: numbers, strings, elements or an array (or fragment) containing these types.

jquense:
gonna close since it includes a bunch of noise from the lib folder. will take a PR that just adjsuts the propType to `node` though if anyone wants

antoinerousseau:
You already did in #40 ;-)

jquense:
lol

PR #40, opened 2016-02-08
Use the node type for messages instead of string

dantman:
This already works, no code change is needed; this just fixes the warning you get by doing so.

This turns out to be very useful for formatting button text or replacing it with an icon.


jquense:
thanks

PR #42, opened 2016-02-16
Don't treat with-time events that extend into the morning as all-day events

dantman:
Events that went from say 8PM to 2AM the next day were being listed as all-day events.

This change changes them to this formatting.
screen shot 2016-02-16 at 12 15 47 pmscreen shot 2016-02-16 at 12 16 02 pm

Normal all day events (that start and end on date-only dates or have the allDayAccessor set) still display as all day events. Though as a side effect events which span multiple days and have a defined start or end time will be formatted like this, instead of the middle portion that spans the whole day being put in the all day section. This is because the code would need to be significantly rewritten to put just the all-day portions in the all-day header.

screen shot 2016-02-16 at 12 16 36 pmscreen shot 2016-02-16 at 12 16 20 pm

### DIFF
--- a/src/DaySlot.jsx
+++ b/src/DaySlot.jsx
@@ -170,7 +170,7 @@ let DaySlot = React.createClass({
         <div
           key={'evt_' + idx}
           style={{...xStyle, ...style}}
-          title={(typeof label ? label + ': ' : '') + title }
+          title={(typeof label === 'string' ? label + ': ' : '') + title }
           onClick={this._select.bind(null, event)}
           className={cn('rbc-event', className, {
             'rbc-selected': _isSelected,
--- a/src/DaySlot.jsx
+++ b/src/DaySlot.jsx
@@ -119,9 +119,9 @@ let DaySlot = React.createClass({
 
   renderEvents(numSlots, totalMin) {
     let {
-        events, step, min, culture, eventPropGetter
-      , selected, eventTimeRangeFormat, eventComponent
-      , startAccessor, endAccessor, titleAccessor } = this.props;
+        events, step, min, max, culture, eventPropGetter, selected, messages
+      , eventTimeRangeFormat, eventTimeRangeStartFormat, eventTimeRangeEndFormat
+      , eventComponent, startAccessor, endAccessor, titleAccessor } = this.props;
 
     let EventComponent = eventComponent
       , lastLeftOffset = 0;
@@ -129,8 +129,22 @@ let DaySlot = React.createClass({
     events.sort((a, b) => +get(a, startAccessor) - +get(b, startAccessor))
 
     return events.map((event, idx) => {
+      let _eventTimeRangeFormat = eventTimeRangeFormat;
+      let _continuesPrior = false;
+      let _continuesAfter = false;
       let start = get(event, startAccessor)
       let end = get(event, endAccessor)
+      if ( start < min ) {
+        start = min;
+        _continuesPrior = true;
+        _eventTimeRangeFormat = eventTimeRangeEndFormat;
+      }
+      if ( end > max ) {
+        end = max;
+        _continuesAfter = true;
+        _eventTimeRangeFormat = eventTimeRangeStartFormat;
+      }
+
       let startSlot = positionFromDate(start, min, step);
       let endSlot = positionFromDate(end, min, step);
 
@@ -140,7 +154,13 @@ let DaySlot = React.createClass({
       let style = this._slotStyle(startSlot, endSlot, lastLeftOffset)
 
       let title = get(event, titleAccessor)
-      let label = localizer.format({ start, end }, eventTimeRangeFormat, culture);
+      let label;
+      if ( _continuesPrior && _continuesAfter ) {
+        label = messages.allDay;
+      } else {
+        label = localizer.format({ start, end }, _eventTimeRangeFormat, culture);
+      }
+
       let _isSelected = isSelected(event, selected);
 
       if (eventPropGetter)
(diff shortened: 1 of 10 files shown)


jquense:
Hey thanks a lot for this, it looks really cool!  Hopefully I can dog into it and give it a review tomorrow!

src/DaySlot.jsx
```diff
@@ -150,11 +170,13 @@ let DaySlot = React.createClass({
         <div
           key={'evt_' + idx}
           style={{...xStyle, ...style}}
-          title={label + ': ' + title }
+          title={(typeof label ? label + ': ' : '') + title }
```
jquense: whats going on with this `typeof` here? Also is there a way in which this could not exist?

src/localizers/globalize.js
```diff
@@ -10,6 +10,12 @@ let timeRangeFormat = ({ start, end }, culture, local)=>
   local.format(start, { time: 'short' }, culture) +
     ' — ' + local.format(end, { time: 'short' }, culture)
 
+let timeRangeStartFormat = ({ start, end }, culture, local)=>
+  local.format(start, 't', culture) + ' — '
```
jquense: I would mirror what the complete format above does in this case: `{ time: 'short' }`

src/localizers/globalize.js
```diff
@@ -10,6 +10,12 @@ let timeRangeFormat = ({ start, end }, culture, local)=>
...
+
+let timeRangeEndFormat = ({ start, end }, culture, local)=>
+  ' — ' + local.format(end, 't', culture)
```
jquense: ditto here

src/DaySlot.jsx
dantman: Remember my change that let messages be React components.

In the rare event that messages.allDay has been customized with a Component as a message and one of these rare multi-day but not all-day events exists, then label may end up a React component instead of a string.
screen shot 2016-02-16 at 12 16 20 pm

It's a rare situation and the textual label doesn't seem that important so I just decided to drop it in that rare situation.

jquense: ok that make sense, however you have no actual `typeof` check here, it should be `typeof label === 'string` right?

jquense:
So this is great, and thanks for taking the time to do it. I do think we need to avoid that case you talk about, with multiday events spanning the entire daySlot. Admittedly those aren't the must common but when they do occur they are going to completely wash out the rest of the events for that day/week.

I think there is a relatively surgical way to handle this case without to much rewriting. If [eventSegments()](https://github.com/intljusticemission/react-big-calendar/blob/master/src/utils/eventLevels.js#L4) took an additional argument `snapDate` function that is handed the start and end dates of the event you could return the actual dates you want.

So in the case of the TimeGrid EventRow's snapping the events to the nearest ceiling/floor of the dy would size the event to only the full days. That approach just allow you to change the behavior for the TimeGrid but leave it alone for the Month grid.

src/DaySlot.jsx
dantman: Whoops, you're right.

dantman:
@jquense I suppose, though first for comparison, here's what it would look like if we changed the code so that the all-day portion could be displayed in the all-day event list.

screen shot 2016-02-17 at 9 40 08 amscreen shot 2016-02-17 at 9 40 22 am

The long event may wash out other events. But I was also a little worried that it would be disruptive, switching between the grid and the all-day row, leaving an entire day's time grid empty event though an event spans from the previous day's time grid to the next day's time grid.

jquense:
Thanks for the mock-up. I agree with you on some of those concerns, but I think I want to lean towards the visualization that is clearest for busy calendars, and not optimize sparse one. In our own cases we often have ltos of teams coordinating many events mixed with events that run a week, or two. Having one (or two) of those sorts of long running events amid say 20 other events for that two weeks, starts to monopolize a lot of space, that otherwise doesn't add to understanding that those long events are long.

Perhaps another alternative is to just consider events that span more than an entire day as "all day" and only use the above for events that run late into the night? I am not quite sure which option is better, but I feel fairly certain that the current way isn't going to work great for us.

dantman:
> Perhaps another alternative is to just consider events that span more than an entire day as "all day" and only use the above for events that run late into the night?

Note that it's already possible for the user to do this themselves. Just tweak the allDayAccessor so it returns true, the start/end accessors so they remove the time/snap to the start/end of the dates, or tweak either part of original data; and the multi-day event will go into the all day row instead of being a timed event spanning multiple days in the time grid.

jquense:
> Note that it's already possible for the user to do this themselves. 

Not without messing up the month view behavior.

dantman:
> Not without messing up the month view behavior.

How? The month view is the same for an all day event and a multi-day event with times that start and end mid-day on the same days.

dantman:
Here's the month view for that "Multi-day Event" I added to the examples that spans `new Date(2015, 3, 20, 19, 30, 0)` to `new Date(2015, 3, 22, 2, 0, 0)`.

screen shot 2016-02-17 at 12 36 56 pm

Here's the month and week view for the same event after I modify `events.js` so that the event instead spans `new Date(2015, 3, 20)` to `new Date(2015, 3, 23)`.
screen shot 2016-02-17 at 12 37 57 pm
screen shot 2016-02-17 at 12 38 05 pm

Event becomes all day since it no longer starts or ends on a defined time and the month view stays the same since the days the event are part of doesn't change.

jquense:
You example doesn't touch the start/end accessors, it also doesn't create a situation where the start and end slices of the event show up in the time slots where the middle full day segments show up in the all day row. Using to the start/end accessors to snap events to the nearest day will shorten some events, making them appear like they start a day later. 

Using the allDayAccessor does get you closer, but it would change the behavior of the sorting and layout algo which takes into account starting times, snapping everything to a day would change the display logic of event.


More to the point tho I think even if can get what you want by messing with the settings I still think that the wrong default is being codified in the API for these situations. I'd prefer that the inverse be true. by default long events don't clog up the grid. I think that I am leaning towards my first suggestion as the normal behavior, but insofar as it doesn't add a lot more complexity I'm open to it being configurable, since you seem to prefer the current behavior.

6pm:
Please add example how use it property. Tnx.

PR #54, opened 2016-03-03
Do not show view toolbar if only one view is specified

jrm2k6:
Hi!

I was wondering if it would make sense not to display the views on the right side, in the toolbar if you pass only one view. Let's say you only want a monthly view, it doesn't really male sense to show only a month tab, not doing anything when clicking on it, and it doesn't show any valuable information.


jquense:
good call, thanks for the PR.

Issue #62, opened 2016-03-24
Events from 27 to 31 of march


jesty:
Hi, I have an issue with events at the end of march, but not of all years and only in "month" view.
Starting from the input below, I had a wrong result where I put a "-" sign, and a good result where I put a "+" sign:

\- "start":new Date(2011, 2, 28),"end":new Date(2011, 2, 28)
\- "start":new Date(2012, 2, 28),"end":new Date(2012, 2, 28)
\+ "start":new Date(2013, 2, 28),"end":new Date(2013, 2, 28)
\+ "start":new Date(2014, 2, 28),"end":new Date(2014, 2, 28)
\+ "start":new Date(2015, 2, 28),"end":new Date(2015, 2, 28)
\- "start":new Date(2016, 2, 28),"end":new Date(2016, 2, 28)
\- "start":new Date(2017, 2, 28),"end":new Date(2017, 2, 28)
\- "start":new Date(2018, 2, 28),"end":new Date(2018, 2, 28)
\+ "start":new Date(2019, 2, 28),"end":new Date(2019, 2, 28)
\+ "start":new Date(2020, 2, 28),"end":new Date(2020, 2, 28)
\+ "start":new Date(2021, 2, 28),"end":new Date(2021, 2, 28)

What happens?

Thanks in advance,
Davide

jesty:
Thanks to solve this issue so fast, I tried the fix on my locale machine.
When did you think to release this fix officially on NPM?
thanks, Davide

PR #65, opened 2016-03-27
Fixed Month and Week Rendering Bug

### LINKED ISSUE #36: Rendering bug in week and month views
Hi,

The row segment are rendered too short when I create an event spanning over an entire month:

2016-02-04 14_28_29-renderingbug

This behavior is also reproduced in the *week* view.

I have tried to look up the code for the possible error but I could not find it. 

I produced this bug with version 0.9.8

I use `react` 0.14.7

My browser is chrome version 48.0.2564.97

If you don't have time to look it up, I take any pointer to help out :)

Thank you for this superb react component!

KerenChandran:
Fixes #36 

@jquense I noticed in dade2b9 you added the slots variable and in 8ad4ee7 you modified the span variable. Upon debugging the issue, I noticed that the span was getting the correct calculation (which is the beginning of the start date to the end of the end date), but slots was retrieving 6. Can you please describe exactly what those commits were trying to fix? I really like your component and we are currently using it in a project so I would be glad to help you solve this issue!

### DIFF
--- a/src/utils/eventLevels.js
+++ b/src/utils/eventLevels.js
@@ -8,7 +8,7 @@ export function eventSegments(event, first, last, { startAccessor, endAccessor, 
 
   let span = dates.diff(start, end, 'day');
 
-  span = Math.floor(Math.max(Math.min(span, slots), 1));
+  span = Math.floor(Math.max(span, 1));
 
   let padding = Math.floor(dates.diff(first, start, 'day'));
 


jquense:
I'd have to poke around more to remember why it was changed, but I'm sure it fixed some other layout bug like this one, but who can say what exactly... I would be apprehensive to revert something in that space. 

can you break down specifically what is happening and how this addresses it? to my mind slots _should_ 7 I think?

jquense:
if I am remembering correctly non-even dates sometimes created spans greater than 7 (days in the week) so the slots check is supposed to provide an outer boundary on possible event lengths for a row, ie it should never span more than 7 days for a week

KerenChandran:
Ya I'm happy to explain how I got to this conclusion.

Slots variable calculates the difference between the first day of the week (Sunday midnight) and the end of the last day of the week (Saturday Midnight). This means that the difference between the days is always  6. 

On the other hand, span does the calculation correctly for the end date as it takes whatever was passed in and adds an extra day, making it go from Sunday midnight to next Sunday Midnight.

I removed slots from that line cause you were taking the min between span and slots and then comparing it to 1 (which I believe is used in case if the event is for a few hours on a day and you default it to 1 so it shows up on the calendar). I didn't want to modify or remove the variable completely from the code as I wasn't too sure what it was originally used for and I thought it might be used somewhere else (although thinking about it, its defined in the scope of that function so it might be safe to remove it)

KerenChandran:
Also I would highly recommend referencing your PRs to the issues or atleast provide a good description as to what it fixes and why its implemented. That way when someone does a log on the file, its to see why the commit was made and provides more details about stuff that should be tested.

I'll try to log the file again and look through the past commits, so I can help you verify if this is a good fix, or if it will introduce more bugs.

jquense:
So I think the issue here is that we are getting `6` instead of the desired `7` for the max slot count, and greatest possible span. I just want to make sure we don't reintroduce that bug...but it looks like slots is unnecessary here since `end` is constrained by `last + 1` day

KerenChandran:
Hm... Don't merge the PR as of now. I noticed that if I changed Long Event to run for a full year, it suddenly cuts on march 19 >_>. This may also be related to #62 and have to do with other parts of the code :confused: I'll try to look into it hopefully this week.

antoinerousseau:
fyi this fix does not fix https://github.com/intljusticemission/react-big-calendar/issues/36#issuecomment-204806210

jquense:
So there was two bugs here, inconsistent range math and a DST issue, I ended up fixing both while exploring it. thanks for the contribution it really helped narrow down the issues!

Issue #66, opened 2016-03-27
End Date


KerenChandran:
@jquense So I noticed that the end date specified isn't actually included in the span calculation. E.g. Saturday April 2, 2016 to Saturday April 9, 2016, the event spans from April 2, 2016 to April 8, 2016. Is this intended behaviour?

jquense:
the end is included, but I assume you mean that the date doesn't show up on the 9th if it ends on the 9th midnight? tbh I can't decide if this behavior makes more sense then the alternative. my initial thought is this is a bug, since if it was one minute over midnight it would show. I believe the issue is that the ceil() date function doesn't round up in the case where the date is also the floor

jquense:
On further thought I think the current behavior is correct, thinking about what the All Day rows in Week and Day view would look like for those dates, I think not having them count towards taht day, allows for easier ranges, otherwise you'd have to end dates that are just shy of midnight, or sort of odd behavior for zero width date ranges.

KerenChandran:
Hm... okay so just to confirm, you are saying that in the example code, Long Date runs from 7 to 10 and since there's no time stamp and due to the ceil function, it takes the date before instead of the current date. 

I find it weird to have it end on the 9th if the user wants it to go to the 10th, but I guess on our end we can add an inclusive/exclusive so when the event is made they can check it to say its not inclusive. Shall I close this issue then since we can conclude its not a bug?

jquense:
When dealing with dates, there is just an ugly bucket of trade-offs and weighting of use cases, since, well dates are hard.

> [...]it takes the date before instead of the current date

Its not so much taking the day before as it is not quite considering the exact moment the day flips as part of the range. if you were to ask "how many days does Apr 8th midnight to Apr 9th midnight span?" I would concede that "two" is an acceptable answer from a certain perspective, but that most people, i.e. the "common sense" answer is "one" and so we'd show that event as spanning one day.

I admit it gets even fuzzier when talking about the "8th to 9th 12:01am" but in that case the range actually extends into the second day. In the case of say "Apr 8th 12:00 - Apr 8th 12:00" I'd like to say that that event span no day vs 1 day. i.e it has _no_ duration. 

Ultimately I think either interpretation is "right" but the current one makes the most sense over a wider set of uses and examples. Plus you have to draw the line on some side and as I said I think just ding apr8th - apr9th is easier than apr8th -apr8th 23:59.999, give the lack of robust API on JS date objects.

[...]

PR #67, opened 2016-04-02
Fix multiday events rendering bug

### LINKED ISSUE #36: Rendering bug in week and month views

antoinerousseau:
Fixes #36.

Inspired from https://github.com/SlavaKvak/react-big-calendar/commit/a359db3736b618ef4ed4e26ba4d59c50249b7d84

The main change is that `slots` and `span` are now rounded, the same way `padding` was already rounded. No `Math.floor` anymore.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,9 +1,9 @@
 {
   "name": "react-big-calendar",
-  "version": "0.9.8",
-  "description": "Calendar! with events",
+  "version": "0.9.9",
+  "description": "Google Calendar like React component, inspired by fullcalendar.io",
   "author": "Jason Quense <monastic.panic@gmail.com>",
-  "repository": "jquense/react-big-calendar",
+  "repository": "intljusticemission/react-big-calendar",
   "license": "MIT",
   "main": "lib/index.js",
   "files": [
@@ -25,7 +25,7 @@
     "clean:examples": "rimraf examples/static",
     "less": "lessc -x src/less/styles.less ./lib/css/react-big-calendar.css",
     "assets": "cpy src/less/* lib/less",
-    "build": "npm run clean && babel src --out-dir lib && npm run assets & npm run less",
+    "build": "npm run clean && babel src --out-dir lib && npm run assets && npm run less",
     "build:examples": "npm run clean:examples && webpack --config webpack/docs.config.es6.js",
     "build:visual-test": "webpack --config webpack/visual-test.es6.js",
     "examples": "npm run clean:examples && babel-node ./examples/server.js",
--- a/README.md
+++ b/README.md
@@ -1,12 +1,13 @@
 react-big-calendar
 ========================
 
-An event Calendar component built for React.
+An events calendar component built for React.
 
-[__DEMO and Docs__](http://intljusticemission.github.io/react-big-calendar/examples/index.html).
-big calendar is built for modern browsers (read: ie10+) and uses flexbox over the classic tables-ception approach.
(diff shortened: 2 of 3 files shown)


antoinerousseau:
I replaced the bug with another: the event is now properly split accross two rows in both `en` and `fr` cultures, but its length is wrong in the `en` culture: it is one day too short.

Test event starts on `2016-04-01 12:00:00` and ends on `2016-04-04 12:00:00`

capture d ecran 2016-04-02 15 56 39
capture d ecran 2016-04-02 15 56 54

antoinerousseau:
And @okcompute king of bug is still present:

capture d ecran 2016-04-02 16 04 57

antoinerousseau:
After some debugging, it seems that the problem comes from the `slots` variable:

capture d ecran 2016-04-02 16 30 02

antoinerousseau:
OK, all good now in both cultures, this fixes both bugs :)

@jquense can you take a look please?

It seems to also resolve the interrogations raised by @KerenChandran in #66 since the end date is included.

antoinerousseau:
The main change is that `slots` and `span` are now rounded, the same way `padding` was already rounded. No `Math.floor` anymore.

jquense:
hi there thanks for the the contribution, I will take a closer look when I get a moment but a cursory glance has me I bit confused by what is going on and why the changes were made. In particular I don't think that math round is the right thing here... could you explain why you are using it? in general "because it made these dates correct" isn't great since there are A LOT of possibilities on a calendar and we need to not break other cases accidentally

antoinerousseau:
Well, I see that the date diff calculation for `slots` (number of cells in a row) and `span` (number of cells the event is spanning on the current row) are not giving integers, e.g. `6.95833..` instead of `7`, and I don't see why, since the `diff` method is called with `day` as the third argument. So it could be understandable for `span` if the start and end times are different, but what about `slots`? `first` and `last` always have the same time, right, since we are talking about entire days?

jquense:
As I was diving into the PR here I ended up changing stuff and it ultimately was easier just to commit may changes. Thanks again for the contribution it helped!

antoinerousseau:
Glad I could help! Thanks for the fix :)

You are also welcome to take some of the updates I made to the `README`, and also please add that missing `&` to the npm `build` script

And please publish on NPM :)

package.json
```diff
@@ -25,7 +25,7 @@
     "clean:examples": "rimraf examples/static",
     "less": "lessc -x src/less/styles.less ./lib/css/react-big-calendar.css",
     "assets": "cpy src/less/* lib/less",
-    "build": "npm run clean && babel src --out-dir lib && npm run assets & npm run less",
+    "build": "npm run clean && babel src --out-dir lib && npm run assets && npm run less",
```
jquense: this isn't missing an `&` and single `&` is how you run parallel tasks on windows, I assume not so on Mac/*nix?

antoinerousseau: single `&` is also for parallel tasks on OSX/unix, but I don't think it's a good idea to run these tasks in parallel because if you wait for an exit code from the command, you'll get it before it completes, in addition to have loose output consistency (you don't know when exactly it finishes)...

PR #73, opened 2016-04-04
Update dependencies

antoinerousseau:
Update babel & eslint (fixes issues in Atom)


package.json
```diff
@@ -41,10 +41,10 @@
   },
   "devDependencies": {
     "autoprefixer-core": "^5.2.1",
-    "babel": "^5.8.34",
-    "babel-core": "^5.8.34",
-    "babel-eslint": "^3.1.20",
-    "babel-loader": "^5.3.0",
+    "babel": "^6.5.2",
```
jquense: Just upgrading to Babel6 doesn't work, the babelrc file would need to be upgraded, and presets added, specific plugins picked, etc,etc. I don't want to deal with upgrading that right now, especially since there are a bunch of dev deps taht require babel 5 in order to build the docs

Issue #78, opened 2016-04-10
Calendar first day of week changed to Saturday without provocation


harel:
**NOTE** : I've discoverd the reason but leaving the ticket as is - see both edits at the bottom - I do think its a bug in the Calendar code. 

My Calendar is loaded with an "en-GB" locale. Globalize is used as the localizer. Everything was working fine until today I noticed the calendar has switched from starting the week on Monday as it should for this locale, to starting it on Saturday (as no locale I can think of does). 
I've been trying to figure out what has changed (it was fine yesterday). I've tried upgrading Globalize/CLDR to the latest versions, as well as ensuring the "en-GB" culture is provided and changing my localization of the calendar from 

```
BigCalendar.globalizeLocalizer(locale)
```

to 

```
localizer(Globalize)
```

(as I see now in the examples), but nothing seems to revert it back to start the week on Monday. 

![image](https://cloud.githubusercontent.com/assets/985978/14412338/78bdcad8-ff58-11e5-8eed-1cac89dec088.png)

What could cause this all of the sudden?
**Edit:** I've confirmed via a screenshot I sent last night that it was indeed fine and started the week on monday. Today when I opened my laptop and reloaded the page it flipped to Saturday. 

**Edit 2:**
At first I thought it was too far fetched to mention, but I've just tried to move my clock a day forward and it fixed it! There is something either on "Sunday" or today's date that causes the calendar to start on Saturday. I cannot replicate this on the Demo site. 

Thanks!

harel:
And at the stroke of midnight, the calendar "righted" itself and reverted back to Monday as first day of the week.

jquense:
that is a very weird issue... honestly I have idea what might have caused it.

harel:
This was tested on multiple computers/OSes by the way and by myself in the UK and my partner in the West Coast (PST). Both of us saw the calendar in the same way during yesterday. I've never seen anything like it

harel:
I can confirm that every Sunday, the calendar flips to having Saturday as the starting day of the week in my calendar. This is a calendar initialised to the en-UK locale, using Globalize. Its fine all week, except on Sundays where it goes crazy.

harel:
This could very well be a problem in Globalize. 
When ``userlocale`` is a Globalize object loaded with 'en-UK' locale: 

```
// On any non sunday day:
userlocale.formatDate(new Date(), {raw:'e'});
> "2"
```

```
// On a sunday :
[...]

PR #81, opened 2016-04-13
Fix first day of the week evaluation when using Globalize

harel:
This PR resolves issue #78  where the evaluation of first day of the week fails on every Sunday (for en-GB locale at least). On Sundays the first day of the week was evaluated as Saturday instead of Monday. 
The PR is an update to the ```firstOfWeek``` function in ```src/localizers/globalize.js``` and it uses the CLDR supplemental data to correctly retrieve the first day of the week. Failing to use this method, there is a fallback to using the original Math based approach to determine the first day. 

Another change is an update to packages.json file, which removes the ```files``` key. The problem with that was that having files there, meant any ```npm install``` from github resulted in JUST those files listed, and none of the src files. Apparently this is a common problem (also exists in projects like react-router etc.).

### DIFF
--- a/src/localizers/globalize.js
+++ b/src/localizers/globalize.js
@@ -43,11 +43,12 @@ export default function(globalize) {
   // method of getting first day of week.
   function firstOfWeek(culture) {
     try {
-        const days = ['sun', 'mon', 'tue', 'wed', 'thur', 'fri', 'sat'];
+        const days = ['sun', 'mon', 'tue', 'wed', 'thu', 'fri', 'sat'];
         const cldr = locale(culture).cldr;
         const territory = cldr.attributes.territory;
         const weekData = cldr.get('supplemental').weekData;
         const firstDay = weekData.firstDay[territory || '001'];
+        console.log("FIRST OF WEEK", days.indexOf(firstDay));
         return days.indexOf(firstDay);
     } catch (e) {
         console.error("Could not use suplemental", e)
--- a/src/localizers/globalize.js
+++ b/src/localizers/globalize.js
@@ -50,6 +50,7 @@ export default function(globalize) {
         const firstDay = weekData.firstDay[territory || '001'];
         return days.indexOf(firstDay);
     } catch (e) {
+        console.error("Could not use suplemental", e)
         // maybe cldr supplemental is not loaded? revert to original method
         const date = new Date();
         //cldr-data doesn't seem to be zero based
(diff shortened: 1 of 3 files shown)


src/localizers/globalize.js
```diff
@@ -37,13 +37,28 @@ export let formats = {
...
+  // method of getting first day of week.
   function firstOfWeek(culture) {
-    let date = new Date();
-    //cldr-data doesn't seem to be zero based
-    let localeDay = Math.max(
-      parseInt(locale(culture).formatDate(date, { raw: 'e' }), 10) - 1, 0)
-
-    return Math.abs(date.getDay() - localeDay)
+    try {
+        const days = ['sun', 'mon', 'tue', 'wed', 'thu', 'fri', 'sat'];
```
jquense: sorry for my lack of experience here, but will these be english for every culture?

harel:
Those are constants used in the cldr to denote a day. The will be the same for all locales, as far as I know. Those are just used to determine the numeric value of the day. The same method is used inside Globalize to get first day of the week (though their method is not exposed outside). The supplemental package of CLDR contains one block of all the territories (by country code) and their first day of the week using those markers.

harel:
One thing - i forgot to revert my update to the example whre i defaulted to en-GB locale instead of fr. i'll fix it now

src/localizers/globalize.js
harel: Those are constants used in the cldr to denote a day. They will be the same for all locales, as far as I know. Those are just used to determine the numeric value of the day. The same method is used inside Globalize to get first day of the week (though their method is not exposed outside). The supplemental package of CLDR contains one block of all the territories (by country code) and their first day of the week using those markers.

jquense:
ok great :) 

One thing I would change is instead of using `console.warn` directly just use the `warning` package to be consistent with warnings elsewhere in the module

harel:
Sure. I'll take a look now. 
Didn't know about it :)

harel:
You mean the ```invariant``` packages yes? I see you are installing warning but using invariant throughout.

harel:
done

jquense:
invariant will throw a hard error, where warning will only warn... I think in this case we want to warn not throw?

harel:
Yes, that is correct. I'll change to warning.

jquense:
thanks

PR #84, opened 2016-04-18
Expose components to allow drag & drop example

srajko:
Exposes components that can be extended with `react-dnd` to provide drag & drop functionality, and adds a d&d example.


srajko:
@leesalminen @jquense I tweaked the approach a bit - instead of having to isolate and expose a few different event-related components that are used in different views, I added a `EventWrapper` component which can wrap events wherever they are used.  The default `EventWrapper` is just a pass-through component (`render` returns its child), so it doesn't change how the calendar is rendered, and it allows the d&d code to override a single class to enable d&d of events.

The d&d example now allows events to be dragged from the Month, Week, and Day views.  The next step would be to do add a similar Wrapper class for time slots, and extend the example to allow events to be dropped into time slots.

Please let me know if you have any feedback.

srajko:
I added wrappers for drop targets, and added some style modification when a drag source is dragged over a drop target.  The style modification happens directly on the DOM node, similar to how `connectDragSource` adds the `draggable` attribute to the DOM node.

With the `isOver` visual, you can see a few areas where the drag-drop has gaps.  For example, in the month view, `rbc-row-segment` divs are preventing the drop onto the background cell.  When the `rbc-row-segment` has an event in it, that feels fine, but when it's empty, it feels broken.  I'm not sure what the best way around this is.

jquense:
hey @srajko sorry for the lack of response. I am very excited about this, but I just don't have a ton of time right now to take a really close look.

Ultimately I think I'd like to downplay user access to those structural components that are exposed for integrations like this, but not for general use or styling. That, I think would just mean wrapping up the behavior into some sort of "enhancer" function that a user can jsut drop in.

```
let DndBigCalendar = withDragAndDrop(BigCalendar);
```

where `withDragAndDrop` just sets the default components, or wraps the Calendar in another component that handles it.

This is really cool though :) thanks for all the work you'r doing here!

jquense:
@srajko anything you need from me on this? Is it at a place you feel good about?

srajko:
@jquense Hey sorry I got busy and am travelling at the moment.  I think this needs a little more work before it's finalized - the biggest problem is with the gaps in drop areas mentioned in https://github.com/intljusticemission/react-big-calendar/pull/84#issuecomment-214773099.  I'd also like to get it to the form you suggested (`let DndBigCalendar = withDragAndDrop(BigCalendar);`).

If you can think of a good way to get around the `rbc-row-segment`s interfering with the drop zones, that would be helpful. Perhaps styling the spacer segments as `visibility:hidden` would do the trick.

srajko:
@jquense Finally got a chance to get back to this.  `visibility:hidden` did not help resolve the gaps in the drop zones.  Also, it wasn't just `rbc-row-segment`s - `rbc-row` divs are also covering the drop areas.

The only way I was able to get the drop areas to be fully usable was to increase their z-index during an active drag (directly, and with a `setTimeout`, so it feels extra hacky).  It works, though it does have a visual side-effect of the grid borders showing through the event divs above them.

I pushed the change so you can try it out.  The only other way I can think of is to add transparent divs with a high z-index that would serve as the actual drop targets (add them during an active drag).

jquense:
@srajko thanks for the work, the z-index thing is ugly :/ I'm not sure there is away to fall through on those though...have you tried setting `pointer-events: none` on the layers above, that should work I think, but it's not widely supported in browsers.

I wish it was easy to have the drop targets ignore levels above them...

srajko:
@jquense Thanks for the `pointer-events` suggestion - that worked really well!

I also rebased on top of master, and changed the example to define an explicit `DndBigCalendar` class.  Does the class address your concerns from https://github.com/intljusticemission/react-big-calendar/pull/84#issuecomment-214776510?  FWIW, I think `EventWrapper` and `BackgroundWrapper` could have uses outside of dnd - they just provide a single customization point for all components that represent events or time slots (actually, `TimeSlotWrapper` or something similar might be a better name than `BackgroundWrapper`).

One big thing still missing is that `BackgroundWrapper` doesn't know what time period it represents. In `TimeSlot.jsx` I think we can pass `this.props.value` (https://github.com/intljusticemission/react-big-calendar/blob/899f7a3bf3a2f7d1deab0786769447d9f1ca8a4c/src/TimeSlot.jsx#L26), but in `BackgroundCells.jsx` (https://github.com/intljusticemission/react-big-calendar/blob/899f7a3bf3a2f7d1deab0786769447d9f1ca8a4c/src/BackgroundCells.jsx#L45) looks like we'll need to have additional data passed in from `Month.jsx` and `TimeGrid.jsx`.

srajko:
I have the date values passed to `BackgroundWrapper` now, so the dnd example can tell what an event was dropped on.

I think this gives enough of an extension point to make dnd feasible, so I'm taking off the WIP.  Please let me know if there's anything you'd like me to tweak / polish.  For example, now that `values` is being passed to `BackgroundCells`, we might be able to do away with passing `slots` (it should be the length of `values`).

srajko:
@jquense I rebased the branch. Is there anything you would like me to do to get this merged in?

mtetlow:
Would really love to see this merged

jquense:
Sorry team gonna be a bit, lit going on in my personal life and this, while awesome, is a bit down the list. 

I will take a look tho and provide feedback when I can feel free to keeping ping me tho

TigleyM:
@jquense Have you had time to take a look/review ?

srajko:
Rebased again to resolve conflicts.  Also kudos to @TigleyM for extending the example to actually update events on drop.

123batman:
@jquense Could you please merge this? It is a pity that @srajko 's and @TigleyM 's work just sits here collecting dust :(
I would really like to use your component but without this I can't.

jquense:
it is a pity! I'm hoping to get some time real soon to dive in optimistically this week

TeaBough:
Playing with this PR I've noticed that is unfortunatly not compatible with selectable ...

jquense:
yes we may need to tweak the logic as well. maybe even make it the same sort of HOC addons, tho I've some concerns there around perf and feasibility. short term I'm not quite sure what would be required

srajko:
@TeaBough could you please post an example of what you're trying to do, to see if we can make it work?

TeaBough:
`DndBigCalendar.js` simply add `selectable` : 
```
return <BigCalendar selectable  {...propsCopy} />
```
And then dragging doesn't work. 
What I'd like to be able to do is drag and drop event and still being able to select slots of time on free space.

TeaBough:
I'm trying to prevent `mouseDown` event bubbling when on an event without any success...

jquense:
I think is is probably because the selection code is really greedy, and fires even when over events (just a guess)

TeaBough:
Btw I really don't get why in `Selection.jsx` the listener is attached to the all document instead of simply the DayColumn ?

TeaBough:
Well, I'll try to refactor `Selection.jsx` to make it less greedy because this is really problematic for me. I want to add a small space in the event that is clickable which would allow me to quickly alter an event (kind of the same behaviour as in Trello). But that would necessitate to prevent the click from the whole event to trigger, which seems impossible without modifying the way we handle events listeners...

jquense:
@TeaBough I don't think you need to change anything to add custom click behavior to events, event clicks are already handled fine with selectable enabled it only assumes your dragging if you actually drag the mouse not just clicking

TeaBough:
Well not exactly but I had to add to my new `handleClick` method the `clearTimeout` to prevent this to happen : https://github.com/intljusticemission/react-big-calendar/blob/master/src/DayColumn.jsx#L260

jquense:
why can you use the onSelectEvent prop?

TeaBough:
Because onSelectEvent is called when you click everywhere on the event. I want to trigger a special modal when clicking on a small part of the event : 
operation_teetsh

And I don't know how to do that using only the `onSelectEvent` since I won't differenciate a click on the right top corner and a click on the rest of the event.

src/DayColumn.jsx
```diff
@@ -155,24 +156,26 @@ let DaySlot = React.createClass({
...
-        >
-          <div className='rbc-event-label'>{label}</div>
-          <div className='rbc-event-content'>
-            { EventComponent
-              ? <EventComponent event={event} title={title}/>
-              : title
-            }
+        <EventWrapper event={event}>
+          <div 
+            key={'evt_' + idx}
```
TeaBough: I think that the key should be inside the EventWrapper component.

srajko: Indeed, that does make the warnings go away - good catch!

srajko:
I updated the example to apply the hover styling via `cloneElement` and a css class, I didn't like the way I was doing it via the `ref` before.

srajko:
Since `BackgroundWrapper` and `EventWrapper` always wrap a single div, we can also just call `connectDropTarget` and `connectDragSource` directly instead of going through the ref / dom instance, which simplifies the example a little.

jquense:
I _finally_  had some time to give this the attention it deserves.

please check out #217 which has a few commits by me. I mostly just moved stuff around so that it was easier to release and changed some names and messed with the styles a bit. Let me know if I accidentally changed/removed something that you need here

srajko:
@jquense the changes look good to me, thanks for tightening things up.  I just pushed a small path fix on top of your commits (to this PR).  There is an issue with all-day events no longer being draggable (in weekly view, for example), I'll take a closer look at that tonight.

srajko:
Looks like it was a styling issue.  I readded `pointer-events: all` to `.rbc-event`.  Does that mess up something else?

jquense:
Ok. lets merge this. we still need some documentation, would anyone like to add an 'addons' section to the docs showing how to use this? I want to try and keep this a bit separate from main component, at least for a bit.

wonbyte:
Just tried this out. Can not wait for it to be merged in. Really looking forward to it.

jquense:
Merged in #231 🎉 🎉 

Thanks so much everyone! and sorry for the terrible delay on my part. 👏  Awesome work

PR #91, opened 2016-05-14
allow a custom toolbar to be passed in

craig-day:
Allows a custom Toolbar component to be passed in and used in place of the default.  This makes it easy to move around the date and view navigation buttons. 

Sample custom toolbar:

@jquense


jquense:
thanks!

craig-day:
@jquense any chance we can get a patch version bump so I can pull this in locally? Thanks for the quick merge!

PR #93, opened 2016-05-17
allow replacing components for day, week, month and agenda, not just event

cellog:
This allows for extreme customization.  For instance, a week view for a
sign-up sheet where only 2 days are allowed should be able to display just those 2 days.  Another example is implementing a special grid that blocks off unavailable times using a different color.

Currently, the only way to do this is to extend Calendar, copy/paste the render() method, replace the view in question, and that's not very DRY.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,5 @@ build/Release
 # Dependency directory
 # https://www.npmjs.org/doc/misc/npm-faq.html#should-i-check-my-node_modules-folder-into-git
 node_modules
+
+.idea
--- a/src/Calendar.jsx
+++ b/src/Calendar.jsx
@@ -286,12 +286,13 @@ let Calendar = React.createClass({
       agenda: PropTypes.shape({
         date: elementType,
         time: elementType,
-        event: elementType
+        event: elementType,
+        component: PropTypes.element
       }),
 
-      day: PropTypes.shape({ event: elementType }),
-      week: PropTypes.shape({ event: elementType }),
-      month: PropTypes.shape({ event: elementType })
+      day: PropTypes.shape({ event: elementType, component: PropTypes.element }),
+      week: PropTypes.shape({ event: elementType, component: PropTypes.element }),
+      month: PropTypes.shape({ event: elementType, component: PropTypes.element })
     }),
 
     /**
@@ -338,7 +339,7 @@ let Calendar = React.createClass({
 
     formats = defaultFormats(formats)
 
-    let View = VIEWS[view];
+    let View = this.props[view].component || VIEWS[view];
     let names = viewNames(this.props.views)
 
     let elementProps = omit(this.props, Object.keys(Calendar.propTypes))
--- a/src/Calendar.jsx
+++ b/src/Calendar.jsx
@@ -339,7 +339,8 @@ let Calendar = React.createClass({
 
     formats = defaultFormats(formats)
 
-    let View = (this.props[view] && this.props[view].component) || VIEWS[view];
+    const viewComponent = (components && components[view] && components[view].component) || VIEWS[view];
+    let View = viewComponent;
     let names = viewNames(this.props.views)
 
     let elementProps = omit(this.props, Object.keys(Calendar.propTypes))
--- a/src/Calendar.jsx
+++ b/src/Calendar.jsx
@@ -339,7 +339,7 @@ let Calendar = React.createClass({
 
     formats = defaultFormats(formats)
 
-    let View = this.props[view].component || VIEWS[view];
+    let View = (this.props[view] && this.props[view].component) || VIEWS[view];
     let names = viewNames(this.props.views)
 
     let elementProps = omit(this.props, Object.keys(Calendar.propTypes))


jquense:
Hi there, this can actually already be done by specifying a component in the `views` prop with a an object.

```js
views={{
  month: MyMonth
}}
```

The reason this is seperate from the other components api is because views are special, they require that the component implement a specific API, and provide certain static methods. It could all be documented a bit better, if you'd like to enhance that a bit I'd take that PR :)

cellog:
Ah cool, thanks.

I would actually love to take a stab at simplifying the code base overall

I will open an enhancement ticket for ideas if this appeals to you.

Thanks,
Greg


Sent using CloudMagic Email [https://cloudmagic.com/k/d/mailapp?ct=pa&cv=8.4.52&pv=5.1.1&source=email_footer_2] On Tue, May 17, 2016 at 6:18 AM, Jason Quense < notifications@github.com [notifications@github.com] > wrote:
Hi there, this can actually already be done by specifying a component in the views prop with a an object.

views={{
  month: MyMonth
}}

The reason this is seperate from the other components api is because views are special, they require that the component implement a specific API, and provide certain static methods. It could all be documented a bit better, if you'd like to enhance that a bit I'd take that PR :)

jquense:
always happy to get feedback and contributions. Feel free to open up tuff to chat about it

cellog:
actually, I just tried this, and it didn't work.  So it should be:

`<Calendar views={{month: MyMonth}}`?

I don't see where this is even accessed in the code.  What I see is on line 342, View is set to VIEWS[view] but nowhere is the `views` prop accessed.

jquense:
views prop is accessed, and the names of each view are passed to the toolbar, but you are right there is a bug :P It should be something like

```js
function getViews(views) {
  if (Array.isArray(views) {
    return transform(views, (obj, name) => obj[value] = VIEWS[name], {})
  }
 
  return defaults(views, VIEWS)
}

let view = getViews(views)[viewName]
```

jdeniau:
I started making a PR for this, but I ran into a little detail: 
With the `defaut(views, VIEWS)` implementation, you can not limit the views to be rendered.

The solutions I see: 
* if you specify the views, then no defauts views are used, only the one listed: same comportment as now, but weird if you want to use defaut "month" view and custom "week" for example
* allow a syntax like this, but I found it a little weird: 
```js
<BigCalendar  views={{ week: MyWeek, month: 'month' }} />
```
* Use the same implementation as before but use default views as key:
```js
<BigCalendar  views={{ week: MyWeek, month: true }} />
```
* set a custom property for view customization

Do you have another idea ? What would be the choice that matches the most the current API ?

jquense:
I think the second option is the ones, tho maybe we just use a boolean instead of repeating the key,then it works for explicit disabling as well. `{ month: true}`

cellog:
This solution certainly works. I think it might be useful to document each subcomponent individually and provide some examples of combining them. Then we could do something like this instead:

<BigCalendar monthComponent={(
<MyMonth>
{BigCalendar.Month.generateDays(
MyDay)}
</MyMonth>
)} />

Or leave generateDays blank to use default day component.

This would also make it easier to tweak options passed to existing components

<BigCalendar weekComponent={(
<BigCalendar.Week start={...}>
{BigCalendar.Week.generateDays()}
</BigCalendar.Week>
)} />

Obviously not fleshed out yet. What do you think?


Sent using CloudMagic Email [https://cloudmagic.com/k/d/mailapp?ct=pa&cv=8.5.28&pv=5.1.1&source=email_footer_2] On Sun, May 22, 2016 at 7:44 AM, Jason Quense < notifications@github.com [notifications@github.com] > wrote:
I think the second option is the ones, tho maybe we just use a boolean instead of repeating the key,then it works for explicit disabling as well. { month: true}

jquense:
Generally i'd prefer to not document the internal View's, as they really aren't part of the public API. Keeping the public API surface area low helps tremendously with maintainability and the abilty to iterate internally on components.

If we were going to pass elements (vs Component classes) then using `children` makes more sense than a `views` props. And I do like how just adding the Views to children feels as an API, but it makes stuff kinda tough internally because they aren't really children, each view replaces the other, we'd have to require keys (ugly :/) and we'd have to clone the elements each time, which incurs a slight perf concern. I does allow the uscase of being able to adjust view props dynamically, which is cool, but thats also doable (albeit in an uglyier way) with the classes.

For folks that want to dig into the nitty gritty of the internal components then I'd prefer the extra work of wrapping thembecause it isolates internal API usage to a component spec, vs each use of the component.

cellog:
Yes, something like this?

<BigCalendar monthComponent={(props, children) => <MyMonth {...props} myspecialprop={...} >{children}</MyMonth>} />

That is much cleaner, and easily controlled.

Is that more what you were thinking?


Sent using CloudMagic Email [https://cloudmagic.com/k/d/mailapp?ct=pa&cv=8.5.28&pv=5.1.1&source=email_footer_2] On Sun, May 22, 2016 at 9:28 AM, Jason Quense < notifications@github.com [notifications@github.com] > wrote:
Generally i'd prefer to not document the internal View's, as they really aren't part of the public API. Keeping the public API surface area low helps tremendously with maintainability and the abilty to iterate internally on components.

If we were going to pass elements (vs Component classes) then using children makes more sense than a views props. And I do like how just adding the Views to children feels as an API, but it makes stuff kinda tough internally because they aren't really children, each view replaces the other, we'd have to require keys (ugly :/) and we'd have to clone the elements each time, which incurs a slight perf concern. I does allow the uscase of being able to adjust view props dynamically, which is cool, but thats also doable (albeit in an uglyier way) with the classes.

For folks that want to dig into the nitty gritty of the internal components then I'd prefer the extra work of wrapping thembecause it isolates internal API usage to a component spec, vs each use of the component.

jquense:
It still think an `views` object with Components is the easiest and most straightforward way forward here. The code already expects that and advantages of the other approaches are significant enough to warrent a big api change.

@jdeniau I'm 👍  on the `{ month: MyMonth, week: true, day: false }` API if you want to PR that. We can revisit the API in general at a later point if there are significant benefits, right now I'd like to just fix this bug.

PR #94, opened 2016-05-17
Fix full week span event being truncated on last day of the week.

okcompute:
This should fix issue #36 (the week view issue).

**Note**:  I'm not sure of the potential collateral this fix could bring up. I manually tested that it works for my use cases (month and week views) so there may not be any. Should we start adding functional tests?

**Second Note**: I needed to update some unrelated code because the `npm test` command was failing. There was a lint error.

### DIFF
--- a/src/TimeGrid.jsx
+++ b/src/TimeGrid.jsx
@@ -93,7 +93,9 @@ let TimeGrid = React.createClass({
 
     allDayEvents.sort((a, b) => sortEvents(a, b, this.props))
 
-    let segments = allDayEvents.map(evt => eventSegments(evt, start, end, this.props))
+    let {first, last} = endOfRange(range);
+
+    let segments = allDayEvents.map(evt => eventSegments(evt, first, last, this.props))
     let { levels } = eventLevels(segments)
 
     return (


jquense:
Hi, thanks for the PR. I am a bit afraid to just remove the `startOf` logic on the util method as that may have far reachin effects and should actually be the issue here. I haven't checked but I think the problem is that I missed a spot when normalizing how ranges are calculated in: https://github.com/intljusticemission/react-big-calendar/commit/a2a49c88a404ca97ea95274bde7249546badc016 I would think that a fix along those lines may be better

okcompute:
Hi,

Thanks for the quick feedback on this pull request :)

For my understanding of the issue, this is the `diff` function which is faulty. When I debugged my use case, the input parameters sent to the function were correct.

`dateA` was sunday at 00:00 and `dateB` was saturday at 23:59:59.

As it was, the function computation was returning 6 days span for these two dates. This is because of the `startOf` applied to `dateB`. This effectively removed 23 hour and 59 minutes from the full span. From my understanding, that *is* the source of the issue causing the function returning 6 instead of the expected 7 days span.

I would understand keeping the `startOf` on the `dateA` but I'm not sure about `dateB`. Maybe `endOf` could be applied to `dateB`? In the end, offsetting both dates by either `startOf` or `endof` seems inappropriate for this API (IMHO).

Is there other post-process that is applied after this function that could do a +1 on the span?

Is applying `startOf` on dates make senses for other `unit` but `day`?

Thanks again!

jquense:
glad to try and help, @okcompute if those are the dates the `diff` function is correct, ranges are _exclusive_ so if you want a full week the it's Sunday to Sunday, and not Sunday to Saturday 23:59:59. Which is why I don't think the problem is in the `diff` method, but further up. since this is working for Month weeks, and the AllDay views essentially use the same code it probably a min/max that is not being passed correctly.

jquense:
my guess is that `start` and `end` are wrong here: https://github.com/intljusticemission/react-big-calendar/blob/master/src/TimeGrid.jsx#L96

okcompute:
This is not I'm witnessing. Both dates are correct at that point.

This is where I started seeing `6` days span instead of `7` days span:

https://github.com/intljusticemission/react-big-calendar/blob/master/src/utils/eventLevels.js#L17

and why I started targeting the `diff` function.

okcompute:
Oh, I did not notice your first answer about the `diff` function. So indeed, if for the `span` to be 7 the datess need to be sunday to sunday, the function is correct. Ok, i'll do another debug pass to isolate where the end date is computed and set to saturday @ 23:59:59 (instead of sunday @ 00:00:00)

jquense:
I thought you said above the start/end dates were Sunday to Saturday? if they are, those are incorrect.

okcompute:
Yes, sorry. Like I said, I missed one of your comment before posting my comment.

jquense:
hehe, you correction popped up just as I posted. Whereever the problem is, you can see how Month uses `endOfRange` which is the place that adds 1 to the end to get the correct range.

okcompute:
I think the fix is now correct!

src/TimeGrid.jsx
```diff
@@ -93,7 +93,11 @@ let TimeGrid = React.createClass({
 
     allDayEvents.sort((a, b) => sortEvents(a, b, this.props))
 
-    let segments = allDayEvents.map(evt => eventSegments(evt, start, end, this.props))
+    var _endOfRange = endOfRange(range);
+    var first = _endOfRange.first;
+    var last = _endOfRange.last;
+
+    let segments = allDayEvents.map(evt => eventSegments(evt, first, last, this.props))
```
jquense: thanks! would you mind writing this as destructuring?

`let { first, last } = endOfRange(range);`

its a bit less noisy, not a big deal if you can't

okcompute: No problem! Didn't notice ES6 was an option :)

jquense:
🎉

okcompute:
Hehe! Thanks a lot :smile:

okcompute:
No rush but I'm curious. When a new release is planned?

jquense:
when I get a free moment!

okcompute:
Hehe. Cool.  no problem. I understand :smile:

PR #98, opened 2016-05-22
allow to pass custom views

### LINKED ISSUE #97: Show only work days for "week" view ?
Hi, 

First of all thanks you for the job you have done here !

Is it possible to "hide" some days, or display only work days.

I saw in your constant file a `WORKWEEK` constant, but seems never used.
I saw the "range" method in the Week.jsx file is calling `endOf('week')` .

I tried hiding in in CSS but did not manage to do this.

Do you have a way to do it ? I can make you a PR if you point me on something, but as the localizers does not have the notion of working days, it is not really easy. The solution I may have is to force de Week view range. 

Thank you

jdeniau:
Fixes #97 

Do follow the comments of #93 

I did not manage to set the propType good in `src/utils/propTypes.js`, can you help me with that ? 

Doing this, I tried to extends the `Week` view, but it was not as easy as I though, so I had to reimplement a full view. The fact is that we can not extends the `range` method. I do not search very well for that as it is not the purpose of this code, but it might be a good future evolution.

### DIFF
--- a/examples/App.js
+++ b/examples/App.js
@@ -29,7 +29,8 @@ const Example = React.createClass({
       selectable: require('./demos/selectable'),
       cultures: require('./demos/cultures'),
       popup: require('./demos/popup'),
-      rendering: require('./demos/rendering')
+      rendering: require('./demos/rendering'),
+      customView: require('./demos/customView'),
     }[selected];
 
     return (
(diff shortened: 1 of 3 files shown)


dmoskovsov:
+1

jdeniau:
up ? thanks

jquense:
hi, sorry fo the delay there. I actually think it makes sense _not_ to add an example for the custom views right now. I appreciate documentation of course, so perhaps instead of removing it completely we just remove the link from the examples right now? I want to make sure the API is really solidified before telling everyone about it ;)

jdeniau:
Done. Thanks, no problem for the delay :)

jquense:
thanks!

jdeniau:
Thanks a lot to you, as I said there is still a warning on the `propTypes`, if you can give a look at it !

jdeniau:
Sorry to bother you with that but can you drop a new tag for this feature ? (I can not checkout the master branch as the packages.json force the deletion of the `src` dir).

Thanks

jquense:
yup I'll get to it in a bit

PR #99, opened 2016-05-23
refactor TimeGutter to TimeSliceGroup/TimeSlice

cellog:
This change adds

1. The render() method of TimeGutter now generates a series of TimeSliceGroups that render individual TimeSlice
2. react storybook story to test the output.

### DIFF
--- a/src/TimeGutter.jsx
+++ b/src/TimeGutter.jsx
@@ -1,39 +1,53 @@
 import React from 'react';
-import cn from 'classnames';
 import dates from './utils/dates';
-import localizer from './localizer'
+
+import TimeSliceGroup from './TimeSliceGroup.jsx'
 
 let TimeGutter = React.createClass({
 
   propTypes: {
     step: React.PropTypes.number.isRequired,
+    slices: React.PropTypes.number,
+    now: React.PropTypes.instanceOf(Date).isRequired,
     min: React.PropTypes.instanceOf(Date).isRequired,
-    max: React.PropTypes.instanceOf(Date).isRequired
+    max: React.PropTypes.instanceOf(Date).isRequired,
+    showlabels: React.PropTypes.bool,
+    timesliceClassnames: React.PropTypes.string,
+    culture: React.PropTypes.string,
+    timeGutterFormat: React.PropTypes.string
+  },
+
+  getDefaultProps() {
+    return {
+      step: 10,
+      slices: 2,
+      now: new Date(),
+      showlabels: true,
+      culture: 'en'
+    }
+  },
+
+  renderTimeSliceGroup(key, isNow, date) {
+    return <TimeSliceGroup key={key} isNow={isNow} slices={this.props.slices} showlabels={this.props.showlabels}
+                           culture={this.props.culture} size={this.props.step*this.props.slices} value={date}
+                           format={this.props.timeGutterFormat}
+                           classNames={this.props.timesliceClassnames}
+    />
   },
 
   render() {
-    let { min, max, step, timeGutterFormat, culture } = this.props;
-    let today = new Date()
+    let { min, max, step, slices: slices, now } = this.props;
     let totalMin = dates.diff(min, max, 'minutes')
-    let numSlots = Math.ceil(totalMin / step)
+    let numSlots = Math.ceil(totalMin / (step * slices))
     let date = min;
+    let next = date;
     let children = []; //<div key={-1} className='rbc-time-slot rbc-day-header'>&nbsp;</div>
+    let isNow = false
 
     for (var i = 0; i < numSlots; i++) {
-      let isEven = (i % 2) === 0;
-      let next = dates.add(date, step, 'minutes');
-      children.push(
-        <div key={i}
-          className={cn('rbc-time-slot', {
-            'rbc-now': dates.inRange(today, date, next, 'minutes')
-          })}
-        >
-        { isEven && (
-            <span>{localizer.format(date, timeGutterFormat, culture)}</span>
-          )
-        }
-        </div>
-      )
+      isNow = dates.inRange(now, date, dates.add(next, step * slices - 1, 'minutes'), 'minutes');
+      next = dates.add(date, step * slices, 'minutes');
+      children.push(this.renderTimeSliceGroup(i, isNow, date));
 
       date = next
     }
--- a/src/TimeSlice.jsx
+++ b/src/TimeSlice.jsx
@@ -1,0 +1,36 @@
+import React, { PropTypes, Component } from 'react'
+import cn from 'classnames'
+import localizer from './localizer'
+
+export default class TimeSlice extends Component {
+  static propTypes = {
+    format: PropTypes.string.isRequired,
+    value: PropTypes.instanceOf(Date).isRequired,
+    isNow: PropTypes.bool,
+    classNames: PropTypes.string,
+    showlabel: PropTypes.bool,
+    selected: PropTypes.bool,
+    culture: React.PropTypes.string,
+    format: React.PropTypes.string.isRequired
+  }
+
+  static defaultProps = {
+    time: () => '',
+    isNow: false,
+    showlabel: true,
+    selected: false,
+    culture: 'en'
+  }
+
+  render() {
+    return (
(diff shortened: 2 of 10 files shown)


src/TimeSlice.jsx
```diff
@@ -0,0 +1,36 @@
...
+import localizer from './localizer'
+
+export default class TimeSlice extends Component {
+  static propTypes = {
+    format: PropTypes.string.isRequired,
+    value: PropTypes.instanceOf(Date).isRequired,
+    isNow: PropTypes.bool,
+    classNames: PropTypes.string,
+    showlabel: PropTypes.bool,
+    selected: PropTypes.bool,
```
jquense: unused prop, lets remove it.

src/TimeSliceGroup.jsx
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  renderSlice(i, className, value) {
+    return <TimeSlice key={i} showlabel={this.props.showlabels && !i} format={this.props.format}
+              isNow={this.props.isNow} classNames={className} value={value} />
+  }
+
+  renderSlices() {
+    let className = this.props.timesliceClassnames ? `${this.props.timesliceClassnames} ` : ''
+    switch (this.props.slices) {
```
jquense: this isn't ideal...first the classes aren't prefixed, and they are limited arbitrarily. What is the value of the className here anyway? A user could use `nth-child` to target the one they want if need be.

src/TimeSliceGroup.jsx
```diff
@@ -0,0 +1,61 @@
+import React, { PropTypes, Component } from 'react'
+import TimeSlice from './TimeSlice.jsx'
+import date from './utils/dates.js'
+
+export default class TimeSliceGroup extends Component {
+  static propTypes = {
+    slices: PropTypes.oneOf([2,3,4,5]),
```
jquense: like the comment below, this should allow any number

src/TimeGutter.jsx
```diff
@@ -1,39 +1,53 @@
...
+      step: 10,
+      slices: 2,
+      now: new Date(),
+      showlabels: true,
+      culture: 'en'
+    }
+  },
+
+  renderTimeSliceGroup(key, isNow, date) {
+    return <TimeSliceGroup key={key} isNow={isNow} slices={this.props.slices} showlabels={this.props.showlabels}
```
jquense: for consistency, if components are multi-line can we do 1 prop per line and surround the whole component in `( )`

src/TimeGutter.jsx
```diff
@@ -1,39 +1,53 @@
...
+                           culture={this.props.culture} size={this.props.step*this.props.slices} value={date}
```
jquense: do the size calculation in variable at the top of the block please

src/TimeSliceGroup.jsx
```diff
@@ -0,0 +1,61 @@
...
+  static defaultProps = {
+    slices: 2,
+    isNow: false,
+    showlabels: true,
+    timesliceComponent: TimeSlice,
+    debug: false
+  }
+
+  renderSlice(i, className, value) {
+    return <TimeSlice key={i} showlabel={this.props.showlabels && !i} format={this.props.format}
```
jquense: long element, better split on multiple lines

src/TimeSliceGroup.jsx
```diff
@@ -0,0 +1,61 @@
...
+    showlabels: PropTypes.bool,
+    isNow: PropTypes.bool,
+    timesliceClassnames: PropTypes.string,
+    debug: PropTypes.bool
+  }
+  static defaultProps = {
+    slices: 2,
+    isNow: false,
+    showlabels: true,
+    timesliceComponent: TimeSlice,
```
jquense: unused

src/TimeSliceGroup.jsx
```diff
@@ -0,0 +1,61 @@
...
+    debug: false
+  }
+
+  renderSlice(i, className, value) {
+    return <TimeSlice key={i} showlabel={this.props.showlabels && !i} format={this.props.format}
+              isNow={this.props.isNow} classNames={className} value={value} />
+  }
+
+  renderSlices() {
+    let className = this.props.timesliceClassnames ? `${this.props.timesliceClassnames} ` : ''
```
jquense: use `classnames` over string concat here please

src/less/day-slot.less
```diff
@@ -1,5 +1,10 @@
 @import './variables.less';
 
+.rbc-timeslice-group + .rbc-timeslice-group {
+  border-top: 1px solid @cell-border;
+  margin-bottom: -1px;
```
jquense: the the negative margin isn't great, but I see that its needed until the Dayslots switch over to use these components...I would add a FIXME comment, or just include the DaySlot stuff in the PR as well.

jquense:
thanks! Made a few notes mostly nits. Is there a reason that you didn't also include the DaySlot here? It's a bit weird to have the code base in this intermediate state. I'd cut those components over as well. Even if you do want to do more work on the Day bits, you could make the structural change separate i'd think

src/TimeGutter.jsx
```diff
@@ -1,39 +1,53 @@
...
+                           format={this.props.timeGutterFormat}
+                           classNames={this.props.timesliceClassnames}
+    />
   },
 
   render() {
-    let { min, max, step, timeGutterFormat, culture } = this.props;
-    let today = new Date()
+    let { min, max, step, slices: slices, now } = this.props;
```
jquense: no need to rename `slices` in the destructure

src/TimeGutter.jsx
```diff
@@ -1,39 +1,53 @@
...
+    max: React.PropTypes.instanceOf(Date).isRequired,
+    showlabels: React.PropTypes.bool,
+    timesliceClassnames: React.PropTypes.string,
+    culture: React.PropTypes.string,
+    timeGutterFormat: React.PropTypes.string
+  },
+
+  getDefaultProps() {
+    return {
+      step: 10,
```
jquense: where is `10` coming from, I think this is being overwritten by the parent, I'd remove the default to avoid confusion

src/TimeGutter.jsx
```diff
@@ -1,39 +1,53 @@
...
+      slices: 2,
+      now: new Date(),
```
jquense: this is subtlety different logic then the old code, This value is calculated once, whereas the date in `render()` is refreshed often, which means that "now" will only update on a refresh. Admittedly the other one isn't much better, but lets avoid possible bugs unnecessarily.

src/TimeSliceGroup.jsx
```diff
@@ -0,0 +1,61 @@
+import React, { PropTypes, Component } from 'react'
+import TimeSlice from './TimeSlice.jsx'
+import date from './utils/dates.js'
+
+export default class TimeSliceGroup extends Component {
+  static propTypes = {
+    slices: PropTypes.oneOf([2,3,4,5]),
+    size: PropTypes.number.isRequired,
```
jquense: `size` might be better off having a more descriptive name. right now its a bit confusing as to what it is. maybe `totalMinutes` or something? 

It looks tho like `size` is never used as is. TimeGutter does `step * slices` and its then pulled back out as `size / slices` here. Maybe a better prop is `minutesPerSlice` instead?

jquense:
another note, the padding is missing on the labels.

cellog:
Awesome work so fast. I'm leaving for a week, maybe I should just create a PR from the branch that has all the work? I was trying to split them out for you.

I would do this when I return. All the unused props are used in the other thing, I was busy stripping them down.

Sent using CloudMagic Email [https://cloudmagic.com/k/d/mailapp?ct=pa&cv=8.5.28&pv=5.1.1&source=email_footer_2] On Mon, May 23, 2016 at 8:06 AM, Jason Quense < notifications@github.com [notifications@github.com] > wrote:
another note, the padding is missing on the labels.

jquense:
no rush, just go on vacation :P

cellog:
I implemented all of your concerns on the more recent PR.  Closing this one

src/TimeSliceGroup.jsx
cellog: this is impossible.  visually, when you get to more than 5 slices, they just look like gray lines

PR #102, opened 2016-05-28
Implement unified TimeGutter in Week and Day view using refactored code

cellog:
changes:

1. re-factored TimeSlice/TimeSliceGroup in TimeGutter removes some loop code in favor of mapping, makes selection easier to implement
2. re-factored DaySlot uses same code as TimeGutter to generate html
3. re-factored selection code reduces code size substantially, separates concerns (selection is not handled within components, but with higher-order components Selection and Selectable)
4. otherwise, it's the same!  Visually functioning the same way, external API identical

Potential API to expose:

1. because the selection is now configurable, it is possible to expose the function for creating the selection div
2. time slices, can make smaller slices (up to 5 per group) default is 2
3. rather than using new Date() for now, we should pass this as a prop, with default value of new Date().  Makes the component slightly more pure.

potential bugs to fix (pre-existing behavior):

1. selection of the all-day slots doesn't do anything (and never did).  Perhaps we need to expose onSelectAllDay?
2. react/react-dom can be upgraded to 15.1.0.  I left that for you to decide, I don't use anything here that breaks in earlier versions to my knowledge

P.S. had a nice vacation, hope your week was good :)


cellog:
Forgot to copy over gutter realignment and all-day events. Will do that later

src/Calendar.jsx
```diff
@@ -349,6 +349,7 @@ let Calendar = React.createClass({
     )
 
     let ToolbarToRender = components.toolbar || Toolbar
+    const now = new Date()
```
jquense: what would be the advantage to exposing this prop?  this should be an implementation detail

src/components/TimeGridHeader.jsx
```diff
@@ -0,0 +1,40 @@
...
+    range: PropTypes.arrayOf(React.PropTypes.instanceOf(Date)).isRequired,
+    onClick: PropTypes.func,
+    formatter: PropTypes.func,
+    gutterWidth: PropTypes.number,
+    gutterRef: PropTypes.func
+  }
+  static defaultProps = {
+    range: [],
+    onClick: () => null,
+    formatter: (date) => moment(date).format('ddd D/M'),
```
jquense: please remove this and use the localizers

jquense:
hi there, thanks for the PR. Most of my comments on the other PR still apply here, it also leaves Moment as the default and changes the localization strategy. 

like I mentioned in the other threads and PR please leave the localization code as it it was. The selection code as well, I disagree that this code both saves a significant amount of code or is simpler artitecturally. I think your approach to selection is a fine one i just disagree with the tradeoffs for this project and would prefer to not unify the selection approach. thanks again for all the work

jquense:
also glad your vacation was great.  Im actually also going to be on vacation till the middle of the week as well so may be a bit slow to respond!

jquense:
final bit for now. I'm happy to have a longer conversation about selection and explain why I think the current approach is preferable for this project.  but I'd like to do it in its own issue/PR and not mixed in with other changes.

cellog:
Hi,

I totally understand why, but I don't think you looked quite carefully enough to see how many of your comments are reflected in the code.  moment is gone from everything *except* for TimeGridHeader because I made a mistake (didn't notice it there), not because I'm trying to force moment!  I also reformatted per your requests, removed some things.

There are a couple of use cases that are less obvious that you should know about that are driving me:

1. Storing UI state has a valid use case, and it shows up most obviously when you are writing an app that is designed to look the same with javascript off (which is now easy to do with react, server-side rendering from the same components).  In this case, clicking "next" would reload the page, with the new ui state, for example.  Selecting would have to be impossible, this would be a view-only calendar, or handled with click on start, click on end.
2. I'd like to be able to easily re-configure the internal components to add drag and drop, or implement a subtle difference.

Also, regarding selection, the current code is broken in one subtle case:

If you click on an event, and then drag, it selects the slot instead of the event.  Users of google calendar (or any calendar program like iCal) would expect it to instead drag the event to a new time slot.

This detail requires a major refactor to fix, which I have done.

In addition, if you are using the calendar to schedule events with pre-set length (like, say, a 15 minute jury), you would want the selection code to instead display a dragging event of that length, and detect when it is not possible and display that information.  None of this is possible with the existing code, and is straightforward with the new selection code.  I actually need this feature, so is there any wiggle room here?

If you dig deeper you will see that localization is truly fixed, back to what it originally was with that one exception of the default value.  The stories all reflect this change as well (And is the clearest way to see the changes)

src/Calendar.jsx
cellog: rather than using new Date() for now, we should pass this as a prop, with default value of new Date(). Makes the component slightly more pure.  It allows easily testing the code for highlighting the current time, and also things like displaying a calendar for a different time, or 2 calendars on the same page with different now value.

src/components/TimeGridHeader.jsx
cellog: was a mistake, as I said, happy to do so.

jquense:
hi, I didn't mean to suggest you ignored my comments or that stuff was far off. just my quick glances at stuff and not having a computer it's bit terse sounding. 

> If you click on an event, and then drag, it selects the slot instead of the event. 

This isn't a bug it's a feature and by design. this is also how Google calendar works and it's intentional to allow larger "hit" target for event creation.  dnd can be handled via a drag handle. 

> In addition, if you are using the calendar to schedule events with pre-set length (like, say, a 15 minute jury), you would want the selection code to instead display a dragging event of that length, and detect when it is not possible and display that information. 

I'm not sure why the current code cant handle this,  it may not right now but there is no reason it can't... without a dramatic rewrite. 

like I said, we can talk about this stuff in _another_ PR that focuses just on selection and meeting your use case.  there too many changes here that make that all hard to discuss clearly.

src/components/DaySlot.jsx
```diff
@@ -0,0 +1,225 @@
...
+import dates from '../utils/dates';
+import { isSelected } from '../utils/selection';
+import { notify } from '../utils/helpers';
+import { accessor, elementType } from '../utils/propTypes';
+import { accessor as get } from '../utils/accessors';
+import localizer from '../localizer'
+
+export default class DaySlot extends Component {
+  static propTypes = {
+    formatter: PropTypes.func,
```
jquense: unused I believe now?

src/components/EventRow.jsx
```diff
@@ -0,0 +1,16 @@
+import React, { PropTypes, Component } from 'react'
+
+export default class EventRow extends Component {
+  static propTypes = {
+    param: PropTypes.string.isRequired
```
jquense: can we name something more descriptive?

src/components/TimeGrid.jsx
```diff
@@ -0,0 +1,238 @@
...
+    ...TimeGutter.propTypes,
+
+    end: PropTypes.instanceOf(Date),
+    start: PropTypes.instanceOf(Date),
+    selectRangeFormat: PropTypes.func.isRequired,
+    eventTimeRangeFormat: PropTypes.func.isRequired,
+    timeGutterFormat: PropTypes.string,
+    dayFormat: PropTypes.string.isRequired,
+    culture: PropTypes.string.isRequired,
+    components: PropTypes.shape({
```
jquense: this doesn't belong here right?

src/components/TimeGrid.jsx
```diff
@@ -0,0 +1,238 @@
...
+    titleAccessor: 'title'
+  }
+
+  _adjustGutter() {
+    const isRtl = this.props.rtl;
+    const header = this._headerCell;
+    const gutterCells = [findDOMNode(this.refs.gutter), ...this._gutters]
+    const isOverflowing = this.refs.content.scrollHeight > this.refs.content.clientHeight;
+
+    let width = this._gutterWidth
```
jquense: I'd actually be happy if this was also a prop as well for edge cases when the width can't be correctly measured (such as when hidden).  it still should do this logic if the prop is empty but could avoid it if a dev provides it

src/components/TimeGridAllDay.jsx
```diff
@@ -0,0 +1,39 @@
...
+  render() {
+    
+    return (
+      <div className="rbc-row">
+        <div ref={this.props.gutterRef} className="rbc-gutter-cell" style={this.props.gutterWidth ?
+          {width: this.props.gutterWidth} : {}}>
+          { message(this.props.messages).allDay }
+        </div>
+        <div className="rbc-allday-cell">
+          <SelectableBackgroundCells selectable={this.props.selectable}
```
jquense: does the original code allow selection in this component? I don't actually think it makes sense to allow that but maybe I am disagreeing with an earlier me?

src/components/TimeGridHeader.jsx
```diff
@@ -0,0 +1,45 @@
...
+    format: formats.dayFormat,
+    gutterRef: () => null
+  }
+
+  render() {
+    return (
+      <div className="rbc-time-header" ref={this.props.headerRef}>
+        <div className="rbc-row">
+          <div ref={this.props.gutterRef} className='rbc-gutter-cell' style={this.props.gutterWidth ?
+          {width: this.props.gutterWidth} : {}} />
```
jquense: the conditional isn't needed if gutterWidth is empty it won't render anything on style

src/components/TimeSliceGroup.jsx
```diff
@@ -0,0 +1,67 @@
...
+              showlabel={this.props.showlabels && !i}
+              format={this.props.timeGutterFormat}
+              culture={this.props.culture}
+              isNow={this.props.isNow}
+              value={value} />
+  }
+
+  renderSlices() {
+    let classNames = this.props.timesliceClassnames
+    switch (this.props.slices) {
```
jquense: see my previous note about this in the other PR

src/components/TimeSliceGroup.jsx
```diff
@@ -0,0 +1,67 @@
...
+    value: PropTypes.instanceOf(Date).isRequired,
+    showlabels: PropTypes.bool,
+    isNow: PropTypes.bool,
+    timeGutterFormat: PropTypes.string,
+    timesliceClassnames: PropTypes.string,
+    culture: PropTypes.string
+  }
+  static defaultProps = {
+    slices: 2,
+    step: 10,
```
jquense: see previous PR comment about this

src/less/day-slot.less
```diff
@@ -44,6 +44,30 @@
   &:nth-child(even){
     border-top-color: lighten(@cell-border, 10%);
   }
+
+  &.three {
```
jquense: I think that this can be handled via flex box styles for any slice amount, which would be better then needing hard coded values

src/components/DaySlot.jsx
cellog: no, this is the code that displays the selection div.  It's now external to the component to demonstrate how simple and easy it is to expose this in a controlled way so that users can extend it in unique ways

src/components/EventRow.jsx
cellog: this class is a placeholder that I hadn't worked on at all yet :).  It's fleshed out in another branch.  I implemented displaying events for week/day and month in separate branches to make pull requests simpler.  In this case, I would do the pull request to the streamline branch (this one) so you can see the difference more clearly.

src/components/TimeGrid.jsx
cellog: components is used by EventRow, and is required therefore.

src/components/TimeGrid.jsx
cellog: makes sense.  I left this code unchanged to minimize the changes to process.  Plus it already works.  What we really need is the new grid css, but oh well :)

src/components/TimeGridAllDay.jsx
cellog: The original code shares BackgroundCell between month and week.  And it does make sense: you can create multi-day all-day events in google calendar by selecting slots in the all-day gutter.  As I mentioned, I've fully implemented this for month, week and day in other code, so you can review that later

src/components/TimeSliceGroup.jsx
cellog: this is (was) used to pass a day-slot specific classname in.  In this commit: https://github.com/intljusticemission/react-big-calendar/pull/102/commits/daae7c84b6869022accd4b8cb6cd5c889f3def7b#diff-6c8b49f3b9d007e9ae15ba43ecc8029aL39 I removed the css in question.  Basically, for many slices, you get lines underneath the text, I thought it looked unclean, but we can talk about that later.

src/components/TimeSliceGroup.jsx
cellog: slices become invisible when you go above 5 per slot.  Makes selection very difficult to see

src/less/day-slot.less
cellog: That would be cool.  Not entirely sure how to implement that easily.

src/components/DaySlot.jsx
jquense: what is it formatting? if its date related then it more correctly should be a date format prop, which has a hook into the localizer. Also in case you missed you it a user can customize formatting already, as formats can also be functions.

src/components/EventRow.jsx
jquense: ok great, then lets remove this then for now if its not relevant to the PR

src/components/TimeGrid.jsx
jquense: they don't use Agenda, month, or toolbar components, so they should be specified here. At this point each component should be passed in individually as `eventComponent`, and so on. In any case any props from EventRow should already be covered in the spread of DaySlot and TimeGutter above

src/components/TimeSliceGroup.jsx
jquense: "visibility" should be a consequence of how tall the sections are which is controlled by User styling correct? I don't think it makes sense to pick hard cap here.

Moreso though, the comment is about he `10` number for step, when the original default was `30` can you talk about why it was changed?

src/components/TimeSliceGroup.jsx
jquense: if they aren't used then can we remove them? time slices should still get a prefixed `rbc-time-slice` class or similar regardless I think.

src/less/day-slot.less
jquense: `flex: 1` on the slices, and `display: flex` on the group.

src/components/TimeGrid.jsx
```diff
@@ -0,0 +1,239 @@
...
+       )
+
+      return (
+        <DaySlot
+          {...this.props }
+          min={dates.merge(date, min)}
+          max={dates.merge(date, max)}
+          now={new Date()}
+          eventTimeRangeFormat={this.props.eventTimeRangeFormat}
+          formatter={({start, end}) => {
```
jquense: this should be a bound method to avoid creating a new function on each render

src/components/TimeGrid.jsx
```diff
@@ -0,0 +1,239 @@
...
+           get(event, startAccessor),
+           get(event, endAccessor), 'day')
+       )
+
+      return (
+        <DaySlot
+          {...this.props }
+          min={dates.merge(date, min)}
+          max={dates.merge(date, max)}
+          now={new Date()}
```
jquense: `today` above already exists

src/components/TimeGrid.jsx
```diff
@@ -0,0 +1,239 @@
...
+    const addGutterRef = i => ref => this._gutters[i] = ref
+
+    const range = dates.range(this.props.start, this.props.end, 'day')
+    const { allDayEvents, rangeEvents } = this.separateEvents()
+
+    return (
+      <div className='rbc-time-view'>
+        <div className="rbc-time-header" ref={ref => this._headerCell = ref}>
+          <TimeGridHeader range={range}
+                          gutterRef={addGutterRef(0)}
```
jquense: fyi (not a big deal), sorry by align I ment like he components above in renderAllDayEvents, props should be indented once and all on a new line under the tag, not aligned to the first prop. Just for going forward

src/components/TimeGridHeader.jsx
```diff
@@ -0,0 +1,42 @@
...
+    onClick: () => null,
+    format: formats.dayFormat,
+    gutterRef: () => null
+  }
+
+  render() {
+    return (
+      <div className="rbc-row">
+        <div ref={this.props.gutterRef} className='rbc-gutter-cell' style={this.props.gutterWidth ?
+        {width: this.props.gutterWidth} : {}} />
```
jquense: unnecessary ternary, `{ width: this.props.gutterWidth }` is fine.

src/components/TimeGutter.jsx
```diff
@@ -0,0 +1,63 @@
...
+    max: PropTypes.instanceOf(Date).isRequired,
+    hidelabels: PropTypes.bool,
+    timesliceClassnames: PropTypes.string,
+    timeGutterFormat: PropTypes.string
+  }
+  static defaultProps = {
+    step: 10,
+    slices: 2,
+    selectable: false,
+    timesliceClassnames: '',
```
jquense: we don't expose classnames this way in other places do we? It feels a bit out of place to allow that here. If anything it should be a getter function to allow styling as well

src/components/TimeGutter.jsx
```diff
@@ -0,0 +1,63 @@
...
+    const numGroups = Math.ceil(totalMin / (this.props.step * this.props.slices))
+    const children = []
+    const groupLengthInMinutes = this.props.step * this.props.slices
+
+    let date = this.props.min
+    let next = date
+    let isNow = false
+
+    for (var i = 0; i < numGroups; i++) {
+      isNow = dates.inRange(this.props.now, date, dates.add(next, groupLengthInMinutes - 1, 'minutes'), 'minutes')
```
jquense: what's with the `- 1` here? Generally we treat all ranges as inclusive, meaning the range goes from 1 - 2 includes everything including 1 till 2 but not including 2

src/components/DaySlot.jsx
cellog: it is the text that is displayed inside the selection box.  A user may choose to render this (which I will do) as "15 minute major jury" instead of "10:15-10:30" using this callback

src/components/TimeGutter.jsx
cellog: if it is inclusive, then the rbc-now classname is added to 2 timeslicegroups at the same time.

src/components/TimeSliceGroup.jsx
cellog: changed back to 30.  Was 10 for no good reason

src/less/day-slot.less
cellog: implemented

src/components/TimeGrid.jsx
cellog: done

src/components/TimeGrid.jsx
cellog: used

src/components/TimeGrid.jsx
cellog: done

src/components/TimeGridHeader.jsx
cellog: done

src/components/TimeGutter.jsx
cellog: removed

cellog:
OK.  I've gone through your comments.  First things first, I have to say using flexbox for time slices is quite a bit too sexy for its shirt.  I like it :)

I've probably missed a few spots, but I've addressed the formatting issues and other minor stuff you mentioned.

The major question of selection code remains.  I've implemented selection properly for all the components (except agenda, where I assume, like google calendar, it doesn't do anything).  There are 3 branches that I have done this in.  1 is just the fixes to all day select (which doesn't work in this pull request).  I had to re-do the event creation to pop out absolutely positioned elements using top/left instead of flex-basis, so that the rbc-row elements wouldn't block mouse clicks from reaching the blank space.  I did the same thing for month view, but implemented a slightly different selection strategy which is both more efficient and smarter.

Then, a third branch combines both of the new codebases into a single one.  All of them are up to date with these changes and also your most recent merge of the PR.

I do have to challenge one assertion you made: google calendar does not implement selection the way you have.  If you click/drag on an event that is not draggable, google calendar ignores it.  You get normal browser selection behavior (it selects the text of the event link), and not what BigCalendar does, which is to start selecting the background cells.  My code captures that change, but as I said, the event generation code has to be modified for it to work as expected.  The changes are trivial in code length, and I actually added a couple tests before refactoring to make sure I didn't break the code when I changed it (npm run tdd is awesome for this work).

I assume you'll want examples updated as well?

But we need to talk about the selection code.  It's integral to the changes I've made.  I tried to find a way to extract out the exact code you've written in a way that it can fit with this code, and I got stuck.  It's too tightly integrated with the actual components.  I could keep trying, but I really believe the solution I've written is simpler.  Want me to open an issue where we can calmly explore the options?

jquense:
thanks for making those changes, looks good.

With regard to selection, I don't feel like starting to have another discussion now is going to be productive, in part because its still not clear that the original approach doesn't work for your use-cases, _and_ considering that the new approach has tradeoffs that are not acceptable for _our_ use cases. As a side point we use the original Selection code for a lot of different things internally and I disagree fairly strongly that its tightly coupled to anything, its not even coupled to React. 

I don't really care about the selection code being simpler, or more maintainable, or more general, it serves a specific use case well and is designed to be tunable for the use cases of a this particular calendar. It's not helpful to keep talking about changing it until there is a _good_ reason why it even needs to be changed, and right now we can't talk about that because the code is already removed and changed.

I want to avoid going in circles on this, From the two concerns you've raised about the current approach and your needs: One is a not a bug, and by design and at this point for our users a hard requirement, The second I feel can be addresses fairly easily using the original code. It hard to keep going back and forth about why that approach doesn't work, without more concrete reasons.

thanks for spending all this time working on making the library better, we do appreciate it.

cellog:
ok, thanks

jquense:
why close this? It would be a shame to not incorporate all the work you've done here, especially over a disagreement about an approach to code organization (if the selection stuff is the blocker)

cellog:
I closed it because I probably misread this: "With regard to selection, I don't feel like starting to have another discussion now is going to be productive, in part because its still not clear that the original approach doesn't work for your use-cases, and considering that the new approach has tradeoffs that are not acceptable for our use cases. As a side point we use the original Selection code for a lot of different things internally and I disagree fairly strongly that its tightly coupled to anything, its not even coupled to React."

I thought you were saying the selection stuff isn't going to happen, no matter what.

I think it would be most helpful for you to at least understand what is blocking using the existing selection code from my end.  The re-factor of all the time grid stuff is actually not possible unless the selection code is pulled out.  Otherwise, you can't re-use the same code, and have to duplicate a lot of it.  One week of my work was spent trying to figure out why this was so hard, and once I pulled the selection code out, it became easy.  I think it's harder to understand why I came to this conclusion until you sit down and grind at it, perhaps.  I should also note that I had an idea of sitting down, and pulling out the existing selection code into a HOC after I had done it, and I sat down, even created a branch to do it, but I couldn't find a way to do it without basically doing what I had already done.

Part of why this is true is that the only major implementation difference between the internals of the selection code I wrote and the existing code is where the onMouseDown handler is declared.  In the existing code, it is a global listener, which causes the selection code to fire whenever a mouse click begins anywhere over the containing element.  In the new code, I use onMouseDown on the container element, and use React's built-in event simulation, in order to limit the mousedown capturing to where it is needed.  Otherwise, the logic is exactly the same.  MouseUp and MouseMove are both captured globally so we find them even if the user strays outside the container element.  Changing it to use a global mousedown is not hard, and in fact, could be a simple (like 6 lines of code simple) change.  I just thought the behavior I observed was a bug, when I click-dragged on an event and it triggered the selection of the underlying time slot, but only triggered a slot select if I dragged outside the event, since Google calendar does not do this.  Is the drawback to your side this behavior?  Something else?  Adding this as a configuration option (globalMouseDown or something) enabled by default would preserve behavior.  Basically, imagine a scenario where drag and drop is enabled.  A user click-drags on an event, and at the same time, the slot underneath is selected.  Now, if the user drops the event on a new slot, what should happen?  With the existing code, won't we get both an event drop and an onSelectSlot triggered?

So for this to be productive, I need a clearer sense of requirements, or of things that cannot change.  Are there users who are delving into the internal API of big calendar and writing their selection code specifically using the Selection API?  I get the sense from your organization (International Justice) that your code is being used in serious, life-saving situations overseas, so I am not trying to impose my tiny needs from Nebraska onto the world.  What I'm looking for is more flexibility.

I want, for instance, to be able to display a week with 2 days, 5 slots per step so we can view midnight to midnight on a small screen, click drag to select 9-5, have the selection preserve, then click a few 10 minute slots, and an hour lunch break, and then save the work as a series of time slots.  Then I want to be able to take that data and format a time grid so that some time slices are frozen (the lunch breaks, etc.), and selection displays a 15-minute chunk sliding up and down, stopping at the boundaries.  Once I'm done, I click an external button that captures the date ranges (or in my case, each specific slot selected, ranges can be extrapolated if needed, but I don't actually need them).

The selection code I wrote will do the selection part of this with only a few lines of new code, using the preserveSelection mode.  The formatting stuff only works if the selection code is not tightly bound to the timegrid, because then I can use the formatter prop of DaySlot to display the selection div in a custom fashion without touching the DaySlot or TimeGrid code at all.  With the current solution, I have to subclass DaySlot, and then subclass TimeGrid to use it, and then I might as well subclass Week in order to pass all of these changes in, and then pass that to the Calendar.  All of which is possible, but is many more lines of code and makes maintenance much harder, testing nearly impossible because all the dependencies are more tightly coupled.  This is what I mean by tightly coupled.  It's not the old, crufty way of tightly coupling, but the fact that the logic of what each component does is blurred.  A TimeGrid handles both displaying the day slots and the selection of them, as well as the selection of all-day time slots.  Then, if you want to make changes to both (selection and the display), you have to subclass or replace all of the components that TimeGrid uses.

So, at this stage, because I don't know what tradeoffs I need to address, I can't make progress, and I don't wish to impose on you.  If you feel there is a chance our interests do align, I will try my best to accomodate the needs that I don't understand yet, as I have been so far.  I think we've worked pretty well.  No hard feelings if I just need to fork, I understand if what I am asking is too radical for your and your client's needs!

cellog:
a last note: I have not yet had time to pull the selection code out into its own react npm repo, but I still plan to do that.  But this is secondary priority.  Because this is MIT license, and it is based in part on your code, I will mention that in the LICENSE file as required.

One change I'm probably going to look at is instead of doing the calculation in mousemove, firing a timer that uses animation frames to do the selection calculation, and only storing current mouse position in the mousemove handler.  I used this approach inside a custom implementation of parallax on our website because scroll events are implemented so differently in every browser, and it was the only thing that worked in all the browsers smoothly.  But that's a separate question until the larger one is resolved.

jquense:
> The re-factor of all the time grid stuff is actually not possible unless the selection code is pulled out. Otherwise, you can't re-use the same code, and have to duplicate a lot of it. One week of my work was spent trying to figure out why this was so hard, and once I pulled the selection code out, it became easy.

This is the part I struggle with. I look at all the changes you've made to how the TimeGrid is structured and I do not see anything that requires a change to how Selection works. What specifically is the problem you are running into? It looks like the only change that would need to be made to DaySlot, is that `snapToSlot` needs to understand slices as well as slots.

jquense:
one more note:

> So for this to be productive, I need a clearer sense of requirements, or of things that cannot change. 

I agree with you on this, but it should be the other way around, there needs to be a strong reason _to_ change things, before we change things, because new code always has a cost. I appreciate that you see that there _are_ reasons to change them and that trying to describe code limitations is hard. 

From my side its also really difficult to see why code doesn't work for you when its been removed completely. I also hear you when you say the need code is functionally similar (or the same) as the old code which is great, however my impulse then in that case is "If its not much different then lets not change it at all, its not worth have new code". Also at a high level I hear what you are trying to do, but I am not sure what that means implementation wise. Are you planning on using all the internal components apart from the Calendar itself? That may be fine, but we'd have to weigh whether supporting that costs something for the main purpose of providing a mostly batteries included Scheduler component, or the cost of now vastly increasing the size of the public API. 

So we need to get more specific: 

- What does the new code allow that the old did not, and conversely Why not was tweaking the old code not tenable (specifics). 

I agree that it's tough to explain that stuff, but there is also no way for me to jump in and try and see what we are seeing, because all the old Selection code is gone.

cellog:
OK.  Here's the thing: you were completely right about the separation of selection.  The problem I had was figuring out an easy way to do it, and what I needed was a fresh codebase, in a separate window, that I could copy/paste work into.  I have done just that.

HOWEVER, there is a huge problem.  I already had a forked copy of this repo, and github won't let you fork to a different name for the repo, so instead I just imported the history from here.  This gave me a fresh checkout, but.... I can't do a pull request, because github has decided that the repos are not related.

You can see the code, but I have no idea how to get it into here.

https://github.com/cellog/react-big-calendar-try-again/commits/master

These 3 commits do 2 things:

1. setup react storybook so we can see how the changes work
2. modify the day-slot.less and make DaySlot use TimeGutter and TimeGutter use TimeSliceGroup and TimeSlice

It's simple, it works, no changes to selection code.  I have no idea why I couldn't do this before.  Probably an earlier design flaw in TimeSlice that made it impossible to work with the existing selection code.  If you have any ideas on how to get a fresh branch into my existing fork, that would be a way to solve it, I can copy/paste from the new repo into there once I have some clue how to do it.

cellog:
found this: http://jayunit100.blogspot.com/2015/07/github-you-dont-need-more-than-one-fork.html so I'm going to give it a try

cellog:
#107 got it!

cellog:
do you have an objection to my next project being simpler code cleanup, cosmetic stuff like moving to a modified airbnb eslint?

jquense:
I'd like to leave style stuff alone. One of the nice joys of my open source stuff is that I can do whatever, with styling. I have to follow airbnb's draconic style guides elsewhere in my life and it hurts my spirit, these repos are nice free-sprited breaks for me :), in general I am trying to just not care at all about non-bug related style issues (still getting better at that hence a few of my comments about it sorry).

We can certainly clean up the lint guide tho to make sure error-inducing things are present. though I draw the line as nit picky stuff stuff like `foo !== null || foo !== undefined` or `typeof foo === 'undefined'`

jquense:
btw the react-storybook thing is really awesome. I'm gonna use it on a bunch of other stuff, thanks for introducing that

cellog:
Ok. Can we make semicolons strict though? I prefer never, but if you want them always that's fine too

:)

via CloudMagic Email [https://cloudmagic.com/k/d/mailapp?ct=pa&cv=8.5.38&pv=5.1.1&source=email_footer_2] On Sun, Jun 05, 2016 at 11:41 AM, Jason Quense < notifications@github.com [notifications@github.com] > wrote:
I'd like to leave style stuff alone. One of the nice joys of my open source stuff is that I can do whatever, with styling. I have to follow airbnb's draconic style guides elsewhere in my life and it hurts my spirit, these repos are nice free-sprited breaks for me :), in general I am trying to just not care at all about non-bug related style issues (still getting better at that hence a few of my comments about it sorry).

We can certainly clean up the lint guide tho to make sure error-inducing things are present. though I draw the line as nit picky stuff stuff like foo !== null || foo !== undefined or typeof foo === 'undefined'

cellog:
https://www.npmjs.com/package/react-selection-hoc FYI.  I will work on this for a little bit, get it all tested up and also implement the exact mode of selection the package uses, then can we talk more?  I'll be able to quickly make demos in a new pull request of how it can make things more powerful.

cellog:
http://cellog.github.io/react-selection/ Live demos there, you can see the logic.  I've got saucelabs/browserstack set up, so after I can create some tests to verify that the thing works on every browser, and that I have full coverage, I'll come back here.

jquense:
site looks nice :)  I will take a look when I get a moment, stuff got a bit busy for me so I n8t as available at but I'm happy to open up some issues on the repo with things I think we'd need to use it

cellog:
fyi: just figured out basic touch support for selection.  Have to tweak for multi-finger and also make sure it doesn't disable scrolling, but this is pretty trivial!  So more incentive to use it in big calendar once I have the stuff tested.

jquense:
this is stale now right??

PR #103, opened 2016-05-30
Styling of a individual day cell in monthly view. 

dmoskovtsov:
screen shot 2016-05-30 at 3 08 37 pm


dmoskovsov:
Sorry, wrong repo.

jquense:
:wave: Hey there, thanks for the PR. Allowing styling of the background cell seems like a reasonable additional, however I think teh API here could be cleaner. The props getter should jsut be props that are spread into the component, whereas you have some magic properties like "header". For those special cases I think exposing a Header component makes more sense then the props getter

PR #107, opened 2016-06-05
unify DaySlot/TimeGutter

cellog:
Success!

This pull request unifies DaySlot and TimeGutter, using new TimeSliceGroup and TimeSlice.  It exposes new configurable properties on Calendar, "slices" and "now" for controlling the number of time slices, and what the calendar displays as today/right now.  It defaults to 2 slices and the current date/time for backwards compatibility.


jquense:
woohoo! that looks great, minimal changes but definitely better split up internally.


I just have a few comments and I think it's good to go

src/TimeGutter.jsx
```diff
@@ -1,39 +1,49 @@
-import React from 'react';
-import cn from 'classnames';
+import React, { Component, PropTypes } from 'react'
 import dates from './utils/dates';
-import localizer from './localizer'
+import TimeSliceGroup from './TimeSliceGroup.jsx'
 
-let TimeGutter = React.createClass({
+export default class TimeGutter extends Component {
```
jquense: I think now that this is used for all the large slots, `TimeGutter` is no longer a good name for this component. Perhaps TimeSegment, or TimeColumn (to go with TimeGrid?) makes more sense.

src/TimeGutter.jsx
```diff
@@ -1,39 +1,49 @@
...
+    now: PropTypes.instanceOf(Date).isRequired,
+    min: PropTypes.instanceOf(Date).isRequired,
+    max: PropTypes.instanceOf(Date).isRequired,
+    hidelabels: PropTypes.bool,
+    timeGutterFormat: PropTypes.string
+  }
+  static defaultProps = {
+    step: 10,
+    slices: 2,
+    selectable: false,
```
jquense: this isn't used now correct?

src/TimeGutter.jsx
```diff
@@ -1,39 +1,49 @@
...
+    hidelabels: false
```
jquense: Perhaps since the common usage case is no labels that should be the default and the prob be `showLabels` instead of the negative.

src/TimeGutter.jsx
```diff
@@ -1,39 +1,49 @@
...
+    step: PropTypes.number.isRequired,
+    slices: PropTypes.number.isRequired,
+    now: PropTypes.instanceOf(Date).isRequired,
+    min: PropTypes.instanceOf(Date).isRequired,
+    max: PropTypes.instanceOf(Date).isRequired,
+    hidelabels: PropTypes.bool,
+    timeGutterFormat: PropTypes.string
+  }
+  static defaultProps = {
+    step: 10,
```
jquense: I still think this is not correct here. `TimeGrid` specifies `30` as the default for step which overrides this. Perhaps we should make them the same default?

src/TimeSlice.jsx
```diff
@@ -0,0 +1,29 @@
...
+    selected: PropTypes.bool,
+    content: PropTypes.string,
+    culture: PropTypes.string
+  }
+
+  static defaultProps = {
+    isNow: false,
+    showlabel: true,
+    selected: false,
+    culture: 'en',
```
jquense: I wouldn't set a default here, in general we want the "default" to be the globally set culture...I realize that's leaky sounding from a component purity standpoint but its managed by the localizer and opaque to the component and or whatever reason that i18n libs decided enables the best interop :/

src/DaySlot.jsx
```diff
@@ -78,49 +81,52 @@ let DaySlot = React.createClass({
...
       start: this.state.startDate,
       end: this.state.endDate
     };
 
     return (
-      <div {...props} className={cn('rbc-day-slot', props.className)}>
-        { children }
-        { this.renderEvents(numSlots, totalMin) }
+      <div {...props} className={cn('rbc-day-slot', this.props.className)}>
+        <TimeGutter
```
jquense: I wish we could avoid the new extra div here...

One thought is to allow TimeGutter to accept children like:

```jsx
<TimeGutter>
  {renderEvents()}
  {renderSelection()}
</TimeGutter>
```

That avoids the extra div but may not be semantically correct, what do you think?

the reason I bring this up is that the extra div breaks the styling, causing the Gutters to not extpan to their "intrinsic height" which I think can also be fixed with some style tweaks as well if the extra div feels more correct

src/TimeSlice.jsx
```diff
@@ -0,0 +1,29 @@
...
+    content: '',
+  }
+
+  render() {
+    return (
+      <div className={cn('rbc-time-slot', this.props.isNow ? 'rbc-now' : '')} {...this.props.style}>
+        {this.props.showlabel ? <span>{this.props.content}</span> : ''}
```
jquense: The padding on the slice seems to have been lost for the label

src/TimeGutter.jsx
cellog: TimeColumn: check

src/TimeGutter.jsx
cellog: check

src/TimeGutter.jsx
cellog: I need to camelCase it though

jquense:
I should have it as a configured npm script but could you also run eslint on `src` it should revealed a left over few things to clean up as well

src/less/day-slot.less
```diff
@@ -30,20 +31,25 @@
...
   }
+}
 
-  &:nth-child(even){
-    border-top-color: lighten(@cell-border, 10%);
-  }
+.rbc-timeslice-group {
+  border-bottom: 1px solid @cell-border;
+  min-height: 40px;
+  height: 40px;
```
jquense: the height may not be necessary here... at least not with min height.  actually does this make more sense on TimeSlice  vs the group?

src/TimeGutter.jsx
cellog: I also think we should rename the DaySlot and TimeSlice and TimeSliceGroup to DayColumn, TimeSlot, TimeSlotGroup

src/TimeGutter.jsx
cellog: check

src/TimeGutter.jsx
cellog: check

src/TimeSlice.jsx
cellog: check

src/TimeSlice.jsx
cellog: gonna tackle this before the previous problem (unless they are related, which is likely)

cellog: fixed.  Was css issue

src/DaySlot.jsx
cellog: sure, let me take a peek at a solution

src/less/day-slot.less
```diff
@@ -30,20 +31,29 @@
...
+
+  .rbc-timeslice-group {
+    padding: 0;
+  }
 }
 
 .rbc-time-slot {
   flex: 1 0 0;
-  min-height: 20px;
-  height: 20px;
```
cellog: this was where the 40px 40px comes from.  rbc-time-slot was 20px, and we want the height of the group to equal 2 of these.  I agree min-height and height is a weird redundancy, but it came from here.  No problem correcting it if you want, just moving to height

src/less/day-slot.less
cellog: this is taking the css from rbc-time-slice, which was min-height/height: 20px (see comment below).  If we do height on time-slice, then multiple slices breaks.  We flex the height now with the new commit

src/DaySlot.jsx
cellog: fixed

cellog:
try it now, all the issues you mentioned are fixed.  I am going to rename the cosmetic DaySlot and so on.

One question: css is still based on old component names, can we make aliases to the new component type and have the old ones just for backwards compatibility?  so rbc-time-slot instead of rbc-time-slice, rbc-time-column instead of rbc-time-gutter, etc.

cellog:
oh and both docs and examples need to be updated once all the code is approved before you merge the commit, I can do that easily.

cellog:
I think there is a bug in the eslintrc.  The semi rule is turned off, but set to never, I think it should be either 1 or 2 for warning or error?

    "semi": [0, "never"],

I noticed semi-colons are inconsistent, probably because of this.  Whatever you want though

cellog:
didn't find any other eslint issues

src/less/day-slot.less
jquense: What do you envision the behavior of the slice to be? seems like there are two options, 

1. what is there now which is min heights on the Group which shrinks slices proportionally
2. height on the slice which makes them scale bit better, but groups expand in height without bound.

I am not sure which makes the most sense honestly. It does seem like setting 20px min height on the slices and let the group have whatever height would be a good default as well. it sort of depends on how you envision the slices.

src/TimeColumn.jsx
```diff
@@ -0,0 +1,75 @@
...
+  render() {
+    const totalMin = dates.diff(this.props.min, this.props.max, 'minutes')
+    const numGroups = Math.ceil(totalMin / (this.props.step * this.props.slices))
+    const timeslots = []
+    const groupLengthInMinutes = this.props.step * this.props.slices
+    let baseCss
+    
+    switch (this.props.type) {
+      case 'gutter' :
+        baseCss = 'rbc-time-gutter'
```
jquense: any reason why we can't let the parent DaySlot add the `rbc-day-slot` class? I don't think the conflict with each other so having both might not matter in that case. It's a bit less magical then the `type`

src/less/day-slot.less
cellog: The entire point of the slices was to shrink them, so that you can fit an entire day onto a single page without scrolling, to allow selecting a large time chunk.  Otherwise you have to change the selection code to scroll the page when the mouse gets to the bottom.  To do the behavior you're describing, all we have to do is change the step value.  The slice (or time slot, as I'm about to rename it) value is about making it tinier.  This is one of the simplybook.me-inspired features.

src/TimeSlice.jsx
```diff
@@ -0,0 +1,28 @@
...
+  static defaultProps = {
+    isNow: false,
+    showLabel: true,
+    selected: false,
+    content: ''
+  }
+
+  render() {
+    return (
+      <div className={cn('rbc-time-slot', this.props.isNow ? 'rbc-now' : '')} {...this.props.style}>
```
jquense: `cn('rbc-time-slot', this.props.isNow && 'rbc-now')` works as well here

src/TimeColumn.jsx
cellog: that may be the answer, yes

cellog: Nope, tried this, broke everything.  rbc-time-gutter and rbc-day-slot conflict :)

jquense: ok then thats fine for now, I'll take a stab at cleaning up the styles after

src/less/day-slot.less
jquense: ahh ok. So there is a bit of a conflict inherent in that, since some also would rather that the days not shrink below a certain size and scroll beyond that. We should make sure then we have css defaults that are easy to adjust/override for folks that want either behavior.

In your case getting everything to fit on a page should be a simple matter of setting the calendar height to the page size and the flexbox styles could handle everything from there for all the inner components. as it is that won't work without some adjustments up the line... in any case I can tweak that stuff later, nothing to be done here now.

cellog: I think I'm being unclear. :)

If you haven't already, run npm run storybook and browse to localhost:9002 
then check out the examples with 3 and 9 slots.

Then click week view, and select as much as you can. The default selectable 
view allows you to make a slot of at most about 15.5 hours, depending on 
screen size. The 3 slots one allows selecting about 23 hours, the 9 slots 
one allows selecting the whole thing with about half the page representing 
24 hours.

The key thing to remember is that the step is preserved, the only thing 
that changes is the visual representation. I will use this to have 5 minute 
step and still visualize 9 to 6 in a single page.


via CloudMagic Email 
[https://cloudmagic.com/k/d/mailapp?ct=pa&cv=8.5.38&pv=5.1.1&source=email_footer_2]

jquense: o wow, you are right i did not understand, or notice :P 

If this is about fitting an entire day on a page there were deeeeefinately quicker ways to handle that. Not that this isn't helpful in and of itself. Do you really want the labels to be less frequent or is that a side effect of shrinking?

also let me branch your branch so you can see waht I mean...

jquense: This is what I meant: https://github.com/intljusticemission/react-big-calendar/tree/full-height

image

image

cellog: I just added 3 new storybook stories to show how I envisioned the thing working.  The purpose is to take the same space that has 2 30 minute selectable time slots, and render it with different lengths.  So, for instance, you change the step to 10, and use 6 time slots, or you change the step to 20 and have 3 time slots.  The visual representation of an hour is the same for each.

jquense: (fyi i'm not suggesting this PR is not needed, just that I misunderstood it a bit :P you've opened up a bunch of fun new style options)

jquense: I think we are on the same page, I would just say that at a minimum the columns should extend the  heightof the calendar so we don't have have a mostly empty bordered box, Whether or not the columns "grow" the calendar itself or overflow with a scrollbar is up to the user and whether they apply min-heights on the slices/groups. From a default perspective tho I think the tweaks in my branch (minus the removal of the min-height) should be reasonable

cellog: Gotcha. So we use simple flex for the time slice groups and let the Calendar box size them, with a min-height of 40 and no max height

via CloudMagic Email [https://cloudmagic.com/k/d/mailapp?ct=pa&cv=8.5.38&pv=5.1.1&source=email_footer_2] On Sun, Jun 05, 2016 at 1:48 PM, Jason Quense < notifications@github.com [notifications@github.com] > wrote:
In src/less/day-slot.less [https://github.com/intljusticemission/react-big-calendar/pull/107#discussion_r65822319] :

>  
> -  &:nth-child(even){
> -    border-top-color: lighten(@cell-border, 10%);
> -  }
> +.rbc-timeslice-group {
> +  border-bottom: 1px solid @cell-border;
> +  min-height: 40px;
> +  height: 40px;


I think we are on the same page, I would just say that at a minimum the columns should extend the heightof the calendar so we don't have have a mostly empty bordered box, Whether or not the columns "grow" the calendar itself or overflow with a scrollbar is up to the user and whether they apply min-heights on the slices/groups. From a default perspective tho I think the tweaks in my branch (minus the removal of the min-height) should be reasonable

src/Day.jsx
```diff
@@ -7,6 +7,14 @@ let Day = React.createClass({
 
   propTypes: TimeGrid.propTypes,
 
+  getDefaultProps() {
+    return {
+      now: new Date(),
+      type: 'gutter',
+      timeslots: 2
+    }
```
jquense: To me its a big confusing to define defaults for props that the component doesn't explicitly use. Its makes it a bit harder to understand which value will be used ultimately and harder to change the default in the right spot.

Can we remove them for these components?

src/TimeSlot.jsx
```diff
@@ -0,0 +1,28 @@
...
+    isNow: PropTypes.bool,
+    showLabel: PropTypes.bool,
+    selected: PropTypes.bool,
+    content: PropTypes.string,
+    culture: PropTypes.string
+  }
+
+  static defaultProps = {
+    isNow: false,
+    showLabel: true,
```
jquense: can we make `showLabel` have no default (or default `false`) since its the more common value.

src/TimeSlotGroup.jsx
```diff
@@ -0,0 +1,51 @@
...
+    culture: PropTypes.string
+  }
+  static defaultProps = {
+    timeslots: 2,
+    step: 30,
+    isNow: false,
+    showLabels: true
+  }
+
+  renderSlice(i, content, value) {
```
jquense: `i` is a bit unclear here for me, can we change it to `isFirst` or something and coerce the `0` to `true in `renderSlices`

src/TimeSlotGroup.jsx
```diff
@@ -0,0 +1,51 @@
...
+    showLabels: PropTypes.bool,
+    isNow: PropTypes.bool,
+    timeGutterFormat: PropTypes.string,
+    culture: PropTypes.string
+  }
+  static defaultProps = {
+    timeslots: 2,
+    step: 30,
+    isNow: false,
+    showLabels: true
```
jquense: same as above, default to `false`

src/TimeSlotGroup.jsx
```diff
@@ -0,0 +1,51 @@
...
+    step: PropTypes.number.isRequired,
+    value: PropTypes.instanceOf(Date).isRequired,
+    showLabels: PropTypes.bool,
+    isNow: PropTypes.bool,
+    timeGutterFormat: PropTypes.string,
+    culture: PropTypes.string
+  }
+  static defaultProps = {
+    timeslots: 2,
+    step: 30,
```
jquense: these two props are required, so a default is unnecessary

src/Week.jsx
```diff
@@ -9,6 +9,14 @@ let Week = React.createClass({
 
   propTypes: TimeGrid.propTypes,
 
+  getDefaultProps() {
+    return {
+      now: new Date(),
+      type: 'gutter',
+      timeslots: 2
```
jquense: ditto the extra defaults

src/TimeGrid.jsx
```diff
@@ -44,7 +44,9 @@ let TimeGrid = React.createClass({
     return {
       step: 30,
       min: dates.startOf(new Date(), 'day'),
-      max: dates.endOf(new Date(), 'day')
+      max: dates.endOf(new Date(), 'day'),
+      type: 'gutter',
+      now: new Date()
```
jquense: extra and unused defaults

jquense:
Noted a last few nits, can you also squash the PR down and then we should be good to go

src/Day.jsx
cellog: This is the problem with re-using attributes from other components.  These properties are absolutely required for TimeGrid to function.  Thus, when clicking the buttons week/day in the toolbar, there are a slew of warnings that these proptypes are not present, and in fact the calendar fails to display the current time unless this is present.

2 choices (yours to make):

1. also use TimeGrid's default props
      getDefaultProps() {
        return TimeGrid.defaultProps
      }

2. declare these properties on the Calendar component with the proper defaults

src/TimeSlotGroup.jsx
cellog: done

src/TimeSlotGroup.jsx
cellog: done

src/Day.jsx
jquense: Ahh right. I believe in that case it makes sense to not mark them as required (since they are actually provided), and provide the defaults at the lowest level they are needed. That shouldn't trigger warning right?

second choice for me would be 1, since it mirrors how we spread propsTypes

src/TimeSlot.jsx
cellog: done

src/Day.jsx
cellog: I would like to go with the 2nd choice.  The reason is that even though you don't wish to document the internals so we don't run into backwards compatibility hell, if anyone does need to make a change, it's much harder to do it unless all the needed props are documented and required ones explicitly marked as so.  This is why I always mark props that are absolutely needed for the component to function as required, even if they have a default value supplied.  It's redundant, but this stuff is only checked in development anyways, so I think of it as another layer of documentation for the developer.  So is it cool to use the default props for that reason?

src/TimeSlotGroup.jsx
cellog: my explanation above for why the redundant required/default value is present hopefully covers this?

src/TimeGrid.jsx
cellog: this is a special case: I was using ...DaySlot.defaultProps and there was some weird stuff going down.  Occasionally it just crashed, so I removed the spread and pulled in the defaults we really needed.  Probably an obscure React bug

src/Day.jsx
jquense: fine with me 👍

src/Week.jsx
cellog: done

src/TimeSlotGroup.jsx
cellog: done

src/Day.jsx
cellog: done

src/TimeGrid.jsx
jquense: ok then please, note that with a comment, so someone else doesn't remove them thinking they are redundant.

cellog:
I think we're in syzygy now...?

src/Week.jsx
```diff
@@ -10,11 +10,7 @@ let Week = React.createClass({
   propTypes: TimeGrid.propTypes,
 
   getDefaultProps() {
-    return {
-      now: new Date(),
-      type: 'gutter',
-      timeslots: 2
-    }
+    return TimeGrid.defaultProps
```
jquense: in case you didn't know, you could have done `TimeGrid.getDefaultProps()` I believe. switching to a class is fine too

src/TimeGrid.jsx
```diff
@@ -196,16 +194,16 @@ let TimeGrid = React.createClass({
         </a>
       </div>
     )
-  },
+  }
 
   _headerClick(date, e){
```
jquense: these now need to be bound to the class instance, or they will break

jquense:
just the one last note about the event handlers that break when switching to es classes

cellog:
ok, ready

jquense:
oops it looks like you missed the _headerClick handler as well.

jquense:
(accidental close)

jquense:
👏  thank you!

PR #108, opened 2016-06-06
Fix dayHeaderFormat when using Moment.js

imjoehaines:
In Globalize `dd` is the numeric day of the month but in Moment it's the [short weekday](http://momentjs.com/docs/#/displaying/format/), meaning that the day header was rendered as 'Monday May Mo' instead of 'Monday May 01' when using Moment.


jquense:
thanks!

PR #110, opened 2016-06-09
Upgrade lodash to 4.x

avindra:
other: upgrade eslint* packages, and activate react/recommended


jquense:
hi there, thanks for the PR. There are a few issues here. The first is just that this PR does a lot of different unrelated things, some of which break structuring that is intentional.

In terms of upgrading lodash, we've had this conversation before, and for right now i am selfishly going to pass because internally the stuff that consumes lodash is using v3, and until v3 isn't supported I don't feel like upgrading it offers anything

.eslintrc
```diff
@@ -1,15 +1,15 @@
 {
   "parser": "babel-eslint",
-  "extends": "eslint:recommended",
+  "extends": [
+    "eslint:recommended",
+    "plugin:react/recommended"
```
jquense: I wouldn't just turn on new lint rules, especially since no code has been changed to ensure the existing code doesn't fail.

examples/Api.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import metadata from 'component-metadata!react-big-calendar/Calendar';
-import transform from 'lodash/object/transform';
+import { transform } from 'lodash/object';
```
jquense: this will import all lodash.object, whereas the original code only imports transform, so the bundle size is going to increase dramatically here.

.eslintrc
avindra: I ran the linter before I made the PR and everything was passing with `plugin:react/recommended`, so I saw no harm in enabling it.

examples/Api.js
avindra: You are correct about the larger bundle size. I would suggest including [babel-plugin-lodash](https://github.com/lodash/babel-plugin-lodash) (an officially supported plugin), which automatically modularizes the build size, regardless of whether using the long, manual includes, or the ES6 style, destructured imports that I switched to in this PR.

avindra:
Re: v3 vs v4: please consider upgrading as soon as possible, as many projects have already migrated to v4, and those of us wanting to use `react-big-calendar` would end up with 2 versions of lodash in our projects.

examples/Api.js
jquense: That's an awful lot of work to avoid writing `import transform from 'lodash/object/transform';` while also being less clear.

jquense:
> wanting to use react-big-calendar would end up with 2 versions of lodash in our projects.

This is not true, you would end up with maybe two versions of the 4 functions rbc uses, since we only import those parts. That is the whole advantage to using the individual imports.

examples/Api.js
avindra: The destructured imports are a lot clearer because:

1. They are more terse
2. `import { something } from 'package'` would immediately cause your code to break if `something` doesn't exist, whereas `import somethingThatDoesntExist from 'package'` would be more of a silent failure somewhere in your code.

I don't consider adding one dev dependency and one plugin to the `babelrc` "an awful lot of work" for such a clear advantage.

avindra:
Correct... chances are you'd end up with duplicated code, especially with functions like `mapValues`, `chunk` or `omit`.

PR #114, opened 2016-06-09
remove `.jsx` extension from import

jdeniau:
Fixes #113 ? (I did not test but it definitively fixes that :) )

PR #116, opened 2016-06-10
Get rid of .jsx

### LINKED ISSUE #113: ERROR in ./~/react-big-calendar/lib/TimeColumn.js
Hi, since the 0.10 update, I got this error.

```
ERROR in ./~/react-big-calendar/lib/TimeColumn.js
Module not found: Error: Cannot resolve 'file' or 'directory' ./TimeSlotGroup.jsx in .../my-project/node_modules/react-big-calendar/lib
 @ ./~/react-big-calendar/lib/TimeColumn.js 17:21-51
```

I did not try (for now) to implement custom views, I just made the update.

It may come from #107

antoinerousseau:
Fixes #113 (but really this time — follow up of #114)


jdeniau:
+1 for removing the import with jsx, but I think the jsx files should stay jsx files (as they are not JS compatible)

antoinerousseau:
Well, there are some debates about it, but I saw some `.js` files in the project too so I though it would be better to not mix... what do you prefer @jquense?

I personally use `.js` everywhere in my project, since it all goes through Babel.

https://www.reddit.com/r/reactjs/comments/4kkrwg/ask_js_or_jsx_extension/

jquense:
let's just remove the extension from the requires, I like keeping the jsx extension for clarity as to what they are. thanks!

jquense:
i got it :) https://github.com/intljusticemission/react-big-calendar/commit/741c88238c00fabb23be47af3cd4a832aa48cc97

PR #120, opened 2016-06-16
allow a custom popup component and use the right event component in default popup

craig-day:
cc @jquense 

In the _Show X More_ popup, it was trying to pull `eventComponent` from props but that was never passed in.  Lets make sure it gets passed in.

### DIFF
--- a/src/Month.jsx
+++ b/src/Month.jsx
@@ -287,6 +287,7 @@ let MonthView = React.createClass({
 
   _renderOverlay(){
     let overlay = (this.state && this.state.overlay) || {};
+    let { components } = this.props;
 
     return (
       <Overlay
@@ -298,6 +299,7 @@ let MonthView = React.createClass({
       >
         <Popup
           {...this.props}
+          eventComponent={components.event}
           position={overlay.position}
           events={overlay.events}
           slotStart={overlay.date}


jquense:
Hi there! thanks for the PR, I don't think we want to expose a Popover component here, it would couple folks to the specific popover and overlay API that we are using internally...instead I think it makes sense to turn off the popover and insert your own logic `onShowMore`

src/Popup.jsx
```diff
@@ -51,7 +51,7 @@ class Popup extends React.Component {
             <EventCell key={idx}
               {...props}
               event={event}
-              component={eventComponent}
+              component={components.event}
```
jquense: we don't actaully want to pass the `components` prop through to this component, Instead can we fix why eventComponent wasn't being sent correctly?

craig-day:
@jquense thanks for the feedback. I changed this to be a more narrow fix and just ensure thate `eventComponent` gets passed to the popup.

TMUND:
@jquense I'm setting popover to false and trying to use `onShowMore` to display a custom popover component but it's not registering my click at all. Any advice on how to show a custom popover when "+# more" is clicked?

palashkaria:
Hey @jquense, you mentioned here that we can use our own logic for onShowMore. Is there a prop exposed for this or did you mean forking the repo? I have the same issues as @TMUND

PR #121, opened 2016-06-19
Documentation: 'step' prop

Commander-lol:
Just adds the `step` prop from `src/TimeGrid` to the propTypes of `src/Calendar` so that it appears in the documentation - it's a super useful prop to know about. Sets a default value of 30, much like in `src/TimeGrid` since half hour increments seem to be the default elsewhere in the code.


jquense:
thanks!

PR #122, opened 2016-06-27
Added scrollTime prop like in FullCalendar

manutenfruits:
I added a new prop scrollTime so that the TimeGrid view is initially scrolled down to a specified hour on every mount or change of prev/next week. Let me know if I need to add any checks!

Original prop: http://fullcalendar.io/docs/agenda/scrollTime/

Takes a Date object, which I don't find really useful (a moment.duration would be more suitable, or an amount in millis, but I get how this is moment-agnostic) but props `min` and `max` used that too so I figured I'd use the same method :)

### DIFF
--- a/examples/demos/selectable.js
+++ b/examples/demos/selectable.js
@@ -14,6 +14,7 @@ let Selectable = React.createClass({
           selectable
           events={events}
           defaultView='week'
+          scrollToTime={new Date(1970, 1, 1, 17)}
           defaultDate={new Date(2015, 3, 1)}
           onSelectEvent={event => alert(event.title)}
           onSelectSlot={(slotInfo) => alert(
--- a/examples/demos/selectable.js
+++ b/examples/demos/selectable.js
@@ -14,7 +14,7 @@ let Selectable = React.createClass({
           selectable
           events={events}
           defaultView='week'
-          scrollToTime={new Date(1970, 1, 1, 17)}
+          scrollToTime={new Date(1970, 1, 1, 6)}
           defaultDate={new Date(2015, 3, 1)}
           onSelectEvent={event => alert(event.title)}
           onSelectSlot={(slotInfo) => alert(
--- a/src/Calendar.jsx
+++ b/src/Calendar.jsx
@@ -218,9 +218,14 @@ let Calendar = React.createClass({
     min: PropTypes.instanceOf(Date),
 
     /**
-     * Constrains the maximum _time_ of the Day and Week views..
+     * Constrains the maximum _time_ of the Day and Week views.
      */
     max: PropTypes.instanceOf(Date),
+
+    /**
+     * Determines how far down the scroll pane is initially scrolled down.
+     */
+    scrollToTime: PropTypes.instanceOf(Date),
 
     /**
      * Localizer specific formats, tell the Calendar how to format and display dates.
--- a/src/TimeGrid.jsx
+++ b/src/TimeGrid.jsx
@@ -64,7 +64,7 @@ export default class TimeGrid extends Component {
 
   componentWillMount() {
     this._gutters = [];
-    this.adjustScroll();
+    this.calculateScroll();
   }
 
   componentDidMount() {
@@ -91,10 +91,10 @@ export default class TimeGrid extends Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    const { start, scrollTop } = this.props;
+    const { start, scrollToTime } = this.props;
     // When paginating, reset scroll
-    if (!dates.eq(nextProps.start, start) || nextProps.scrollTop !== scrollTop) {
-      this.adjustScroll();
+    if (!dates.eq(nextProps.start, start) || nextProps.scrollToTime !== scrollToTime) {
+      this.calculateScroll();
     }
   }
 
@@ -295,7 +295,7 @@ export default class TimeGrid extends Component {
     }
   }
 
-  adjustScroll() {
+  calculateScroll() {
     const { min, max, scrollToTime } = this.props;
 
     const diffMillis = scrollToTime - dates.startOf(scrollToTime, 'day');
--- a/src/TimeGrid.jsx
+++ b/src/TimeGrid.jsx
@@ -73,6 +73,7 @@ export default class TimeGrid extends Component {
     if (this.props.width == null) {
       this.measureGutter()
     }
+    this.applyScroll();
   }
 
   componentDidUpdate() {
@@ -80,13 +81,7 @@ export default class TimeGrid extends Component {
       this.measureGutter()
     }
 
-    if (this._scrollRatio) {
-      const { content } = this.refs;
-      content.scrollTop = content.scrollHeight * this._scrollRatio;
-      // Only do this once
-      this._scrollRatio = null;
-    }
-
+    this.applyScroll();
     //this.checkOverflow()
   }
 
@@ -295,6 +290,15 @@ export default class TimeGrid extends Component {
     }
   }
 
+  applyScroll() {
+    if (this._scrollRatio) {
+      const { content } = this.refs;
+      content.scrollTop = content.scrollHeight * this._scrollRatio;
+      // Only do this once
+      this._scrollRatio = null;
+    }
+  }
+
   calculateScroll() {
     const { min, max, scrollToTime } = this.props;
 
@@ -309,7 +313,7 @@ export default class TimeGrid extends Component {
 
     let isOverflowing = this.refs.content.scrollHeight > this.refs.content.clientHeight;
 
-    if (this.setState.isOverflowing !== isOverflowing) {
+    if (this.state.isOverflowing !== isOverflowing) {
       this._updatingOverflow = true;
       this.setState({ isOverflowing }, () => {
         this._updatingOverflow = false;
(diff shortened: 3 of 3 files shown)


.gitignore
```diff
@@ -30,3 +30,7 @@ node_modules
 
 # Webstorm ide files
 .idea
+
+# Example bundles
```
jquense: these are actually meant to be committed, they are for the docs

manutenfruits: Oops make sense, I'll add them then

src/TimeGrid.jsx
```diff
@@ -77,9 +80,23 @@ export default class TimeGrid extends Component {
...
+      // Only do this once
+      this._scrollRatio = null;
+    }
+
     //this.checkOverflow()
   }
 
+  componentWillReceiveProps(nextProps) {
+    // When paginating, reset scroll
+    if (+nextProps.start !== +this.props.start) {
```
jquense: what about if the scrollTime changes?

.gitignore
manutenfruits: On a second note, I think you can generate them in `npm preinstall` but we can leave that for later.

src/Calendar.jsx
```diff
@@ -218,11 +218,16 @@ let Calendar = React.createClass({
...
     /**
-     * Constrains the maximum _time_ of the Day and Week views..
+     * Constrains the maximum _time_ of the Day and Week views.
      */
     max: PropTypes.instanceOf(Date),
 
     /**
+     * Determines how far down the scroll pane is initially scrolled down.
+     */
+    scrollTime: PropTypes.instanceOf(Date),
```
jquense: this is a bit oddly named (to my mind) maybe something like scrollToTime or just scrollTop?

src/TimeGrid.jsx
manutenfruits: hm... what I want to prevent is an unexpected scroll change when the user has already scrolled down somewhere else, so if you think it's important or that `scrollTime` can change in between renders, I can scroll to the new position if the user hasn't scrolled at all. Although sounds to me like a lot of unneeded complexity.

src/Calendar.jsx
manutenfruits: I copied the same name from FullCalendar, but happy to change it! How about `defaultTimeScroll`?

src/TimeGrid.jsx
jquense: I don't think you need all that complexity, if a dev wants to programmatically scroll the component then I think we want to just let that happen, it's up to them to not make that annoying for their users. if you want the current behavior then leave the prop static and no problems. 

my. main concern is that "on mount" is sometimes not a good time to initialization stuff with regard to layout. for instance if the calendar is hidden when mounted you can't do any scrolling.

manutenfruits: Good point about the hidden calendar. What do you suggest? Just skip this `if` or check if `scrollTime` has changed too?

jquense: I'd only set the scroll if the start time changed _or_ scrollTime changed. 

as an aside I'd also use `dates.eq` using the 'day' unit for the start comparison as well so we don't switch scroll if milliseconds change for instance

src/TimeGrid.jsx
```diff
@@ -277,6 +294,15 @@ export default class TimeGrid extends Component {
     }
   }
 
+  adjustScroll() {
+    const { min, max, scrollTime } = this.props;
+
+    const diffMillis = scrollTime - dates.startOf(scrollTime, 'day');
```
jquense: what if ScrollTime is undefined? probably should return early if so to avoid extra work or possible null exceptions

manutenfruits: I set it on `getDefaultProps` to be the beginning of the day

manutenfruits:
@jquense I fixed all the issues you pointed out, let me know if there's anything else that needs work done

examples/demos/selectable.js
```diff
@@ -14,6 +14,7 @@ let Selectable = React.createClass({
           selectable
           events={events}
           defaultView='week'
+          defaultTimeScroll={new Date(1970, 1, 1, 17)}
```
jquense: this isn't a better name I don't think. the "default" version of props also have a non default counterpart (this does not). Default props also imply that the prop is only set once which isn't true in this case. can we use scrollToTime instead?

manutenfruits:
Fixed. I also squashed it all under one commit.

src/TimeGrid.jsx
```diff
@@ -77,9 +80,24 @@ export default class TimeGrid extends Component {
...
+      this._scrollRatio = null;
+    }
+
     //this.checkOverflow()
   }
 
+  componentWillReceiveProps(nextProps) {
+    const { start, scrollTop } = this.props;
+    // When paginating, reset scroll
+    if (!dates.eq(nextProps.start, start) || nextProps.scrollTop !== scrollTop) {
```
jquense: `dates.eq(nextProps.start, start, 'day')` and probably the same for: `nextProps.scrollTop !== scrollTop` no? Well not `'day'` but perhaps `'minutes'`, ie only recalculate with the scrollToTime has changed more than a minute.

Also `scrollTop` is wrong here right?

src/TimeGrid.jsx
```diff
@@ -77,9 +80,24 @@ export default class TimeGrid extends Component {
...
+      this.adjustScroll();
```
jquense: This doesn't actually adjust the scroll, only calculate the ratio. After componentDidMount the scroll position will never be updated again.

src/TimeGrid.jsx
```diff
@@ -77,9 +80,24 @@ export default class TimeGrid extends Component {
       this.measureGutter()
     }
 
+    if (this._scrollRatio) {
+      const { content } = this.refs;
+      content.scrollTop = content.scrollHeight * this._scrollRatio;
+      // Only do this once
+      this._scrollRatio = null;
```
jquense: ?

jquense:
ok thanks for the updates, I finally had a chance to give it a real close look, just a few concerns

src/TimeGrid.jsx
manutenfruits: yes! when hitting next/previous it should scroll again

manutenfruits: yes, because the scroll needs to be updated on `componentDidUpdate`, when it's all flushed to the DOM. So I calculate a new scrollRatio property in the object and apply it once rendered.

src/TimeGrid.jsx
manutenfruits: This clears the _scrollRatio property so it doesn't get applied on the next render

src/TimeGrid.jsx
manutenfruits: `scrollTop` wrong, yes. The condition, however is correct, because it only calculates the new scroll value when the start is different (i.e. it has changed dates or paginated) or when the `scrollToTime` prop has changed.

src/TimeGrid.jsx
manutenfruits: Also, I'm gonna rename it to `calculateScroll` if that's any better

src/TimeGrid.jsx
jquense: that won't happen though since `componentDidMount` is only ever called once over the lifecycle of the component.

FYI you don't have any `componentDidUpdate` call here

manutenfruits: so the `calculateScroll` function is called from `componentDidMount` (which of course is only called once) and `componentDidUpdate` which gets called every time something is rendered, but is not shown on the diff, but [it's there](https://github.com/intljusticemission/react-big-calendar/pull/122/files#diff-fe4a2572cc6abfc6a7df5455b234527eL75).

jquense: lol tricky diff

jquense: Ok though the problem is still there, you need to set the scroll position in _both_ right?

manutenfruits: I calculate the new position and set it on [`componentWillMount`](https://github.com/intljusticemission/react-big-calendar/pull/122/files#diff-fe4a2572cc6abfc6a7df5455b234527eR67) and [`componentWillReceiveProps`](https://github.com/intljusticemission/react-big-calendar/pull/122/files#diff-fe4a2572cc6abfc6a7df5455b234527eR97). Then later, on [`componentWillUpdate`](https://github.com/intljusticemission/react-big-calendar/pull/122/files#diff-fe4a2572cc6abfc6a7df5455b234527eR83) (I can't do it before because changes haven't been flushed to the DOM) apply the actual scroll. What do you mean by both? `componentWillUpdate` gets called on all cases. Also I applied this to the `selectable` example so you can test it working youself.

jquense: You ONLY update the the actual scroll position on the DOM node in `componentDidUpdate` you need to update it it both cDU and `componentDidMount`. That it's working now is only because the component is triggering a few updates, but its not actually setting the scroll position after the component mounts, only calculating it.

manutenfruits: [you're right](https://facebook.github.io/react/docs/component-specs.html#updating-componentdidupdate) does not get called for initial render. I'll add it there too.

jquense:
many thanks

manutenfruits:
Thank you for creating this plugin! Are you planning a release anytime soon?

PR #126, opened 2016-06-29
Fixes #74 - Use localized time range format

### LINKED ISSUE #74: moment localizer ignores culture setting on time ranges
See [Lines #15 #16 on moment.js](//github.com/intljusticemission/react-big-calendar/blob/master/src/localizers/moment.js#L15). 'h.mma' is always used and and always produces AM/PM regardless culture setting

krokofant:
Use localized time format with MomentJS and trim AM/PM only if end is within 12h from start and AM/PM is in used by the localized result.

### DIFF
--- a/src/localizers/moment.js
+++ b/src/localizers/moment.js
@@ -8,12 +8,16 @@ function inSame12Hr(start, end){
   return (s <= 0 && e <= 0) || (s >= 0 && e >= 0)
 }
 
+function shouldTrimAMPM(start, local, culture) {
+  return /(AM|PM)$/i.test(local.format(start, 'LT', culture));
+}
+
 let dateRangeFormat = ({ start, end }, culture, local)=>
   local.format(start, 'L', culture) + ' — ' + local.format(end, 'L', culture)
 
 let timeRangeFormat = ({ start, end }, culture, local)=>
-  local.format(start, 'h:mma', culture) +
-    ' — ' + local.format(end, inSame12Hr(start, end) ? 'h:mm' : 'h:mma', culture)
+  local.format(start, 'LT', culture) +
+    ' — ' + local.format(end, (inSame12Hr(start, end) && shouldTrimAMPM(start, local, culture)) ? 'h:mm' : 'LT', culture)
 
 let weekRangeFormat = ({ start, end }, culture, local)=>
   local.format(start, 'MMM DD', culture) +


krokofant:
Any input on this? 😄

jquense:
yup sorry, thanks for the PR, busy this weekend. ping me again on Tuesday if I don't get to it :P

krokofant:
@jquense ping

src/localizers/moment.js
```diff
@@ -8,12 +8,16 @@ function inSame12Hr(start, end){
...
+}
+
 let dateRangeFormat = ({ start, end }, culture, local)=>
   local.format(start, 'L', culture) + ' — ' + local.format(end, 'L', culture)
 
 let timeRangeFormat = ({ start, end }, culture, local)=>
-  local.format(start, 'h:mma', culture) +
-    ' — ' + local.format(end, inSame12Hr(start, end) ? 'h:mm' : 'h:mma', culture)
+  local.format(start, 'LT', culture) +
+    ' — ' + local.format(end, (inSame12Hr(start, end) && shouldTrimAMPM(start, local, culture)) ? 'h:mm' : 'LT', culture)
```
jquense: Hi 👋 , 

I'm not sure this is right, ideally we'd want both formats to use something culture agnostic. Right now this might display different things on either side of 12noon, and not just the AM/PM bits. It probably makes more sense to not do any of this sort of logic as the default and just use the normal short time formats?

krokofant: This basically leaves the AM/PM formatting alone. It would be great if just LT was accepted but then you still would have to trim AM/PM from the ranges which are within the same 24 hours.

The previous format would *force* everyone to submit to the AM/PM format which is really terrible localization. I changed very little compared to the original state. The only difference being that we that don't use the AM/PM system won't have it "forced".

> Right now this might display different things on either side of 12noon, and not just the AM/PM bits

Could you elaborate on this? I feel like I tested it thoroughly to do what I describe above. I.E. leaving the previous functionality as it was for users with AM/PM time and using the actual local format for non AM/PM.

jquense: > The previous format would force everyone to submit to the AM/PM format which is really terrible localization. I changed very little compared to the original state. The only difference being that we that don't use the AM/PM system won't have it "forced".

they will have some other things forced tho by the h:mm format, That's all to say that you are correct, this whole choice to try and trim the AM/PM bits is a bad choice for a default. Which is why i'm saying lets go farther than this PR, and remove the trimming logic completely, and just use generic "time" format regardless of where it falls, in this case `"LT"`

jquense:
long time coming but i've just gone and removed the extra logic around AM/PM formats. If folks want that behavior they can customize it :) but not doing that is a better default

Issue #125, opened 2016-06-28
Add ability to remove link in the month day's date number


manutenfruits:
So I'm using this calendar but don't have a `day` view, only `month` and `week`. The link in the date number in the month view (namely [this `<a>` guy](https://github.com/intljusticemission/react-big-calendar/blob/master/src/Month.jsx#L251)) doesn't make sense if there's no day view, so I'd like to add an option to remove it. I'm happy to post a PR, but can't think of a good prop name or type to add.

s-taylor:
I'm having an issue with this at present as well, we have no use for the day view and have restyled the calendar to make the day large and in the centre of the cell. But this then means the onSelectSlot event doesn't fire when you click on the numbers. Ideally I'd like to be able to specify an event for this like `onSelectDay` or something similar, I could then just set it to the same function.

Pandakls:
You can fix it with : 
```javascript
.rbc-date-cell > a{
  pointer-events: none;
}
```

This will disable the onclik on the days!

s-taylor:
Thanks, we were able to work around using z-index to push the a tag behind the cell so the cell was always clicked. I still think it would be great if you could control the event fired by clicking the `<a>` tag specifically though.

PR #127, opened 2016-06-29
Added prop to remove links in date numbers in month view

manutenfruits:
Fixes this: https://github.com/intljusticemission/react-big-calendar/issues/125

Not sure how the best prop/argument schema should be, so I'm happy to take suggestions.

### DIFF
--- a/src/Calendar.jsx
+++ b/src/Calendar.jsx
@@ -152,6 +152,12 @@ let Calendar = React.createClass({
       PropTypes.number,
       PropTypes.shape({ x: PropTypes.number, y: PropTypes.number })
     ]),
+
+    /**
+     * Whether date numbers should have links to the date
+     */
+    dateLink: PropTypes.bool,
+
     /**
      * Allows mouse selection of ranges of dates/times.
      */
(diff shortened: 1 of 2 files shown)


tobiasandersen:
Couldn't this be solved using e.g. `getDrilldownView`?

jquense:
i think the preventing of links could, but not if you don't want the links at all...i'm pretty on the fence about this (hence why it's languished so long). The api isn't great here, but maybe its a necessary use case?

tobiasandersen:
Yeah you're right, I just feel like there's a risk we end up with way too many props for small things like this. I much prefer the render prop approach (or component prop like we already make use of).

jquense:
yeah i agree, if we were to handle this it should be via a component or renderFoo prop

tobiasandersen:
I'll go ahead and close this since it's been a while. Thanks for doing this though @manutenfruits! And if you want to update the PR to use a component or render prop instead of a boolean, just re-open! :)

v-nlondhe:
I am also facing same issue using react big calendar after rendering calendar each day number rendering as anchor tag i need to convert to button tag please let me know is there is props or any custom way to add button on each date on calendar.

PR #132, opened 2016-07-01
Add highlightToday prop to highlight today's column in week view

jgautsch:
Adds a `highlightToday` optional prop (defaults to `false`) that enables highlighting of today's column in the week view.

Here is a screenshot from the storybook:
screen shot 2016-07-01 at 4 08 53 pm


src/TimeColumn.jsx
```diff
@@ -63,9 +65,18 @@ export default class TimeColumn extends Component {
       date = next
     }
 
+    const classes = cn(
+      this.props.className,
+      'rbc-time-column',
+      {
+        'rbc-today': dates.isToday(this.props.max),
```
jquense: I think it might be enough to add the 'rbc-today' class, then folks can style that how they want without needing to add an additional prop. Does that still work for you?

PR #141, opened 2016-07-18
Added TypeScript Definition

piotrwitek:
Hello,
Today my friend asked me to help him create TypeScript Definition for this library:

I have created typings with all the props described in the Docs. I hope it will be useful to devs who are using React together with TypeScript like me, which is highly beneficial for props validation and autocomplete of the entire API with parameter types. Even pure JS devs can benefit from it using IDE that can parse TypeScript declaration (Visual Studio, VS Code, WebStorm).

If you want to see how TypeScript integrate with React please check my Starter-Kit project, it's very clean and simple to follow and grasp (using ES6):
- https://github.com/piotrwitek/react-ts-jspm-starter-kit

In this PR I have also included typings property for package.json, so developers installing this package can have declarations added to their project automatically from node_modules while importing/requiring packages (TypeScript can do it :smile: )

> Typings property explained: https://www.typescriptlang.org/docs/handbook/typings-for-npm-packages.html

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,9 +7,10 @@
   "license": "MIT",
   "main": "lib/index.js",
   "style": "lib/css/react-big-calendar.css",
-  "typings": "lib/react-big-calendar.d.ts",
+  "typings": "typings/react-big-calendar.d.ts",
   "files": [
     "lib/",
+    "typings/",
     "LICENSE",
     "README.md",
     "CHANGELOG.md"
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
   "license": "MIT",
   "main": "lib/index.js",
   "style": "lib/css/react-big-calendar.css",
+  "typings": "lib/react-big-calendar.d.ts",
   "files": [
     "lib/",
     "LICENSE",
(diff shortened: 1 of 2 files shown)


jquense:
hey there! Thanks a lot for the pull request. I think that TS type defs would be a nice help to folks using this component with TS. Here is the concern though, I/we don't use TS, so as soon as the api changes these type defs will be wrong, and no one maintaining RBC will have the motivation or knowledge to update them, or test that they work still. To that end it may be better to host such things on the DT or similar...I know that TS itself is moving away from that, but I'm not sure we can be good stewards of this code once it was merged.

piotrwitek:
I understand your concerns, no worries, I have already contributed to Definitely Typed, and it was merged with master repo  :)
You can already install it using typings npm package
https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/react-big-calendar

Regarding DT from recent TypeScript blog post I don't think they are moving away from DT, as they have introduced a new way of installing typings from DT using npm install command, and all definitions are automatically redistributed from DT to npm registry and also versioned using semver by Microsoft Worker Service as a response to versioning issues with DT.
You can install most definitions by now using command `npm i -D @types/<libraryName>`, Unfortunately react-big-calendar is not published yet to npm and I didn't have time to investigate why, but I'll check that in my free time.

frankwallis:
@piotrwitek I have submitted a PR to definitely typed which updates the exposed module format to be commonjs as that is what is in fact exported by v0.11.1. Hopefully this is ok, let me know if you have any concerns, thanks.

piotrwitek:
@frankwallis review is ready, I requested some clarification, thx

PR #148, opened 2016-07-28
Pass custom dates for a given month

manjunath-nm89:
Hey,

For a given month, I needed the ability to show different dates which are not necessarily in that month. This will be useful to display fiscal calendars like the 4-5-4 etc.

So, I added a way to pass in a function as a prop, where I can pass an array of date objects which I want.

Check it out and let me know if this can be merged and also if any documentation that I need to provide.


jquense:
long time coming for a reply here, sorry.

I think a better way to handle this is to allow setting the actual month range for a given date

Issue #142, opened 2016-07-18
React warnings


frontenddeveloping:
![image](https://cloud.githubusercontent.com/assets/2063415/16928952/10754bca-4d3e-11e6-846a-4076b5f75d56.png)

hannupekka:
In addition to "Unknown props", I am getting these:

> Warning: You are manually calling a React.PropTypes validation function for the `defaultView` prop on `Uncontrolled(Calendar)`. This is deprecated and will not work in the next major version. You may be seeing this warning due to a third-party PropTypes library. See https://fb.me/react-warning-dont-call-proptypes for details.

> Warning: You are manually calling a React.PropTypes validation function for the `defaultDate` prop on `Uncontrolled(Calendar)`. This is deprecated and will not work in the next major version. You may be seeing this warning due to a third-party PropTypes library. See https://fb.me/react-warning-dont-call-proptypes for details.

> Uncaught Invariant Violation: addComponentAsRefTo(...): Only a ReactOwner can have refs. You might be adding a ref to a component that was not created inside a component's `render` method, or you have multiple copies of React loaded (details: https://fb.me/react-refs-must-have-owner).

andreacab:
Any news on that? I am having those errors and warnings too.

jquense:
I'm a bit busy at theoment but PR's are welcome

hannupekka:
I was able to fix `Uncaught Invariant Violation: addComponentAsRefTo(...)` with https://github.com/hannupekka/react-big-calendar/commit/e37010440a0d1fc8b790926970d9078a46774cc8 and `Warning: You are manually calling a React.PropTypes` with https://github.com/hannupekka/react-big-calendar/commit/3dd22db70c27cc5c9629bfab192f6344d12d2b3f.

These fix the week view for me. Unfortunately too busy at the moment to fix monthly view but maybe these will help someone else too.

HriBB:
I create a PR for this. Quick and dirty. Got rid of all the errors in examples ... hopefully :)
Please review the PR and let me know what you think #153 
I agree with @hannupekka that this is not the best way to fix this issue.

HriBB:
@jquense what would be a proper solution in your opinion?

andreacab:
@jquense might also be able to help for the PR if needed after HriBB 's PR review

pihyper:
Facing the same issue. What's the fix?

HriBB:
This should already be fixed in #153

[...]

PR #150, opened 2016-08-03
remove useless props on div which cause react warning

guyanbiao:
the props on div seems not make sense and cause warning mentioned here

https://github.com/intljusticemission/react-big-calendar/issues/142


jquense:
thanks for the PR, however the changes are too drastic, you are removing props that are supposed to be there as well as ones that are not

PR #151, opened 2016-08-03
Allow disabling of event list truncation 

bluespore:


jquense:
thanks for the PR can you please clean up the commits and extra changes.

bluespore:
@jquense Sorry about that - Let me know if it's all good now.

jquense:
Ok I think i'm on board with this escape hatch, however I think there are some issues with it. The truncation of events is based on how many will **fit** in the space, whereas API's that return "and x more" will be based on some number of events. The problem is that without event truncation there is no guarantee that all the events will fit in the space. Your final "And x more" event may be cut off of overflow below the current week. 

Which is all to say i'm concerned that this will actually place the component in a mostly broken state, and that's not great. I'm not actually sure how to handle such a situation

bluespore:
@jquense That's a valid point. Right now, I'd say the prop simply comes with a caveat that once you opt-in you're responsible for sizing adjustments. I understand this might not be desirable for something that best accommodates for many people though.

I'll get back to you within the week when I've got a chance to think about this.

@thienanle As you raised the initial issue, do you have any thoughts on how best to accommodate the change for the majority of users?

thienanle:
@bluespore I think the majority of users wouldn't use this. I would set `truncateEvents` to `true` in `defaultProps` so that out of the box the component isn't in a broken state, but only if the user opts in (at which point they should know what they're doing, since they looked for it in the first place).

jquense:
Going to close. I'm still ok with some solution here, but I don't think this covers it, any disabling of truncation would need to allow the events that are returned to fit without breaking the styling, or have a path for the user to handle that themselves. Otherwise we are gonna get issues asking how to handle it, or that events are overflowing their dates

PR #153, opened 2016-08-04
Fix unknown props warnings

HriBB:
Went through the examples and fixed all unknown props warnings I could find. I am not familiar with the source yet, so there are probably a few more left to omit.

Please let me know if the approach is OK.

### DIFF
--- a/src/Calendar.jsx
+++ b/src/Calendar.jsx
@@ -31,6 +31,8 @@ function isValidView(view, { views: _views }) {
   return names.indexOf(view) !== -1
 }
 
+const omitProps = ['defaultView', 'defaultDate', 'culture', 'rtl']
+
 let now = new Date();
 
 /**
@@ -380,7 +382,7 @@ let Calendar = React.createClass({
     let View = this.getView();
     let names = viewNames(this.props.views)
 
-    let elementProps = omit(this.props, Object.keys(Calendar.propTypes))
+    let elementProps = omit(this.props, Object.keys(Calendar.propTypes).concat(omitProps))
 
     let viewComponents = defaults(
       components[view] || {},
--- a/src/Calendar.jsx
+++ b/src/Calendar.jsx
@@ -17,7 +17,7 @@ import moveDate from './utils/move';
 import VIEWS from './Views';
 import Toolbar from './Toolbar';
 
-import { pickHTMLProps } from 'pick-react-known-prop';
+import pick from 'lodash/object/pick';
 import omit from 'lodash/object/omit';
 import defaults from 'lodash/object/defaults';
 import transform from 'lodash/object/transform';
@@ -381,6 +381,8 @@ let Calendar = React.createClass({
     let View = this.getView();
     let names = viewNames(this.props.views)
 
+    let elementProps = pick(this.props, ['selected', 'step'])
+
     let viewComponents = defaults(
       components[view] || {},
       omit(components, names)
@@ -389,7 +391,7 @@ let Calendar = React.createClass({
     let ToolbarToRender = components.toolbar || Toolbar
 
     return (
-      <div {...pickHTMLProps(this.props)}
+      <div {...elementProps}
         className={cn('rbc-calendar', className, {
           'rbc-rtl': props.rtl
         })}
--- a/src/Calendar.jsx
+++ b/src/Calendar.jsx
@@ -53,6 +53,12 @@ let now = new Date();
 let Calendar = React.createClass({
 
   propTypes: {
+
+    /**
+     * Props passed to main calendar <div>.
+     */
+    elementProps: PropTypes.object,
+
     /**
      * The current date value of the calendar. Determines the visible view range
      *
@@ -324,6 +330,7 @@ let Calendar = React.createClass({
 
   getDefaultProps() {
     return {
+      elementProps: {},
       popup: false,
       toolbar: true,
       view: views.MONTH,
@@ -372,6 +379,7 @@ let Calendar = React.createClass({
       , formats = {}
       , style
       , className
+      , elementProps
       , date: current
       , ...props } = this.props;
 
@@ -389,6 +397,7 @@ let Calendar = React.createClass({
 
     return (
       <div
+        {...elementProps}
         className={cn('rbc-calendar', className, {
           'rbc-rtl': props.rtl
         })}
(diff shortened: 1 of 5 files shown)


HriBB:
While I was at it, I also fixed `<pre>` issue in examples.

src/EventCell.jsx
```diff
@@ -1,8 +1,20 @@
 import React from 'react';
 import cn from 'classnames';
+import omit from 'lodash/object/omit';
 import dates from './utils/dates';
 import { accessor as get } from './utils/accessors';
 
+const omitProps = [
```
jquense: it's probably shorter to `pick` the props we need instead of omitting so many

src/Month.jsx
```diff
@@ -28,6 +28,17 @@ let eventsForWeek = (evts, start, end, props) =>
 
 let isSegmentInSlot = (seg, slot) => seg.left <= slot && seg.right >= slot;
 
+const omitProps = [
```
jquense: ditto here maybe pick instead?

jquense:
thanks for making an attempt here! a few comments

HriBB:
Yeah I did not like those big arrays as well :) Anyway, found this helpful little package called [`pick-react-known-prop`](https://github.com/tyv/pick-react-known-prop),  which does exactly what we need I guess. Let me know if you like it

jquense:
I think i'd rather not take another dependency here, it doesn't really fix the issue, which is that we should be specific about which props are passed down. Most of these components are internal so we know exactly what props need to be passed to the dom nodes and which don't

HriBB:
OK got it! I Will fix It and only props that are needed. So no funny stuff like `onClick` can come through, right?

HriBB:
Hmm ... I'm wondering if other props are even needed on `<div>`.

`Calendar` receives `className` and `style`
`MonthView` receives only `className`
`EventCell` receives `className`, `style` and `onClick`, which executes `onSelect` prop

I went through all props in examples and only found extra `selected` and `step` [`html-attributes`](https://github.com/alexmingoia/html-attributes/blob/master/lib/html-attributes.js) that are passed to either `Calendar`, `MonthView` or `EventCell`.

Do you maybe know of any other props that are or can be passed to internal `<div>`? If not, then I suggest we simply remove all {...this.props} spreads ;)

jquense:
I think we should remove the spreads, just explicitly pass the props that are needed to the DOM nodes instead. there aren't a lot of them

HriBB:
Done! @jquense please review this.

src/Calendar.jsx
```diff
@@ -390,7 +388,7 @@ let Calendar = React.createClass({
     let ToolbarToRender = components.toolbar || Toolbar
 
     return (
-      <div {...elementProps}
```
jquense: This is a bit problematic. The user might be adding DOM events or `aria-`/`data-` attributes. Maybe in this case we add a new prop `elementProps` specifically for passing stuff to this div

HriBB: Something like this?

```
let handlers = [
  'onBlur', 'onChange', 'onClick', 'onContextMenu', 'onCopy',
  'onCut', 'onDoubleClick', 'onDrag', 'onDragEnd', 'onDragEnter',
  'onDragExit', 'onDragLeave', 'onDragOver', 'onDragStart', 'onDrop',
  'onFocus', 'onInput', 'onKeyDown', 'onKeyPress', 'onKeyUp',
  'onMouseDown', 'onMouseEnter', 'onMouseLeave', 'onMouseMove',
  'onMouseOut', 'onMouseOver', 'onMouseUp', 'onPaste',
  'onScroll', 'onSubmit', 'onTouchCancel', 'onTouchEnd', 'onTouchMove',
  'onTouchStart', 'onWheel'
]

let props = {
  'data-id': 1,
  something: 'else',
  invalid: 'prop',
  onClick: function() { console.log('click!'); },
  'data-whatever': 'whateva',
  'aria-invalid': true,
}

function pickValidProps(props) {
  let valid = {}
  for (let p in props) {
    if (handlers.indexOf(p) > -1 || p.startsWith('data-') || p.startsWith('aria-')) {
      valid[p] = props[p]
    }
  }
  return valid
}

console.log('validProps', pickValidProps(props));
```

src/EventCell.jsx
```diff
@@ -24,7 +24,6 @@ let EventCell = React.createClass({
 
     return (
       <div
-        {...props}
```
HriBB: What about this? Are users able to add custom props here?

src/Calendar.jsx
jquense: no something even simpler. the Calendar has a prop called: `elementProps` which is an object

 then you do: `<div {...this.props.elementProps>` on the calendar DOM node.

src/EventCell.jsx
jquense: I don't think so, I think this is good being removed completely here

src/Calendar.jsx
HriBB: But how are you going to pick which props are allowed?

jquense: you don't need too the user is in charge of making sure that they are valid

HriBB: Ohh so you mean something like this?

```
// Calendar.jsx

let Calendar = React.createClass({

  propTypes: {
    elementProps: PropTypes.object,
    ...
  }

  render() {
    let { className, style, elementProps, ...props } = this.props;
    ...
    return (
      <div
        {...elementProps}
        className={cn('rbc-calendar', className, {
          'rbc-rtl': props.rtl
        })}
        style={style}
      >
        ...
      </div>
    );
  }

});
```

And then user passes in `elementProps` in the component

```
// UserCalendar.jsx

import ReactBigCalendar from 'react-big-calendar'

let UserCalendar = React.createClass({

  render() {
    let elementProps = {
      onDrop: function() { ... },
      'data-id': 123
    };
  
    return (
      <ReactBigCalendar elementProps={elementProps} ... />
    );
  }
  
});
```

jquense: yup :thumbsup:

HriBB:
@jquense ok please review and let me know if there's still anything to fix. We could also add some instructions to `README.md`.

jquense:
Looks good! thanks a lot

HriBB:
Great!

PR #155, opened 2016-08-11
fix - wrong end time label in agenda

markostakic:
e.g.
    'start': new Date(2015, 3, 7, 13, 15),
    'end': new Date(2015, 3, 10, 17, 30)
shows 
<< 1:15 pm  // time from start date
instead of 
<< 17:30 pm


jquense:
thanks!

PR #156, opened 2016-08-11
Add an indicator that represents the current time

jonathanweiss:
In the week and day view a horizontal line represents the current time.
Apple Calendar and other apps use this concept to provide the current
time as a subtle hint.

The Y position of line is calculated in the TimeGrid by using a date
diff between this.props.max and this.props.min, respectively the
diff between current date and this.props.min.
To convert the result to a pixel value, the height of the time gutter
is used as a reference.
The position of the time indicator is calulated and updated every
minute using setInterval().

Please let me know if you want this feature to be toggled using a
prop.

screen shot 2016-08-11 at 18 03 50

### DIFF
--- a/src/TimeGrid.jsx
+++ b/src/TimeGrid.jsx
@@ -150,7 +150,7 @@ export default class TimeGrid extends Component {
           this.renderHeader(range, segments, width)
         }
         <div ref='content' className='rbc-time-content'>
-          <div className='rbc-current-time-indicator' ref='timeIndicator'></div>
+          <div ref='timeIndicator' className='rbc-current-time-indicator'></div>
           <TimeColumn
             {...this.props}
             showLabels
@@ -344,8 +344,10 @@ export default class TimeGrid extends Component {
     if (timeGutter) {
       const pixelHeight = timeGutter.offsetHeight;
       const offset = Math.floor(factor * pixelHeight);
+      const timeIndicator = this.refs.timeIndicator;
 
-      this.refs.timeIndicator.style.top = offset + 'px';
+      timeIndicator.style.left = timeGutter.offsetWidth + 'px';
+      timeIndicator.style.top = offset + 'px';
     }
   }
 
--- a/src/TimeGrid.jsx
+++ b/src/TimeGrid.jsx
@@ -74,6 +74,17 @@ export default class TimeGrid extends Component {
       this.measureGutter()
     }
     this.applyScroll();
+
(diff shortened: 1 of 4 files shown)


jonathanweiss:
I improved the indicator by moving it away from the time gutter and adding a little dot to the beginning of it.
screen shot 2016-08-15 at 17 25 17

jquense:
thanks for the PR! Looks good, tho maybe the dot needs a bit more positioning? Can you also add some shots of what it looks like amidst events (with overlay)?

jonathanweiss:
I changed the dot from using an unicode char to a div with border-radius. That should work better without having to worry about font-rendering in different browser. Also the indicator overlaps events now (like in iCal).
screen shot 2016-08-17 at 17 04 02
screen shot 2016-08-17 at 17 04 17

jonathanweiss:
I'm also working on an extension for the indicator that displays the current time, but I need to polish the code a bit  more before I'll create a different PR for that.
screen shot 2016-08-17 at 17 13 01

jquense:
nice! I think tho I like the events overlapping the indicator, seems less than ideal to have it obscure information

src/TimeGrid.jsx
```diff
@@ -74,6 +74,17 @@ export default class TimeGrid extends Component {
       this.measureGutter()
     }
     this.applyScroll();
+
+    this.positionTimeIndicator();
+
+    // Update the position of the time indicator every minute
+    this._timeIndicatorInterval = window.setInterval(() => {
```
jquense: better to use a recursive `setTimeout` here, setInterval can "drift" depending on how clogged the JS event loop is, which can cause it to update rather closely together vs actually every minute. `setTimeout` will ensure that its not called more than once every minute

jonathanweiss: I'm not to concerned about an interval that is triggered every second. Using recursive `setTimeout()` we might exceed the stack trace of the browser after having the calendar open for ~six hours. ;-)
Anyway, I did change to code to work with `setTimeout()`.

I also noticed that I forgot to check for the min time frame (`this.props.min`) and added this check.

The indicator is now rendered **behind** the event.

jonathanweiss:
FYI: I changed the order of the events and time indicator so that the time indicator is rendered behind events and switched to `setTimeout()`.
The current time next to the indicator will be done in another pull request when I have the time to improve my current solution...

jquense:
thanks a ton!

PR #169, opened 2016-09-02
Add "time of day" to the time indicator

jonathanweiss:
Hi jquense,

thanks for accepting my last PR. :) I had some time to polish and improve the implementation to display the "time of day" next to the time indicator.

At the moment the time is aligned right with a bit of spacing to the circle:
screen shot 2016-09-02 at 10 21 37

I could also align it left so that it looks like similar to the time slot label:
screen shot 2016-09-02 at 10 22 25

Here is a screenshot of how it looks together with events:
screen shot 2016-09-02 at 10 14 11

Let me know that you think of this. :)

### DIFF
--- a/src/TimeGrid.jsx
+++ b/src/TimeGrid.jsx
@@ -150,7 +150,7 @@ export default class TimeGrid extends Component {
           this.renderHeader(range, segments, width)
         }
         <div ref='content' className='rbc-time-content'>
-          <div className='rbc-current-time-indicator' ref='timeIndicator'></div>
+          <div ref='timeIndicator' className='rbc-current-time-indicator'></div>
           <TimeColumn
             {...this.props}
             showLabels
@@ -344,8 +344,10 @@ export default class TimeGrid extends Component {
     if (timeGutter) {
       const pixelHeight = timeGutter.offsetHeight;
       const offset = Math.floor(factor * pixelHeight);
+      const timeIndicator = this.refs.timeIndicator;
 
-      this.refs.timeIndicator.style.top = offset + 'px';
+      timeIndicator.style.left = timeGutter.offsetWidth + 'px';
+      timeIndicator.style.top = offset + 'px';
     }
   }
 
--- a/src/TimeGrid.jsx
+++ b/src/TimeGrid.jsx
@@ -74,6 +74,17 @@ export default class TimeGrid extends Component {
       this.measureGutter()
     }
     this.applyScroll();
+
+    this.positionTimeIndicator();
+
+    // Update the position of the time indicator every minute
+    this._timeIndicatorInterval = window.setInterval(() => {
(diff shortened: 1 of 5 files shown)


jonathanweiss:
Hi jquense,

I've noticed that my implementation wasn't very React-like because I used methods to update the TimeIndicator component. I've switched to using props in the most recent commit.

Just let me know what you think about this. :)

jonathanweiss:
Any news on this PR from your side? :)

jquense:
hey there, sorry for the delay. I Do think your approach here is more idiomatic React, however I think that might be a bad thing in this case. Storing the style in state, triggers a re-render of the entire grid vs only updating the one div that changes. Since this actually happens on an interval I think minimizing extra updates due to that interval is a good optimization, even if it does result in less idiomatic code.

In terms of the showing the time, while cool i'm not sure it adds a lot of value. It's likely to obscure the surrounding time labels, and doesn't add a lot more context for the user, who will undoubtedly have a clock in their eye line to get the current time. The line itself is helpful for allowing someone to trace over the grid and still have a sense where "now" is but knowing that now is 7:43 its just as easy to glance at the clock vs back to the right of calendar

jonathanweiss:
> I Do think your approach here is more idiomatic React, however I think that might be a bad thing in this case. Storing the style in state, triggers a re-render of the entire grid vs only updating the one div that changes. 

I agree! That's why I added `needsUpdate()` in order to call `setState()` only if the props have changed. It's a bit like `shouldComponentUpdate()` only for the state... It still feels wrong.
Just out of curiosity: what would be a better solution for this?

> In terms of the showing the time, while cool i'm not sure it adds a lot of value. 

Ok, no worries. :) I needed this feature for a project I was working on and thought it would be a nice addition for other users - that's why I wanted to get it back upstream. Feel free to cancel/close this PR.

jquense:
>  Just out of curiosity: what would be a better solution for this?

I think mutating the style directly is the best actually :P as it avoids even needed updates every min. However if you are doing it in statue I wouldn't worry about having a method on the component, you can compare the current state to the last state to see if it needs an update.

> I needed this feature for a project

I'm curious what the user case was for that if you don't mind talking about it?

jonathanweiss:
> I think mutating the style directly is the best actually :P as it avoids even needed updates every min.

Ok, so maybe a "hybrid" approach would be the best fit: render the initial <TimeIndicator> with props and then update the position by updating the styles every minute.

> I'm curious what the user case was for that if you don't mind talking about it?

Well, the task was to implement a time indicator that mimics the behaviour of Outlook or iCal.

Outlook:
screen shot 2016-09-20 at 08 31 35

iCal:
screen shot 2016-09-20 at 08 32 44

ffdead:
+1

PR #171, opened 2016-09-15
Pass 'selected' prop through to All Day Events

christek91:
The 'selected' prop was not being passed through to the header Event Rows for All Day events in Month/Week views, which meant that any logic which relied on that prop (special styling for selected events, etc.) wouldn't work. Fix is easy: just make sure to pass that prop down.


christek91:
You can check this pretty easily by running the demo and selecting an All Day event. It should now change color when selected, when it didn't before.

jquense:
thanks!

PR #172, opened 2016-09-15
Days selection fix when using RTL languages

DorWeid:
When trying to select a range of days ( when in Months view ) and using a RTL language (such as Hebrew) you'll find it buggy. It'll select the opposite days when in the current row and will do some weird stuff when passing between rows. I could not find any open issues or pull requests regarding that matter, so I decided to do it myself. I also added some documentation to the code to help me better understand it while finding a solution (I can take it down if it's a problem).

TL;DR
- The problem: Wrong selection indexes when using RTL language.
- Ways to reproduce the bug: Use the i18n example. Edit cultures.js by adding the **selectable** prop. Run it and select ar-AE.

p.s
This is my first time creating a pull request.

### DIFF
--- a/src/BackgroundCells.jsx
+++ b/src/BackgroundCells.jsx
@@ -11,7 +11,8 @@ class DisplayCells extends React.Component {
   static propTypes = {
     selectable: React.PropTypes.bool,
     onSelect: React.PropTypes.func,
-    slots: React.PropTypes.number
+    slots: React.PropTypes.number,
+    rtl: React.PropTypes.bool
   }
 
   state = { selecting: false }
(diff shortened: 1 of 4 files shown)


jquense:
thanks for the contribution! I'll take a look at the code, its a bit tough for me to see what's new with the comments but that's ok, just take me a bit longer to read through

jquense:
hey there :wave: sorry for the delay, can you rebase this to fix the conflict?

DorWeid:
Hey, I actually found a better solution than the current one, as it applies only to drag selection event, and not single cell selection. I will update soon.

DorWeid:
Hey @jquense 
Project should be OK now and is up to date with master branch.
Also removed documentation for your convenience :p

jquense:
thanks a bunch for this! sorry for the delay

JonathanGilyadov:
Hey,

Still having this issue now, any idea why?

PR #184, opened 2016-10-06
Bugfix: Multiple showmore popups are being rendered

palashkaria:
Issue: One popup per row [per week] was being rendered

Solution: Moved _renderOverlay from renderWeek() to render(), also added check for this.state.overlay

This resolves the issue of popup closing on clicking inside the popup [or selecting any event]

### DIFF
--- a/src/Month.jsx
+++ b/src/Month.jsx
@@ -121,7 +121,9 @@ let MonthView = React.createClass({
         { weeks.map((week, idx) =>
             this.renderWeek(week, idx, measure && this._renderMeasureRows))
         }
-        { this.props.popup && this.state.overlay?this._renderOverlay():null}
+        { this.props.popup &&
+            this._renderOverlay()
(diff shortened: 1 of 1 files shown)


src/Month.jsx
```diff
@@ -121,6 +121,7 @@ let MonthView = React.createClass({
         { weeks.map((week, idx) =>
             this.renderWeek(week, idx, measure && this._renderMeasureRows))
         }
+        { this.props.popup && this.state.overlay?this._renderOverlay():null}
```
jquense: hey there. This is fine generally, however the added check for `state.overlay` isn't needed overlay visibility is handled by the `show` prop, its not a big deal now, but if there was a close animation it would be cut off.

also can you remove the `? : null` bit its fine to keep it how it was.

jquense:
thanks!

Issue #194, opened 2016-10-17
Allow `formats` to return a renderable instead of a string


ffdead:
I need to inject some extra markup into the day header but it seems like the format can only handle a string.

I would need to return a couple of nested react elements in order to support the following design:

![screen shot 2016-10-17 at 15 09 21](https://cloud.githubusercontent.com/assets/420472/19438761/bb09ce74-947b-11e6-8516-4cc9bcc308b0.png)

Something like this would have been good:

```
const formats = {
    dayFormat: (date, culture, localizer) => {
      return (
        <div>
          <span>{localizer.format(date, 'dddd', culture)}</span>
          <span>{localizer.format(date, 'DD', culture)}</span>
          <TaskCountLabel date={date} />
        </div>
      )
    }
  }
```

Any ideas on how to solve this? Big thanks!

jquense:
the `format` is the wrong place for that, its not a really a rendering API but a hook into the serializing and deserializing logic of i18n libraries. Generally we want to solve your case with specific `component` props, that are dedicated (and distinct from) localization.

we'd happily take a PR for that :)

PR #196, opened 2016-10-18
Adding optional custom Header component for month, week, day views

ffdead:
First stab at solving #194

### DIFF
--- a/src/Month.jsx
+++ b/src/Month.jsx
@@ -263,6 +263,7 @@ let MonthView = React.createClass({
       >
         <HeaderComponent
           date={day}
+          label={localizer.format(day, format, culture)}
           localizer={localizer}
           dayFormat={format}
           culture={culture} />
--- a/src/Month.jsx
+++ b/src/Month.jsx
@@ -265,7 +265,7 @@ let MonthView = React.createClass({
           date={day}
           label={localizer.format(day, format, culture)}
           localizer={localizer}
-          dayFormat={format}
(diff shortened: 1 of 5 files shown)


jquense:
Looks good, thanks! One nit, is that instead of passing in the just the localizer/culture/format, lets pass resulting `label` from the localizer, so folks don't have to understand the localizer API, which isn't really exposed elsewhere. I think its fine to continue to pass the localizer and it's args in case there is really a situation where an escape hatch is needed

ffdead:
@jquense like that? Yea I actually need to use the localizer and date in my custom header so it's great if we can keep passing them.

src/Month.jsx
```diff
@@ -252,14 +253,20 @@ let MonthView = React.createClass({
...
         key={'header_' + idx}
         className='rbc-header'
         style={segStyle(1, 7)}
       >
-        { localizer.format(day, format, culture) }
+        <HeaderComponent
+          date={day}
+          label={localizer.format(day, format, culture)}
+          localizer={localizer}
+          dayFormat={format}
```
jquense: maybe just `format` for the prop name it may be confusing which format it is across different usages, `weekdayFormat`, `dayFormat`, etc

ffdead: good point

ffdead:
Sure, it looks like the ellipsis is inherited from the parent:

screen shot 2016-10-18 at 18 00 59

PR #207, opened 2016-11-02
doc update

AubreyHewes:
Simple doc update.. at the least listing the view types for dev usability..


AubreyHewes:
Hm.. ok :sheep:

Issue #115, opened 2016-06-09
Support background events


dgouldin:
Full Calendar has support for forcing events to render in the background:

http://fullcalendar.io/docs/event_rendering/Background_Events/

It would be useful to support this option, either explicitly or via a css class applied through `eventPropGetter`.

Something to note: Full Calendar doesn't gracefully handle the scenario where 2 background events overlap.

jquense:
hmm interesting, some thing like background sections would work fine? I'd be open to a PR for something like this. I do think that making them proper events seems a bit odd though, instead it seems cleaner to have a seperate prop for "segments" or something to define ranges of styled area...iI'm not sure if overlapping ones make sense. what is the use case for that and how would you handle it?

dgouldin:
Background sections would probably work. Today, we use recurring events to contextualize other events which occur within the same range of time. Some example are explicitly noting availability or indicating travel location for a period of time. Because the contextualizations can overlap, we do need a way to visualize multiple "sections". Overlapping events already have an affordance for this with their indentation level. I'd assume whatever background rendering solution we end up with would use a similar indentation strategy for overlapping background sections and would also affect the default indentation level of events which overlap sections.

jquense:
I think in the case of the TimeGrid views (week, day) handling the overlapping bits would be fairly straight forward, and so to would layout, , you'd just need to add another "layer" so to speak. It may be a bit harder to handle the Month view where stuff isn't really absolutely positioned...then again BackgroundCells may be able to handle it with not much tweaking..

dgouldin:
What if we solved it using the idea of a "layer" number on an event which is akin to z-index: lower numbers render below higher numbers? The collision detection to determine indentation level would need to be refactored, but I think everything else would "just work". Default behavior in the absence of a layer value is a design decision. Default to layer 1? Default to always on bottom?

jquense:
I was thinking about that, ultimately though I don't think things like "Busy" or "Work Hours" are really events, and I don't think it makes sense to mix those sorts of things together. I also i'd think that the styling concerns for those ranges would be fairly different from the "normal" events yeah? for instance perhaps handling overlap for background sections is better down via color and opacity vs indentation.

dgouldin:
Yeah that sounds right, or visually splitting the background blocks horizontally to make room for overlaps. But you're right: rendering them as events conflates their concerns.

lorinkoz:
[...]

PR #210, opened 2016-11-05
add background events

bgates:
Hi, I was interested in the suggestion by @dmoskovsov in #103 so I tried to resolve the concerns you had about the API there.  An event object has a Boolean flag to mark it as `background`; the default behavior is to place the title of a background event in the same row as the event date on the month calendar, but it's also possible to set up a separate component and prop getter to structure and style the background event.

### DIFF
--- a/src/BackgroundCells.jsx
+++ b/src/BackgroundCells.jsx
@@ -1,17 +1,23 @@
-import React from 'react';
+import React, { PropTypes } from 'react';
 import { findDOMNode } from 'react-dom';
 import cn from 'classnames';
+import { elementType } from './utils/propTypes';
 import { segStyle } from './utils/eventLevels';
 import { notify } from './utils/helpers';
+import dates from './utils/dates';
 import { dateCellSelection, slotWidth, getCellAtX, pointInBox } from './utils/selection';
 import Selection, { getBoundsForNode } from './Selection';
 
 class DisplayCells extends React.Component {
 
   static propTypes = {
-    selectable: React.PropTypes.bool,
-    onSelect: React.PropTypes.func,
-    slots: React.PropTypes.number
+    selectable: PropTypes.bool,
+    onSelect: PropTypes.func,
+    slots: PropTypes.number,
+    week: PropTypes.array,
+    backgroundEvents: PropTypes.array,
+    backgroundEventComponent: elementType,
+    backgroundEventPropGetter: PropTypes.func
   }
 
   state = { selecting: false }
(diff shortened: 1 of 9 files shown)


jquense:
hey there, thanks a lot for the PR! I'm excited to take a look at this but on vacation this week so there may be a delay.  thanks!

src/BackgroundCells.jsx
```diff
@@ -33,20 +39,43 @@ class DisplayCells extends React.Component {
...
+      let backgroundEvent = backgroundEvents.find(event => {
+        return dates.sameDay(event.start, week[i])
+      })
+
+      let Component = backgroundEventComponent;
+      let title = (backgroundEvent || {}).title;
+
+      var style, xClassName;
+      if (backgroundEvent && backgroundEventPropGetter)
+        ({ style, className: xClassName } = backgroundEventPropGetter(backgroundEvent));
```
jquense: `let { style, className: xClassName } = backgroundEventPropGetter(backgroundEvent) || {};`

src/BackgroundCells.jsx
```diff
@@ -33,20 +39,43 @@ class DisplayCells extends React.Component {
...
+          style={{...segStyle(1, slots), ...style, ...props.style}}
+          className={cn('rbc-day-bg', xClassName, {
             'rbc-selected-cell': selecting && i >= startIdx && i <= endIdx
           })}
-        />
+        >
+        { Component && backgroundEvent
+          ? <Component
+              event={backgroundEvent}
+              eventPropGetter={backgroundEventPropGetter}
```
jquense: this has already been used above, I wouldn't pass it down further if the prop has been consumed

jquense:
hi there 👋  thanks again for the PR. I think there are probably two different features here, which are each a bit confused. 

1. styling background cells 
2. adding "background events" or labeled regions.

I think this PR would work great for 1. (the original issue) but has some issues with being a solution for 2. The big one is that there is no support for overlapping background events, or handling of the styling for the time views (day, week, etc).

I think it probably makes sense to roll back the back event stuff for now and just go with the backgroundPropGetter. Implementing 2. should probably involve some of the talk in https://github.com/intljusticemission/react-big-calendar/issues/115

does that make sense?

src/BackgroundCells.jsx
bgates: That assumes there's a `backgroundEventPropGetter`, doesn't it?

bgates:
I guess I'd call them 'interrelated' features rather than 'confused' 😄 . The problem I see with separating them is that I don't know why the background would be styled differently in the absence of a background event.

PR #214, opened 2016-11-10
Pass in view component instead of view string to allow the navigate m…

nsearle:
…ethod on a custom defined view component to be used


jquense:
thanks!

Issue #215, opened 2016-11-10
Bubble SyntheticEvent up the chain


aray12:
So I have a use case where I want access to the cursor location during `onSelectEvent` to create a context menu filled with possible actions to take on an event. I was wondering if you'd be open to a PR that simply bubbled the SyntheticEvent up or if this is something that had been solved by other mean before?

PR #216, opened 2016-11-10
Included SyntheticEvent object in callback for onSelect in EventCell

aray12:
I created #215 and figured I'd submit a preliminary PR to see what you guys thought. It doesn't seem to be a huge change and would provide some interesting capabilities to users if they wanted. Currently on some of the views the arguments I am getting are 

`[event, SyntheticEvent, null, SyntheticEvent]` instead of `[event, SyntheticEvent]`

If this is something you guys are interested in, I can dig a bit more and find the source of that. I also think it might be interesting to provide this same capability to `onSelectSlot`, but I think that would be a bit more difficult given the dragging and my current use case doesn't require it.

Thoughts?

### DIFF
--- a/src/DayColumn.jsx
+++ b/src/DayColumn.jsx
@@ -294,9 +294,9 @@ let DaySlot = React.createClass({
     })
   },
 
-  _select(event){
+  _select(...args){
     clearTimeout(this._clickTimer);
-    notify(this.props.onSelectEvent, event)
+    notify(this.props.onSelectEvent, args)
   }
 });
 
--- a/src/DayColumn.jsx
+++ b/src/DayColumn.jsx
@@ -159,7 +159,7 @@ let DaySlot = React.createClass({
           key={'evt_' + idx}
           style={{...xStyle, ...style}}
           title={label + ': ' + title }
-          onClick={this._select.bind(null, event)}
+          onClick={(e) => this._select(event, e)}
           className={cn('rbc-event', className, {
             'rbc-selected': _isSelected,
             'rbc-event-overlaps': lastLeftOffset !== 0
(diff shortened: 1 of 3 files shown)


jquense:
looks good thanks! I would appreciate if you could see why the arguments are weird like that if possible though, thanks!

aray12:
Ya if you will accept the changes I will take some time out later today to figure out why those are acting weird. It appears to only being happening when you click on an event in month (not an all day event though) and single day view. If that rings a bell let me know. Otherwise I'll figure it out later.

src/DayColumn.jsx
```diff
@@ -159,7 +159,7 @@ let DaySlot = React.createClass({
           key={'evt_' + idx}
           style={{...xStyle, ...style}}
           title={label + ': ' + title }
-          onClick={this._select.bind(null, event)}
```
aray12: So the extra parameters being passed were due to using `.bind` instead of using an arrow function to explicitly pass the `SyntheticEvent` up the chain. Apparently, React's event handler signature is actually `function (SyntheticEvent, null, SyntheticEvent) {}`. By using `.bind` it was passing the additional parameters as well. 

This looks like a small bug in React that will be fixed in a future release (and really has no impact here other "code-smell"). You can follow more on the topic [here](https://github.com/facebook/react/issues/7902).

Anyway, it is fixed now. I'd appreciate if you could confirm that the existing binding with no context was not implemented for a specific reason. @jquense you wrote both lines of code within 2 days of each other so was there a change of heart on binding vs arrow functions? Or just random coding decision? Or another reason?

aray12:
@jquense I've provided an update with some notes on why we were getting the weird arguments. Let me know if you have any feedback. Any updates on release would be appreciated. 

Thanks so much for the great work on this awesome project. Really improved our platforms calendar functionality!

jquense:
thanks! much appreciated 👍

PR #236, opened 2016-12-20
Fix errors for Meteor

qge:
When being used for Meteor projects,
1. defining propTypes inside class extensions causes errors. Thus moving propTypes and defaultPropTypes outside of class def.
2. Move state variable creation into constructors.
3. Upgrade uncontrollable to 4.0.3, which fixes compatibility error with React >= 15: https://github.com/jquense/uncontrollable/commit/47b1645142cb83c29032cbc4c30dd2eb56c5d292

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,7 +107,7 @@
     "lodash": "^3.10.1",
     "react-overlays": "^0.6.0",
     "react-prop-types": "^0.2.1",
-    "uncontrollable": "^3.1.0",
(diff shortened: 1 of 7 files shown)


jquense:
this is out of date now, but beside that I'm not sure why you'd need this. Specifically moving hte propTypes. meteor along with anyone else should be consuming the compiled code not the `src` directly

PR #241, opened 2016-12-25
Make button toolbar more responsive

brianrc:
The button toolbar wasn't scaling down nicely to smaller devices. This allows the buttons the overflow and keeps the calendar month/label centered.

Also had to modify the demo styles so calendar didn't display beyond its container.

### DIFF
--- a/examples/styles.less
+++ b/examples/styles.less
@@ -38,21 +38,11 @@ h4 {
 
 .examples {
   position: relative;
-  height: 100vh;
-  display: flex;
-  flex-direction: column;
 }
 
 .example {
   font-size: 14px;
   flex: 1;
-
-  &,
-  & > * {
-    display: flex;
-    flex-direction: column;
-    min-height: 0;
-  }
 
   .demo,
   .rbc-calendar {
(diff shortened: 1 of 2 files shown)


jquense:
:wave: sorry for the delay. thanks for the PR :) I'm not sure the buttons should wrap to the next line if the don't fit. Maybe they should jsut shrink instead? Also I don't think you need hte demo style fixes anymore, updates ahve addressed that.

brianrc:
Thanks. Hmm. I'm not sure how to shrink them and I couldn't think of another way to make it responsive.

This was my compromise. Think it's better than simply breaking the site on smaller devices.

jquense:
yeah I mean it's a bit tough to handle from a library perspective since folks are gonna customize and change the text and labels around its hard to make assumptions about how to handle stuff on smaller screens in a way that works for everyone  :/

brianrc:
True, people are going to customize and change. I still think we can come up with something that reasonably works on smaller devices even if it's not perfect.

TechSaq:
> yeah I mean it's a bit tough to handle from a library perspective since folks are gonna customize and change the text and labels around its hard to make assumptions about how to handle stuff on smaller screens in a way that works for everyone :/

If not too complex, it will be better to have a Dropdown for Views part of the calendar to select Months, Weeks, Day.... for smaller devices and it will work for everyone.

PR #246, opened 2016-12-29
BackgroundCells PropTypes - allow `selectable` to take 'ignoreEvents'

### LINKED ISSUE #244: BackgroundCells.jsx `selectable` PropTypes should include 'ignoreEvents' string
it seems the `selectable` prop of BackgroundCells ought to be `selectable: React.PropTypes.oneOf([true, false, 'ignoreEvents'])`, just like DayColumn. 

BackgroundCells uses this logic on line 102: `if (this.props.selectable !== 'ignoreEvents') return`

Right now, it's `selectable: React.PropTypes.bool` only.

IanLondon:
Fixes #244

### DIFF
--- a/src/BackgroundCells.jsx
+++ b/src/BackgroundCells.jsx
@@ -13,7 +13,7 @@ class BackgroundCells extends React.Component {
 
   static propTypes = {
     cellWrapperComponent: elementType,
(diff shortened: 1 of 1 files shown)


jquense:
thanks!

PR #248, opened 2016-12-30
Remove duplicate Calendar proptype

russkel:


russkel:
Solved in latest cleanup.

Issue #257, opened 2017-01-05
Expose constants for navigation


tobiasandersen:
Hi! Really enjoying this library so far, great work!

I'm currently using the `views` constants to change view:
`onViewChange(views.DAY)`

It'd be nice to do the same for navigation: 
`onNavigate(navigate.NEXT)`

But they aren't exposed, so I'm currently using strings in various places. This is obviously not the end of the world, but still, is there a reason you haven't exposed the [navigate constants](https://github.com/intljusticemission/react-big-calendar/blob/master/src/utils/constants.js#L2-L7)? Or would you be okay with a PR for this?

jquense:
i'm happy to take a PR for that thanks

PR #258, opened 2017-01-05
Expose navigate constants

tobiasandersen:
This PR addresses #257. Just let me know if there's anything else needed here!

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -6,7 +6,7 @@ import momentLocalizer from './localizers/moment';
 import globalizeLocalizer from './localizers/globalize';
 import viewLabel from './utils/viewLabel';
 import move from './utils/move';
-import { views } from './utils/constants';
+import { views, navigate } from './utils/constants';
 
 Object.assign(Calendar, {
   setLocalizer,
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -6,14 +6,15 @@ import momentLocalizer from './localizers/moment';
...
-import { views } from './utils/constants';
+import { views, navigate } from './utils/constants';
 
 Object.assign(Calendar, {
   setLocalizer,
   globalizeLocalizer,
   momentLocalizer,
   label: viewLabel,
   views,
+  navigate,
```
jquense: small nit, I'd capitalize this to `Navigate` to be a bit clearer it's a namespace...i should have done that with `views` as well

tobiasandersen: I agree. Do you prefer having one capitalized and the other not then? I.e. do you want me to only change it to `Navigate`? Could also add a duplicate for `views` called `Views`, and deprecate the lowercase in a later version...

jquense: >  Could also add a duplicate for views called Views, and deprecate the lowercase in a later version...

perfect

jquense:
thanks!

tobiasandersen:
Thanks for the quick response!

PR #262, opened 2017-01-11
Update EventRowMixin.js

### LINKED ISSUE #261: Version 0.12.0 does not support custom component for events in month-view anymore?
Hi there,

I was using calendar version 0.11.1 until today which worked fine. After updating to 0.12.0 (because of #201) the calendar seems not support a custom component for components.event anymore.

This only happens in the month-view (week and day seem to work just fine)

Basically i'm doing this:

```javascript
import Entry from '../Entry';
[..snip..]
<BigCalendar
    [..snip..]
    views={['month', 'week', 'day']}
    components={{
        event: Entry,
    }}
/>
```

What still works is passing a custom component for the toolbar.

HerrZatacke:
fixes #261 fixed property name for custom component

### DIFF
--- a/src/EventRowMixin.js
+++ b/src/EventRowMixin.js
@@ -54,7 +54,7 @@ export default {
         allDayAccessor={allDayAccessor}
         slotStart={start}
         slotEnd={end}
-        component={eventComponent}
+        eventComponent={eventComponent}
       />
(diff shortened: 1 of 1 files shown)


jquense:
thanks!

Issue #202, opened 2016-10-27
Clicking "show more" link without popup doesn't always open day view


ffdead:
In Month view: We had some issues with the "+more" popup and decided to disable it and instead have the calendar open the day view which seems to be the default behaviour. This is a good behaviour.

However, when clicking the "+ more" link in days belonging to the previous or next month, the link switches month to that month instead of opening the day view. 

In the example below, clicking any day in the current month (1-31) will correctly open that day in Day view. Clicking previous month (grey 26-30) switches the month view to previous month instead of opening day view. Clicking next month (grey 1-6) switches the month view to next month instead of opening day view.

<img width="974" alt="screen shot 2016-10-27 at 09 58 18" src="https://cloud.githubusercontent.com/assets/420472/19759264/71dd18d8-9c2c-11e6-9b8b-5086d5de82f4.png">

This is very confusing for the user, and I would consider this a bug?

tobiasandersen:
I can (mostly) confirm this. Clicking a date in the current month will navigate to the day view of that day. However, clicking a date in the previous or next month doesn't do anything in our case. I agree with @ffdead that it should behave the same for all dates. 

Is there a reason for the current behavior? I'd be happy to create a PR for it otherwise.

jquense:
I'd take a PR to fix this

tobiasandersen:
Cool, I'll probably have time to fix it later this week.

jquense:
thanks :)

tobiasandersen:
The problem is [this check](https://github.com/intljusticemission/react-big-calendar/blob/master/src/Calendar.js#L527):

`dates.eq(date, nextDate, view, localizer.startOfWeek(culture))`

It will evaluate to false if `date` and `nextDate` are in different months (and therefore won't navigate to the day view `nextDate`). In this particular case I can't see why this check is needed, but I guess it's there for a reason? 

@jquense How do you suggest I go about this? If the check is needed for something else, maybe we can add some parameter to override/enforce the check.

jquense:
ah. I'm not sure wht the check is for, I think it was leftover from older code perhaps. I'd just try and remove it and see if anything bad happens with regard to navigation.

tobiasandersen:
I did, see https://github.com/intljusticemission/react-big-calendar/pull/265, and I haven't found anything weird so far (just clicked through our app and the examples).

tobiasandersen:
@jquense I think this can be closed now?

PR #265, opened 2017-01-12
Always navigate to day view when action is navigate.DATE

tobiasandersen:
Fix for #202

### DIFF
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -518,15 +518,7 @@ let Calendar = React.createClass({
     onNavigate(date, view)
 
     if (action === navigate.DATE)
-      this._viewNavigate(date)
-  },
-
-  _viewNavigate(nextDate) {
-    let { view, date, culture } = this.props;
-
-    if (dates.eq(date, nextDate, view, localizer.startOfWeek(culture))) {
       this.handleViewChange(views.DAY)
-    }
   },
 
   handleViewChange(view){


jquense:
thanks!

MarshallOfSound:
I came here to try to remove this behavior.  Now when you are on month view and you send a custom navigate request to a specific date we always end up on the day view (which is not expected).  Would you accept a PR @jquense to add a flag so that this day view change can be explicitly prevented?

jquense:
@MarshallOfSound how are you doing: " send a custom navigate request". Passing a new date to the Calendar won't hit this code.

src/Calendar.js
```diff
@@ -518,15 +518,7 @@ let Calendar = React.createClass({
     onNavigate(date, view)
 
     if (action === navigate.DATE)
-      this._viewNavigate(date)
-  },
-
-  _viewNavigate(nextDate) {
-    let { view, date, culture } = this.props;
-
-    if (dates.eq(date, nextDate, view, localizer.startOfWeek(culture))) {
```
jquense: @tobiasandersen I remember why this check was here. If you try to navigate to an arbitrary date from the month view you can't now without it also switching to the day view

tobiasandersen: Oh, I see. What do you want to do with it?

jquense: i'm working on something. the problem is conflating these two things in the first place, but i need to add some more props to fix it for good :P I'll have something up in a bit

Issue #263, opened 2017-01-11
Improve day view layout


tobiasandersen:
The day view isn't very good at showing many events at the same time:

![screen shot 2017-01-11 at 13 48 34](https://cloud.githubusercontent.com/assets/3599069/21849240/065a0fde-d805-11e6-940a-15cb18cccb6d.png)
![screen shot 2017-01-11 at 13 48 54](https://cloud.githubusercontent.com/assets/3599069/21849244/08e37a4c-d805-11e6-9773-d07cc4dc1f15.png)

I much prefer the way e.g. Apple handles it in their Calendar app in OS X:

![screen shot 2017-01-11 at 13 49 57](https://cloud.githubusercontent.com/assets/3599069/21849255/13d86124-d805-11e6-9aae-e7418126ede0.png)

Would you be interested in a PR for something similar to the image above?

jquense:
The calendar events shouldn't overflow past their column, thats definitely a bug. In terms how to display them, I prefer the approach it takes now (more like google calendar) vs the the mac calendar app (which also is showing less events?). I'd be open seeing the PR if it does make situations better _without making the other common situtations any worse_, which is always the concern with layout algorithms, especially in css.

tobiasandersen:
Oh I see. That bug threw me off a bit, I didn't even realise it was a Google-like approach. I don't think Apple shows less events (I just failed at creating the same number of events in my example). 

I just took a look at the differences between the two (Apple and Google):

Google:

![screen shot 2017-01-11 at 14 24 04](https://cloud.githubusercontent.com/assets/3599069/21850289/2e93fe92-d80a-11e6-87b9-16c286ac1c5e.png)

Apple:

![screen shot 2017-01-11 at 14 24 11](https://cloud.githubusercontent.com/assets/3599069/21850293/32b05b9c-d80a-11e6-942e-288e1bc416c7.png)

Here I personally prefer Apple's approach. I'm not a fan of having different widths for events on the same row. However, when there are events that "contain" other events, I prefer the way Google handles it:

Google:

![screen shot 2017-01-11 at 14 34 57](https://cloud.githubusercontent.com/assets/3599069/21850515/4132071e-d80b-11e6-9d46-1b1dc979a344.png)

Apple:

![screen shot 2017-01-11 at 14 35 04](https://cloud.githubusercontent.com/assets/3599069/21850520/4667d038-d80b-11e6-8d17-1bd6cdb58da1.png)

[...]

PR #269, opened 2017-01-16
Improve day view layout

tobiasandersen:
This is a PR for #263. A few things to note:

I added some more example events while testing (they will always show up on today's date). Just let me know if you want them removed. 

Layout-wise I consider an event a sibling, rather than a child, if the difference between them is [less than 1 hour](https://github.com/intljusticemission/react-big-calendar/compare/master...nxtedition:improve-day?expand=1#diff-477af4b576754ce4c4204de679991fcdR230). This is by no means a perfect number, I just found it to work well for the examples I tried with.


src/DayColumn.js
```diff
@@ -219,23 +217,155 @@ let DaySlot = React.createClass({
...
+    let isChild = (parentIdx, childIdx) => {
+      if (isSibling(parentIdx, childIdx)) return false
+
+      let parentEnd = getSlot(events[parentIdx], endAccessor)
+      let childStart = getSlot(events[childIdx], startAccessor)
+
+      return parentEnd > childStart
+    }
+
+    let getSiblings = (idx) => {
```
jquense: can we move functions to the top of the file, that don't need a closure?

src/DayColumn.js
```diff
@@ -219,23 +217,155 @@ let DaySlot = React.createClass({
     })
   },
 
-  _slotStyle(startSlot, endSlot, leftOffset){
+  _getStyledEvents() {
+    let { min, startAccessor, endAccessor, step } = this.props
```
jquense: maybe pull out this method into a new util/ file. It's quite large

src/DayColumn.js
```diff
@@ -219,23 +217,155 @@ let DaySlot = React.createClass({
...
+    // Each iteration goes over all related/overlapping events
+    while (idx < events.length) {
+      let siblings = getSiblings(idx)
+      let { childGroups, nbrOfChildColumns } = getChildGroups(
+        idx, idx + siblings.length + 1
+      )
+      let nbrOfColumns = Math.max(nbrOfChildColumns, siblings.length) + 1
+
+      // Set styles to top level events
+      Array.of(idx).concat(siblings).forEach((eventIdx, siblingIdx) => {
```
jquense: `[idx, ...siblings]`. `Array.of` requires a polyfill in older browsers

src/DayColumn.js
```diff
@@ -219,23 +217,155 @@ let DaySlot = React.createClass({
...
+        top,
+        height: bottom - top
+      }
+    }
+
+    let OVERLAP_MULTIPLIER = 0.3
+    let idx = 0
+
+    // Each iteration goes over all related/overlapping events
+    while (idx < events.length) {
```
jquense: I'm a bit concerned about the complexity of these loops....I suppose it's not likely to deal with any large arrays tho.

src/DayColumn.js
jquense: It'd also be great if you could give a high level description (in a block comment in the file) of the algorithm used as a way to help document and maintain the code.

tobiasandersen: Sure!

src/DayColumn.js
tobiasandersen: I suppose we could, but then we also have to pass startAccessor, endAccessor and events each time. Do you prefer that?

jquense: I think so, we do that already for other utils (see eventLevels.js)

src/DayColumn.js
tobiasandersen: Yeah I kind of agree, but I don't think I can simply much more (using this approach) without making it even harder to follow. Not sure it's worth being clever to avoid a few loops in this case?

src/DayColumn.js
tobiasandersen: Oh, didn't think of that. I avoided your suggestions because of the need for semicolons. Guess I'll just add semicolons everywhere?

src/DayColumn.js
tobiasandersen: Ok cool.

tobiasandersen:
Thanks for your comments! I'll fix this tomorrow.

src/DayColumn.js
jquense: feel free to add them where needed. I usually do `; [idx, ...siblings]` but you can just add it to the end of the line above as well. I'm not particularly dogmatic about it :P

tobiasandersen:
I've tried to address your comments now, let me know if there's anything else needed.

examples/events.js
```diff
@@ -66,5 +66,150 @@ export default [
     'title': 'Birthday Party',
     'start':new Date(2015, 3, 13, 7, 0, 0),
     'end': new Date(2015, 3, 13, 10, 30, 0)
-  }
+  },
+  ...today(1)
 ]
+
+function today(idx = 0) {
```
jquense: sorry can we move this to the `stories` events. It doesn't make as much sense in the doc demos but the storybook examples used for dev would definitely benefit from having this

jquense: by that I mean somewhere in: https://github.com/intljusticemission/react-big-calendar/tree/master/stories

src/utils/dayViewLayout.js
```diff
@@ -0,0 +1,237 @@
...
+      // Set styles to child events.
+      group.forEach((eventIdx, i) => {
+        let parentStyles = events[parentIdx].styles
+        let spaceOccupiedByParent = parentStyles.width + parentStyles.xOffset
+        let columns = Math.min(group.length, nbrOfColumns)
+        let width = (100 - spaceOccupiedByParent) / columns
+        let xAdjustment = spaceOccupiedByParent * OVERLAP_MULTIPLIER
+        let { top, height } = getYStyles(eventIdx, helperArgs)
+
+        events[eventIdx].styles = {
```
jquense: is this the user provided event object? I'd prefer not to mutate data if possible, maybe we can do `{ event, style }`?

tobiasandersen: Good point, I'll change that.

examples/events.js
tobiasandersen: Sure!

tobiasandersen:
Wasn't really sure what to do with the stories...feel free to remove it (or tell me to) if you don't like it.

jquense:
thanks a ton for this.

tobiasandersen:
Thanks!

PR #287, opened 2017-01-31
Split non-all-day events to be shown on multiple days in week view

SimoneCorsini:
If an event runs between different days, it is automatically shown as an all day event in the week view.
This solves the problem by creating multiple events from start to end for each day.
I would also suggest a feature flag for this.


jquense:
thanks for the PR. please see my thoughts on this approach in https://github.com/intljusticemission/react-big-calendar/pull/42 

ultimately I think this approach doesn't really suit long events which will overwhelm the week/day views

PR #308, opened 2017-02-22
Month view day props getter & .rbc-day-bg.rbc-off-range

acepoblete:
Added a function / hook following the same idea of the eventPropGetter to allow users to customize the style and class name of the day background cells.

Also added in a css class called rbc-off-range to allow users to style off range day cells via css.

### DIFF
--- a/examples/demos/basic.js
+++ b/examples/demos/basic.js
@@ -3,12 +3,17 @@ import BigCalendar from 'react-big-calendar';
 import events from '../events';
 
 let Basic = React.createClass({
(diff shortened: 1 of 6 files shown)


jquense:
Hi there, thanks for the PR. The prop setting can already be handling by providing a `cellWrapperComponent` that let's you customize the component that get's rendered so I'd use that. I'm ok adding an `rbc-off-range` class if you want to seperate that out, but please don't add any new styling to it, lets leave that up to users.

PR #309, opened 2017-02-23
Business hours - Disable certain time slots on the calendar.

DonovanCharpin:


jquense:
thanks for taking the time to put this together. Contributions are always appreciated. I'm sure this is the right approach to this however. I don't think it's necessarily helpful to have a separate concept of "businessHours". Generally we want to implement less specific API's that allow folks to extend the calendar to meet their needs, rather than try and cover all bases. 

The Day/week views already have `min` and `max` which are functionally equivalent to `businessHours.start`/`end` for limiting views. If you want to show the full day but only limit selection, that can already be done by returning `false` from the `onSelecting` callback for invalid slots. `TimeGrid` is also configureable to only show certain days (see Week vs Day).

Overall I think that everything you're looking for is already implemented and could be all wrapped up into a custom view component. Let me know if i'm missing any feature here, but I'd be happy to include a BusinessHours addon (like the dnd one that includes such a custom View)

DonovanCharpin:
Hi @jquense, thanks. 

It's not exactly the same way it works between min/max and business hours, it's complementary. The business hours can enable a range of hours so it disables the other ranges from the selection. It's really useful when you manage bookings. For instance, a lot of stores are closed between noon and 2pm. I would be able to just catch the event after the selection and check if there is a collision between the booking and the business hours but I think it's better to just disable between noon and 2pm in order to avoid some misunderstanding (the user could think he can book between those hours). 

For us, it's really a lack in react big calendar because we can't use the calendar without this feature and we would have to use the full calendar instead.

We could add this feature as a plugin but I didn't see the way to implement plugins, the code here is really deep in the core.

jquense:
> I would be able to just catch the event after the selection and check if there is a collision between the booking and the business hours but I think it's better to just disable between noon and 2pm in order to avoid some misunderstanding (the user could think he can book between those hours).

I agree, like I said you don't need to wait until they select, you can prevent the selection event from ever happening (like a native event), by returning `false` from `onSelecting`

> For us, it's really a lack in react big calendar 

Just to be clear, I want you to be able to do this :) I just think you already can without any code (or compromise on your features)

DonovanCharpin:
Sure we can, but not in a proper way.

I'm pretty sure this is not a lack just for us and returning false on a selection is not the right way to disable a range of hours because the user can still select it.

Anyway, it's not a big deal, I propose this feature as an open source contribution if it doesn't fit the vision of the project I can understand. On our side, we will maintain it in another repo.

By the way, if it's implemented in full calendar, I think it's because there is a need from the community.

jquense:
>  returning false on a selection is not the right way to disable a range of hours because the user can still select it.

I don't think you are understanding the API i'm suggesting here. http://www.webpackbin.com/V1Vntx2KG

DonovanCharpin:
Thanks for this example. 

I understand this feature because we use it to do more checks when the user selects a range of hours in the business hours. The only thing is at the first glance, I can't see it's disabled. The user will have to test it by trying to select for each day and see the dynamic results. The business hours can be different from one day to another and it seems really complicated for the user to test until he finds a right selection.

jquense:
Right, I don't think that _zero_ changes are needed to the code here to make this work, but I _do_ want to narrow down the changes to missing API's so that this sort of thing can be implemented as addon easily. We want to try and minimize adding stuff to the base calender that is ultra-specific to one use-case. Its harder to maintain and less flexible.

We can still make general changes that can benefit more people and still solve your specific need. A good example is the Drag & Drop addon, it added a bunch of generic wrapper components, instead of hard-coding dragging styles.

In your case, I think you can make use of the `dayWrapperComponent` to add your css class to show a timeslot as disabled. Besides that is there any other functionality that you need that isn't covered?

DonovanCharpin:
Thanks for the explanation @jquense,

I'll look at the plugin way. Unfortunately, I don't really have more time right now, I spent some hours to fix our issue. We really like the flexbox approach and we prefer to work with react component because our app is 100% ReactJS, that's why I took the time instead of wrapping full calendar in a react component.

We will continuously add features to the react-big-calendar because the calendar is one of the main features of our app to manage bookings (We use it at [Nucli](https://nucli.co) to manage coworking spaces and business centers). 

Right now it covers almost everything we need, it just missing this. Great job for this nice component :)

jquense:
Glad its meeting your needs :) I may have some time to add this properly as an addon myself, since it is in fairly high demand

DonovanCharpin:
Seriously it would be great, you have a better overview of the project and you'll know the possible side effect amongst all properties. In the meantime, we will use our fork to handle our use case.

Thanks for considering it as a new plugin :)

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "react-big-calendar",
+  "name": "@nucli/react-big-calendar",
```
jquense: wrong place ;)

DonovanCharpin: Yes sorry. It's because we created the pull request from our master branch and there is still the link to this PR. You can close it, it should stop notify you.

aliogaili:
That was a very meaningful and desirable feature, I'm not sure why this was not merged.

johnunclesam:
@jquense is this closed? I don't understand. Can we start another one? Or closing this is a mistake?

jquense:
anyone wants to contribute a complete PR adding this go for it. The reason this is closed is evident from the discussion, Please don't spam people with messages demanding folks summarize an issue for you. thanks!

PR #310, opened 2017-02-23
Remove invalid less selector for less4j

favila:
time-grid.less has a dangling child selector (`.rbc-row > {`) which is technically illegal less syntax. The normal js-based less compiler everyone uses will ignore the child selector and emit css anyway, but alternative less compilers (like the java-based less4j) have stricter syntax requirements and error-out when they encounter this construct.

Patch removes the dangling child selector.

### DIFF
--- a/src/less/time-grid.less
+++ b/src/less/time-grid.less
@@ -33,7 +33,7 @@
     z-index: 4;
   }
 
-  .rbc-row > {
+  .rbc-row {
     min-height: 20px;
   }
 }


favila:
I don't know what it *should* be, but what it actually is compiled to today (by less compilers that accept this without error) is `.rbc-row {` not `.rbc-row > * {`. See [this example](http://less2css.org/#%7B%22less%22%3A%22.rbc-time-view%20%7B%5Cn%20%20.rbc-row%20%3E%20%7B%5Cn%20%20%20%20%20%20min-height%3A%2020px%3B%5Cn%20%20%7D%5Cn%7D%22%7D).

jquense:
learn some new about LESS all the time :P thanks!

PR #311, opened 2017-02-24
Fix dnd date merge when moving an event to a date cell

jtomaszewski:
Before (you can see the bug here http://intljusticemission.github.io/react-big-calendar/examples/index.html#api ):

- if you move the 7am event to a different date cell, the time will be reset to 0am

After:

- when moving events to a date cell wrapper, time will be preserved (date will be changed, but time will still be 7am)

This PR fully fixes it.


jquense:
thanks!

Issue #323, opened 2017-03-20
onSelectSlot firing when clicking on the calendar


frankjwu:
In the [selectable example](http://intljusticemission.github.io/react-big-calendar/examples/index.html), clicking on the calendar triggers the onSelectSlot callback. Is there a supported way to *only* do so when selecting spots via click-and-drag?

frankjwu:
I just saw #225 -- I am considering creating a fork for our specific use-case, do you think this is a setting that would be useful upstream?

jquense:
the problem is you really can't distinguish between a "click" and a drag with the implementation since everything is in terms of mouse down/move. there is already a 5px tolerance for clicks but if you want to expose it up so that it's configurable I'd probably merge it if it's not too messy

frankjwu:
Let's say tolerance is configurable: the way to disable the 'click' event from being emitted would then involve setting tolerance to a really large number (ie. Infinity) such that the `isClick` function would always return `false`?

An alternate approach could be to prevent the emission of the 'click' event from ever happening in the mouseUp. Thoughts?

jquense:
ahh yeah you just want to disable clicks? That would make it harder to select a single slot. We have the same behavior as google calendar so I know it's not necessarily odd behavior :P what is the reason why you'd want to disable clicks?

frankjwu:
My calendar is similar to Google Calendar's: clicking on an existing event opens a little tooltip with event details. Clicking outside of the tooltip closes it.

Right now, clicking outside of the tooltip to close it also triggers the `onSelectSlot` callback. Google Calendar has the same behavior, but I don't really like that it happens that way 😛

frankjwu:
Proposed solution -- update the object passed to the `onSelectSlot` callback:
```javascript
function(
  slotInfo: object {
    start: Date,
    end: Date,
    slots: array<Date>,
    action: String,
  }
)
```

where `action` is one of `click` or `select`. The user can then handle the two cases as desired. In my case, my `onSelectSlot` function would ignore all cases where `action === 'click'`.

Thoughts? Will put together the PR if you agree!

jquense:
perfect, nice and simple and really minimal change. I'd be happy to merge that

PR #324, opened 2017-03-20
Pass mouse event action to onSelectSlot() callback

frankjwu:
An `action` of `'click'` or `'select'` is now included in the object passed to the `onSelectSlot` callback.

Addresses #323 -- I tried to touch as little code as possible

### DIFF
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -103,7 +103,7 @@ let Calendar = React.createClass({
      *     start: Date,
      *     end: Date,
      *     slots: array<Date>,
-     *     action: string
+     *     action: "select" | "click"
      *   }
      * )
      * ```
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -102,7 +102,8 @@ let Calendar = React.createClass({
      *   slotInfo: object {
      *     start: Date,
      *     end: Date,
-     *     slots: array<Date>
+     *     slots: array<Date>,
+     *     action: string
      *   }
      * )
      * ```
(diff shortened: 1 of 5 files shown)


src/Calendar.js
```diff
@@ -102,7 +102,8 @@ let Calendar = React.createClass({
      *   slotInfo: object {
      *     start: Date,
      *     end: Date,
-     *     slots: array<Date>
+     *     slots: array<Date>,
+     *     action: string
```
jquense: can you change this to `"select" | "click"` so it's clearer what these values are?

src/DayColumn.js
```diff
@@ -271,15 +271,23 @@ let DaySlot = React.createClass({
...
       })
 
     selector
       .on('select', () => {
         if (this.state.selecting) {
-          this._selectSlot(this.state)
+          this._selectSlot(Object.assign({},
+            this.state, {
+              action: 'select'
+            }
```
jquense: `{ ...this.state, action: 'select' }` is a bit neater instead of the Object.assign

src/DayColumn.js
```diff
@@ -271,15 +271,23 @@ let DaySlot = React.createClass({
     selector
       .on('click', (box) => {
         if (!isEvent(findDOMNode(this), box))
-          this._selectSlot(selectionState(box))
+          this._selectSlot(Object.assign({},
+            selectionState(box), {
+              action: 'click'
+            }
```
jquense: same as bellow

jquense:
looks good, just a few notes. Thanks!

minznerjosh:
Any estimate on when this will make it into a release?

PR #328, opened 2017-03-23
Make previous message more obvious to translate

tophsic:
Why previous message is the only message with a different default translation? It can be easier to translate messages if default translation is 'previous', isn't it?

### DIFF
--- a/src/utils/messages.js
+++ b/src/utils/messages.js
@@ -8,7 +8,7 @@ import invariant from 'invariant';
   week: 'week',
   day: 'day',
   month: 'month',
-  previous: 'back',
+  previous: 'previous',
   next: 'next',
   yesterday: 'yesterday',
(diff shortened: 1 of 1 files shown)


jquense:
not sure why it be easier to translate...why do you say that?

tophsic:
As I translated messages to french, I looked very quickly at the documentation how to do it. So I started to add `messages` option to my BigCalendar component using strings I saw in english interface as keys. So I tried `back` key instead of `previous`.

If I read more carefully or even copied messages list from documentation I wouldn't have done the mistake!

I know that this is not a very important change. Do not hesitate to close the PR ;)

jquense:
thanks for the feedback, I'm gonna leave it be for the time being since i'm not sure if it's worth the change right now. but will consider it in the future!

PR #332, opened 2017-03-24
Add option 'showAllDayEvents' + fix invalid gutter in measureGutter

mattrick:
Hello, i've added option (mainly for myself) to hide all day events + quick fix in measuring gutter, when there's no such panel (it get's rid of non-numeric values in gutters array and then selects max ).

### DIFF
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -210,6 +210,11 @@ let Calendar = React.createClass({
     step: React.PropTypes.number,
 
     /**
+     * Show all day events
+     */
+    showAllDayEvents: React.PropTypes.bool,
+
+    /**
      * The number of slots per "section" in the time grid views. Adjust with `step`
      * to change the default of 1 hour long groups, with 30 minute slots.
      */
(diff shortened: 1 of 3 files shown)


src/TimeGrid.js
```diff
@@ -348,7 +352,7 @@ export default class TimeGrid extends Component {
     let gutterCells = this._gutters;
 
     if (!width) {
-      width = Math.max(...gutterCells.map(getWidth));
+      width = Math.max(...gutterCells.map(getWidth).filter(Boolean));
```
jquense: this will give weird results when there are no truthy results. e.g `Math.max(...[0,0,0].filter(Boolean)) // -Infinity` which in javascript land is truthy, and would set the gutterWidth to `-Infinity`

src/Calendar.js
```diff
@@ -210,6 +210,11 @@ let Calendar = React.createClass({
     step: React.PropTypes.number,
 
     /**
+     * Show all day events
+     */
+    showAllDayEvents: React.PropTypes.bool,
```
jquense: I don't love this option because it will cause all and multi day events to just disappear in Week or Day views. Can you talk a bit why you'd want that?

PR #337, opened 2017-03-28
DnD: startAccessor and endAccessor support

minznerjosh:
The DnD addon was reading event start/end times but wasn't using the startAccessor/endAccessor.

### DIFF
--- a/src/addons/dragAndDrop/backgroundWrapper.js
+++ b/src/addons/dragAndDrop/backgroundWrapper.js
@@ -2,10 +2,12 @@ import React from 'react'
 import { DropTarget } from 'react-dnd'
 import cn from 'classnames';
 
+import { accessor } from '../../utils/propTypes';
+import { accessor as get } from '../../utils/accessors';
 import dates from '../../utils/dates';
 import BigCalendar from '../../index'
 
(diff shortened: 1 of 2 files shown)


src/addons/dragAndDrop/backgroundWrapper.js
```diff
@@ -107,11 +111,15 @@ function createWrapper(type) {
...
-      const { onEventDrop } = context
+      const { onEventDrop, startAccessor, endAccessor } = context
 
       onEventDrop({
         event,
-        ...getEventTimes(event, value, type)
+        ...getEventTimes({
+          ...event,
+          start: get(event, startAccessor),
+          end: get(event, endAccessor)
```
minznerjosh: `getEventTimes()` was only looking at the `start`/`end` time so spreading `...event` wasn't strictly necessary. Only did it to minimize the change.

jquense: i'd  change the signature of `getEventTimes` and avoid the spread

jquense:
nice, thanks

minznerjosh:
Thank YOU for this library in the first place. It's been super great working with it.

minznerjosh:
@jquense I changed the signature.

PR #340, opened 2017-03-28
isSibling Fix

eistre91:
The isSibling function introduced previously into dayViewLayout.js had a hardcoded value of an hour. This works fine for the default calendar with step 30 and timeslots 2, but it doesn't make much sense (and doesn't look great) for other steps and timeslots. 

It seems like what works here is step*timeslots.


jquense:
nice! thanks I'll try and review it soon

src/DayColumn.js
```diff
@@ -140,12 +140,14 @@ let DaySlot = React.createClass({
       , eventWrapperComponent: EventWrapper
       , rtl: isRtl
       , step
+      , timeslots
```
tobiasandersen: This will be `undefined ` (and mess up the styling) unless it's explicitly added as a prop to the Calendar. The default value of 2 seems to be set as default only in TimeColumn and TimeSlotGroup.

jquense: sounds good to me

eistre91: Setting timeslots at TimeGrid.defaultProps will work I think. It's a parent to both TimeColumn and DaySlot, and TimeColumn is a parent for TimeSlotGroup. (But I'm not 100% on React yet so I'm not sure.)

tobiasandersen: Sounds good!

jquense:
ping @eistre91

PR #341, opened 2017-03-28
More Functionality from eventPropGetter

eistre91:
Add support for custom positioning returned from eventPropGetter.

It's come up in my own work with this library that I wanted to be able to change the way that events are positioned. It doesn't work for my application to have them overlap as they currently do. This is a bit of a hacky fix...I'm sure there's probably a better way to do it, but I wanted to propose the change and open up discussion about it.

### DIFF
--- a/src/DayColumn.js
+++ b/src/DayColumn.js
@@ -162,7 +162,7 @@ let DaySlot = React.createClass({
       if (eventPropGetter)
         var { style: xStyle, className } = eventPropGetter(event, start, end, _isSelected)
 
-      let { height, top, width, xOffset } = style
+      let { height, top, width, xOffset} = Object.assign({}, style, xStyle)
 
       return (
         <EventWrapper event={event} key={'evt_' + idx}>


eistre91:
I should specify that what I'm allowing to be changed from eventPropGetter are the following style attributes on events:
height, width, top, and xOffset

jquense:
hey, thanks for the PR. I can understand your desire here, however I think this provides a foot gun for most ppl. I'd prefer to not allow overriding these properties, especially since you don't really even have the context here implement a different layout strategy. 

I'm not necessarily opposed to making the layout bits pluggable if it's not too complex, but I'm also concerned about maintenance burden and interoperability with other features

eistre91:
Your concern makes sense. Perhaps a more straightforward change would be allowing passed in values for OVERLAP_MULTIPLIER in dayViewLayout.cljs?

tobiasandersen:
To me this sounds reasonable. That value doesn't affect anything in the layout algorithm, it's just for adjusting overlapping once the positions have already been calculated. Setting it to 0, would result in no overlapping, while 1 would make events overlap their left sibling entirely (probably not what you want).

Issue #350, opened 2017-04-06
Two HTML5 Drag and drop backends


martinnov92:
Hello,
does anybody have idea how to use react-big-calendar with drag and drop and my own component with drag and drop? Now when I use both of these components on the same page I get the error about two HTML5 backends, the only (wrong) solution I came with was to delete HTML5 backend from RBC drag and drop index.js...

Does anyone have some ideas? Or would be possible to publish this package with optional use of drag and drop backend? I would like to use touch backend anyway...

Thank you very much

martinnov92:
Here you can have a look at my little project (still work in progress so it is messy) https://github.com/martinnov92/TSCalendar

jquense:
take a look at the addon code you can provide your own backend optionally

martinnov92:
Hello @jquense , thanks for your answer, I know about that option, but I don't want to provide anything, as I will provide backend from higher component...

martinnov92:
when I delete this `DragDropContext(backend)` from dragAndDrop/index.js, it works perfectly and that's how I need it :D

jquense:
I'm not an expert but why don't you share the backend with both your component and the drag and drop?

martinnov92:
@jquense I am not expert either :D, so I am not sure what do you think to share backend. When I keep them both I get the error about 2 backends on same page, when I remove it from mine - mine is not working, when I remove it from RBC - it works. You can have a look here and see how it looks/work https://martinnov92.github.io/TSCalendar/. RBC is almost on the bottom of my component tree and I would like to share the backend from the component on very top (<App>), but I need to remove it from RBC first :/. Hope it makes sense, sorry about my english.

martinnov92:
Hello again, I created a pull request https://github.com/intljusticemission/react-big-calendar/pull/352 with working solution. I hope this will get merge as I think it would make this component more reusable in case you want to use React DND in other components in your app...
Thank you

PR #352, opened 2017-04-06
Use withDragAndDrop addon without DnDContext

martinnov92:
Hello everybody,

this pull request is for this issue https://github.com/intljusticemission/react-big-calendar/issues/350 , where once you have multiple HTML5Backend or Touch backends in your app, you can't use your app properly as you can't have 2 or more HTML5Backends at once.
This restriction about HTML5backend is documented, but it is not solved very well yet (https://github.com/react-dnd/react-dnd/issues/186). 

This pull request solves it by providing an option to choose if you want to use RBC with DragAndDrop context or you will provide your context in higher order component => you can use it like this: 
`const DragAndDropCalendar = withDragAndDrop(BigCalendar, { withDnDContext: false });`. 

This request does not contain any breaking changes.

I hope this will be useful for anyone else and not just me 😄 . 

Thank you and please merge this request 😄 .

### DIFF
--- a/src/addons/dragAndDrop/index.js
+++ b/src/addons/dragAndDrop/index.js
@@ -14,7 +14,8 @@ try {
 
 
 export default function withDragAndDrop(Calendar, {
-  backend = html5Backend
+  backend = html5Backend,
+  withDnDContext = true
 } = {}) {
 
   class DragAndDropCalendar extends React.Component {
@@ -101,5 +102,9 @@ export default function withDragAndDrop(Calendar, {
     endAccessor: accessor
   }
 
-  return DragDropContext(backend)(DragAndDropCalendar);
+  if (withDnDContext) {
+    return DragDropContext(backend)(DragAndDropCalendar);
+  } else {
+    return DragAndDropCalendar;
+  }
 }
--- a/src/addons/dragAndDrop/index.js
+++ b/src/addons/dragAndDrop/index.js
@@ -14,8 +14,7 @@ try {
 
 
 export default function withDragAndDrop(Calendar, {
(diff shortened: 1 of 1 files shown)


src/addons/dragAndDrop/index.js
```diff
@@ -101,5 +102,9 @@ export default function withDragAndDrop(Calendar, {
     endAccessor: accessor
   }
 
-  return DragDropContext(backend)(DragAndDropCalendar);
+  if (withDnDContext) {
```
jquense: maybe do this when `backend` is `false` (or `null`) and then we don't need to roll in a new option? what do you think?

martinnov92: Yes that seems like a better idea, I updated the code according your comment :)

martinnov92:
@jquense that seems like a better idea, I updated the code according your comment :)

martinnov92:
@jquense Thank you for approve 🙏 Do you have any idea when this will be merged? Thank you

martinnov92:
Hello, does anybody know when this could be merged? We will need this in our project soon :/. Thank you.

arecvlohe:
It looks like this is merged with master but not part of the current NPM package. I will figure out how to use your solution @martinnov92. Thanks!

arecvlohe:
There is something off about this merge: https://github.com/intljusticemission/react-big-calendar/pull/352/commits. It seems as if you made the changes and then reverted them @martinnov92. So @jquense essentially merged no changes. I think that's what happening here. Thoughts?

jquense:
@arecvlohe it just hasn't been released yet

PR #365, opened 2017-04-24
Add a class with the weekday

gabrielpoca:
This change adds a class with the weekday, similar to the
`rbc-today`. The purpose of this class is to allow us to customize the
cells by day of the week. For instance, to customize weekends we could
do:

```
.rbc-weekday-0,
.rbc-weekday-6 {
  background: red;
}
```

I did not write any default styles to avoid breaking the API or the
default behavior and appearance. Hopefully, this will make it easier to
merge this change.

Please let me know what else I need to do to add this functionality.

### DIFF
--- a/src/BackgroundCells.js
+++ b/src/BackgroundCells.js
@@ -68,6 +68,7 @@ class BackgroundCells extends React.Component {
                   'rbc-day-bg',
                   selected && 'rbc-selected-cell',
                   dates.isToday(date) && 'rbc-today',
+                  `rbc-weekday-${dates.weekday(date)}`,
                 )}
               />
             </Wrapper>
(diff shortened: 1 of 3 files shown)


jquense:
its fine code-wise, but i'm sure that it'll provide you the hooks you actually want. "Day of the week" is locale specific, I'd imagine the goal is to do something like "make weekends grayed out" right? You'll run into an issue because in `en` weekday-0` is sunday but in `en-GB` it's monday.

gabrielpoca:
@jquense makes sense 👍  thank you for replying. What about using the name instead of the number? Such as `rbc-weekday-monday` and `rbc-weekday-sunday`. Would it make sense? And is this something that you would merge?

jquense:
sure that sounds reasonable

PR #371, opened 2017-04-30
Add an optional property for initially selected event

mkretz:
I am using big calendar on a view which the user jumps to by clicking on an url in an email. When the calendar is displayed I want the event referenced in the url to be selected. 

My proposed solution: add an optional property `initialSelection` to `Calendar`. The property contains one element of `events`. If the property is set, `handleSelectEvent` is called after the component has mounted with this element as argument.

### DIFF
--- a/examples/demos/basic.js
+++ b/examples/demos/basic.js
@@ -8,6 +8,7 @@ let Basic = React.createClass({
       <BigCalendar
         {...this.props}
         events={events}
+        initialSelection={events[1]}
         defaultDate={new Date(2015, 3, 1)}
       />
     )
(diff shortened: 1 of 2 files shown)


jquense:
Hi there, thanks for the PR! I think the feature you want is already there fortunately. The prop is defaultSelected (like value and defaultValue on react inputs)

mkretz:
You're right, it works! Thanks for the hint! Will close this PR. Just out of interest: I cannot find this behaviour when going through the code. How is this implemented?

jquense:
It's handled via the uncontrollable library used in Calendar

mkretz:
Ok, I will have closer look at uncontrollable. Closing the PR now. Thanks!

Issue #378, opened 2017-05-04
About event accuracy and overlap


niconchow:
I don‘t know why.The event's time is **inaccuracy**.They have a **deviation**!
And I want to set the event in hot pursuit.But the events are overlap.
![2](https://cloud.githubusercontent.com/assets/18158336/25689187/48a7f3a4-30b8-11e7-8978-a23f59b32aad.jpg)

niconchow:
@jquense can you tell me the reason? thanks!

jquense:
What? Sorry I don't understand what you're asking

niconchow:
@jquense  The position is offset.I don't know what effect
![1](https://cloud.githubusercontent.com/assets/18158336/25689871/4a14e458-30be-11e7-84e6-c5aad746cb91.jpg)

niconchow:
The startTime and endTime are inaccuracy.
If the two events are similar, they shrink, and I don't want that effect.
I want them not to shrink.
![2](https://cloud.githubusercontent.com/assets/18158336/25692148/fbc917fa-30d2-11e7-9cab-440986e97648.jpg)

jquense:
The overlap because technically the events overlap, in that the end of eventA is the same as the start of eventB. I agree it probably shouldn't do that tho in this case.

bkniffler:
I agree too, when an event ends at the exact same time another starts, its not really overlapping.

jquense:
PR's Welcome :) should be easy enough to address

bkniffler:
Okay, https://github.com/intljusticemission/react-big-calendar/pull/379 works for my overlaps. But I don't know if its right for you, please check @jquense

tobiasandersen:
I remember working on this, and in the end went with something similar to what Google Calendar have. I believe they consider events overlapping if they end/start within < 15 minutes. I guess the reason being that your events could really blur together into one big mass if you have a lot of them (which we do). 

I know there was a discussion, in some issue or PR, about making the layout algorithm pluggable. Not sure whether that's a good idea or not, but people seem to want different things here, and that would be a way to address that.

bkniffler:
I don't really agree to events ending up blurring into each other, since you can separate them by the borders already. The big issue here is that in some scenarios, overlaps need to be avoided, and by showing events that do not really overlap the same same way like overlapping events, you are confusing the users and you are making it harder to really make them aware of overlaps. 

And of course, if you have a lot of events, the resulting calendar will look all over the place.

[...]

PR #379, opened 2017-05-04
Siblings = overlapping events

bkniffler:
Change the sibbling detection and recognize all overlapping events

### DIFF
--- a/src/utils/dayViewLayout.js
+++ b/src/utils/dayViewLayout.js
@@ -36,7 +36,7 @@ let getSlot = (event, accessor, min, totalMin) => event && positionFromDate(
 /**
  * Two events are considered siblings if their times are overlapping
  */
-let isSibling = (idx1, idx2, { events, startAccessor, min, totalMin }) => {
+let isSibling = (idx1, idx2, { events, startAccessor, endAccessor, min, totalMin }) => {
   let event1 = events[idx1]
   let event2 = events[idx2]
 
--- a/src/utils/dayViewLayout.js
+++ b/src/utils/dayViewLayout.js
@@ -36,25 +36,19 @@ let getSlot = (event, accessor, min, totalMin) => event && positionFromDate(
 /**
  * Two events are considered siblings if their times are overlapping
  */
-var isSibling = function isSibling(idx1, idx2, _ref2) {
-  var events = _ref2.events,
-      startAccessor = _ref2.startAccessor,
-      endAccessor = _ref2.endAccessor,
-      min = _ref2.min,
-      totalMin = _ref2.totalMin;
+let isSibling = (idx1, idx2, { events, startAccessor, min, totalMin }) => {
+  let event1 = events[idx1]
+  let event2 = events[idx2]
 
-  var event1 = events[idx1];
-  var event2 = events[idx2];
+  if (!event1 || !event2) return false
 
-  if (!event1 || !event2) return false;
+  var start1 = getSlot(event1, startAccessor, min, totalMin)
+  var start2 = getSlot(event2, startAccessor, min, totalMin)
+  var end1 = getSlot(event1, endAccessor, min, totalMin)
+  var end2 = getSlot(event2, endAccessor, min, totalMin)
 
-  var start1 = getSlot(event1, startAccessor, min, totalMin);
-  var start2 = getSlot(event2, startAccessor, min, totalMin);
-  var end1 = getSlot(event1, endAccessor, min, totalMin);
-  var end2 = getSlot(event2, endAccessor, min, totalMin);
-
-  return (start2 <= start1 && end2 > start1) || (start2 >= start1 && start2 < end1);
-};
+  return (start2 <= start1 && end2 > start1) || (start2 >= start1 && start2 < end1)
+}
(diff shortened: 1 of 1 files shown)


src/utils/dayViewLayout.js
```diff
@@ -34,20 +34,27 @@ let getSlot = (event, accessor, min, totalMin) => event && positionFromDate(
...
  */
-let isSibling = (idx1, idx2, { events, startAccessor, min, totalMin }) => {
-  let event1 = events[idx1]
-  let event2 = events[idx2]
+var isSibling = function isSibling(idx1, idx2, _ref2) {
+  var events = _ref2.events,
+      startAccessor = _ref2.startAccessor,
+      endAccessor = _ref2.endAccessor,
+      min = _ref2.min,
+      totalMin = _ref2.totalMin;
```
jquense: this looks like babel output...

bkniffler:
Urgh, sorry bout the code, I've been working on it through my node_modules to make sure it works, obviously been using the compiled source. Should be okay now.

bkniffler:
Just as a comparison (with 1px white border on bottom):

Before:
xyz

After:
xy

It would be supercool if this could be merged soon, its one of the things currently holding me up to give access to testers.

tobiasandersen:
I agree that the "after" screenshot you posted looks better, but this PR is gonna affect more complex layouts quite a lot. See e.g. the stories I created when implementing this:

**Set 0**

Before:
0-current

After:
0-pr

**Set 1**

Before:
1-current

After:
1-pr

I'd prefer if we could keep those layouts as is (since you can fit much more), but adjust for the situation where the end and start is the same (so that we get your after layout). Something like this should do just that: 

```
let isSibling = (idx1, idx2, { events, startAccessor, endAccessor, min, totalMin }) => {
  let event1 = events[idx1]
  let event2 = events[idx2]

  if (!event1 || !event2) return false

  let start1 = getSlot(event1, startAccessor, min, totalMin)
  let start2 = getSlot(event2, startAccessor, min, totalMin)
  let end1 = getSlot(event1, endAccessor, min, totalMin)

  return (Math.abs(start1 - start2) < 60 && start2 < end1)
}

```

That `60` number is still quite arbitrary, but will be fixed with https://github.com/intljusticemission/react-big-calendar/pull/340/commits/ef0017975f53e03c79937fed2ed854a95cdea9f8.

src/utils/dayViewLayout.js
```diff
@@ -34,19 +34,20 @@ let getSlot = (event, accessor, min, totalMin) => event && positionFromDate(
...
-let isSibling = (idx1, idx2, { events, startAccessor, min, totalMin }) => {
+let isSibling = (idx1, idx2, { events, startAccessor, endAccessor, min, totalMin }) => {
   let event1 = events[idx1]
   let event2 = events[idx2]
 
   if (!event1 || !event2) return false
 
-  let start1 = getSlot(event1, startAccessor, min, totalMin)
-  let start2 = getSlot(event2, startAccessor, min, totalMin)
+  var start1 = getSlot(event1, startAccessor, min, totalMin)
```
tobiasandersen: Small knit pick, but can we keep `let`?

jquense:
Ha I missed all that when I looked. Thanks for double checking :)

bkniffler:
Hey @tobiasandersen, your code works fine for my requirements, so I updated my PR. The `var`s shouldn't be an issue anymore.

AgamlaRage:
@bkniffler I tried your PR (utils/dayViewLayout.js) modifications, works fine for me. 
Can confirm that fixed the issue #378 for me.

I have an other behavior that is unexpected with event overlaping, I post it here because I think it's related with this PR. Here the save events.

**GoogleCalendar/Fullcalendar**

rbc-issue-fullcalendar-xs

**React Big Calendar**

rbc-issue-react-big-calendar-xs

As I say in the image, **Blue event should be bellow the red one, and stop at the same position that the yellow one**.

My question is:

Is that plan to have the exact same behavior as GoogleCalendar/Fullcalendar ?
If yes, Is that related to this PR ?
If not, should I open a new issue ?

Thx

bkniffler:
It seems like big calendar tries to follow google calendar in terms of look&feel, which is good since the google devs surely put lots of research into their calendar. I bet the maintainers here would be willing to accept a PR on this matter.

There is 2 things that could to be addressed somewhere around [here](https://github.com/bkniffler/react-big-calendar/blob/4ed1442f9ca69227d16e22c91c475c2076054cba/src/utils/dayViewLayout.js#L196) to imitate gcal:
- z-index of sibling events needs to be raised (through inline styles or via an additional class `rdc-sibling-event`) or rendering of these events needs to happen after non-sibling ones.
- only the left offset should vary with `siblingIdx` rising, while still trying to fill the whole width. Currently the algorithm will make the yellow 70% width/0 offset. It could be better to follow google calendar and make the yellow 100% width/0 offset, the red 70% width/30% offset. 

Give it a shot, shouldn't be too hard!

AgamlaRage:
Ok thx ! I'll do that
I open a new issue #404  to ge the tracker clean.

Issue #357, opened 2017-04-14
events in month view do not display correctly


kaijajan:
I tried to create some long events across days. But some events did not display correctly.

`
let testEvents = [
    {
      "start": new Date(2017, 3, 4),
      "end": new Date(2017, 3, 7),
      "title":"Event 1"
    },
    {
      "allDay":true,
      "start": new Date(2017, 3, 13),
      "end":new Date(2017, 3, 16),
      "title":"Event 2"
    },
    {
      "allDay":true,
      "start":new Date(2017, 3, 22),
      "end":new Date(2017, 3, 25),
      "title":"Event 3"
    }
  ];

BigCalendar.setLocalizer(
  BigCalendar.momentLocalizer(moment)
);

class App extends Component {
  render() {
    return (
      <div>
        <div>
          <BigCalendar
            defaultDate={ new Date() }
            events={testEvents}
            views={['month']}
          />
        </div>
      </div>
    );
  }
}
`
I got the result as the link below. All events were all three 3 days long event. But 1 of them showed 4 days across. And all of the were built in the same rule
http://imgur.com/QET5NMf
Any ideas for this issue?

apn-jun:
got blank under month, any update?

john-hadron:
Got a blank under month and agenda too

dpkg:
https://github.com/intljusticemission/react-big-calendar/pull/380 fixes it. 

not very throughly tested though, will appreciate if anyone can test if their calendars are showing correct data with my forked branch :)

jquense:
Thanks for the fix. To test you can create and run a few stories testing different event setups. See: `npm run storybook` and the stories directory

dpkg:
thanks. already tested with some test events in my calendar using the default popup demo. going to do what you suggested above now.

bkgrubb:
@apn-jun, @john-hadron I had this problem too until I read the docs a bit more thoroughly :)

"note: The default styles use height: 100% which means your container must set an explicit height (feel free to adjust the styles to suit your specific needs)."

john-hadron:
@bkgrubb Yup, that was my issue too :+1:

dpkg:
I guess the original issue as mentioned in the very first comment was something else (see screenshot). The events ending at time 00:00:00 are not rendered on the specified end date, as designed. But they do render if the end date is the first day of the week. 

Do I understand wrong?

dpkg:
[...]

PR #380, opened 2017-05-05
#357: event should end after week start to be eligible to be displayed in that week

dpkg:
fixes https://github.com/intljusticemission/react-big-calendar/issues/357

### DIFF
--- a/src/Month.js
+++ b/src/Month.js
@@ -18,10 +18,10 @@ import DateContentRow from './DateContentRow'
 import Header from './Header'
 
 import { accessor, dateFormat } from './utils/propTypes'
-import { segStyle, inRange, sortEvents } from './utils/eventLevels'
+import { segStyle, within, sortEvents } from './utils/eventLevels'
 
 let eventsForWeek = (evts, start, end, props) =>
-  evts.filter(e => inRange(e, start, end, props))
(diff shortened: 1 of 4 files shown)


src/Month.js
```diff
@@ -18,10 +18,10 @@ import DateContentRow from './DateContentRow'
 import Header from './Header'
 
 import { accessor, dateFormat } from './utils/propTypes'
-import { segStyle, inRange, sortEvents } from './utils/eventLevels'
+import { segStyle, within, sortEvents } from './utils/eventLevels'
 
 let eventsForWeek = (evts, start, end, props) =>
-  evts.filter(e => inRange(e, start, end, props))
+  evts.filter(e => within(e, start, end, props))
```
dpkg: the method `inRange()` operates only on the "day" part of the date, so when trying to figure out if an event is in the week range, we're not able to deduce if the event end date is great than the week start date. 

this new method checks for the whole date if it's within the given range.

tobiasandersen:
Hey @dpkg, thanks for doing this, and sorry for not getting back to you on it! This stuff has recently been changed quite a bit by @jquense, so I'm not sure what to do here. Is it still an issue for you?

jquense:
We've made improvements here and I think we cover this case, thanks again tho for the PR!

PR #381, opened 2017-05-05
Updated react-overlays to ^0.7.0.

jasonleibowitz:
React v15.5 deprecates using `React.PropTypes` and instead you must import from `prop-types`. Also, `React.createClass` is deprecated and will be removed in version 16. 

This library has a dependency on react-overlays, which only fixes these issues in `0.7.0`. This PR allows for this new update to avoid these warnings/errors when using React Big Calendar.

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,7 +89,7 @@
     "invariant": "^2.1.0",
     "lodash": "^4.17.4",
     "prop-types": "^15.5.8",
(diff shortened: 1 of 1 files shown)


jquense:
Thanks!

MichaelDeBoey:
@jquense Any news when this will be released? 🙂

PR #383, opened 2017-05-08
Implement WorkWeek view

jquense:
It was easy enough so I’ll just add it since its often asked for


package.json
```diff
@@ -48,39 +48,29 @@
...
-    "css-loader": "^0.15.6",
     "eslint": "^3.0.0",
     "eslint-config-jason": "^4.0.0",
     "eslint-import-resolver-webpack": "^0.8.0",
     "eslint-plugin-import": "^2.0.0",
     "eslint-plugin-react": "^6.8.0",
-    "file-loader": "^0.8.4",
     "globalize": "^0.1.1",
-    "html-webpack-plugin": "^1.7.0",
-    "less": "^2.5.1",
```
xkb: @jquense why was this removed? `npm run less` is failing.

PR #393, opened 2017-05-12
fix(package.json): fix peerDependencies

jedwards1211:
It had react `0.14.9`, which doesn't exist as far as I can tell.  The last release before `15.0.0` was `0.14.8`.

Is there any reason that it's not `|| ^15.0.0` instead of `|| ^15.3.0`?

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,8 +40,8 @@
     "release": "release"
   },
   "peerDependencies": {
-    "react": "^0.14.9 || ^15.3.0",
-    "react-dom": "^0.14.9 || ^15.3.0"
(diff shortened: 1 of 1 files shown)


jquense:
thanks for the PR howver there is a reason, https://github.com/facebook/prop-types/blob/master/README.md#compatibility

jedwards1211:
why doesn't 0.14.9 show up in `npm view react`?

jedwards1211:
@jquense nevermind, sorry, I missed it because I was looking at the `time` section which is ordered by release date instead of version.

Issue #384, opened 2017-05-08
Warning on Event Click


ardsouza:
warning.js:36 Warning: Failed prop type: Invalid prop `selected` of type `object` supplied to `EventRow`, expected `array`.
    in EventRow (created by DateContentRow)
    in DateContentRow (created by TimeGrid)
    in div (created by TimeGrid)
    in div (created by TimeGrid)
    in div (created by TimeGrid)
    in TimeGrid (created by Week)
    in Week (created by Calendar)
    in div (created by Calendar)
    in Calendar (created by Uncontrolled(Calendar))
    in Uncontrolled(Calendar) (created by DragAndDropCalendar)
    in DragAndDropCalendar (created by DragDropContext(DragAndDropCalendar))
    in DragDropContext(DragAndDropCalendar) (created by CalendarContainer)

juanibiapina:
`DateContentRow` has a `selected` prop of type `object` that is just passes through to `EventRow`, which expects it to be an `array`.

johnbian:
@ardsouza Has this problem been solved?

smtheard:
@jasonleibowitz seems to have implemented a fix, but I am still getting this warning in the most recent version from NPM. Any word on when this will be available on NPM?

smilexu:
I get same issue on 0.14.0

jasonleibowitz:
The fix won't be included in the version you download from npm until there is a new release. The latest release is 0.14.0, which was published on May 2. 

If you _need_ the fix before then you can install this package from github:master rather than npm, but then you'll have to run the build script in the repo's folder manually. That is definitely not the cleanest way of doing this, but I don't think you have many other options without a new release being made.

jquense:
Just ignore until there is a new release. It's a prop type warning, it's not an actual issue with any code.

PR #395, opened 2017-05-15
#384 Resolve PropTypes Warning on Event Click

jasonleibowitz:
#### Resolves
- #384 - Warning on Event Click

#### Notes
- Updated PropTypes declaration of selected in EventRowMixin as well as the default propType of selected to be an object


jquense:
Thanks!

Issue #224, opened 2016-12-06
Render icon beside date (number)


lawlessnut:
![screen shot 2016-12-05 at 4 15 44 pm](https://cloud.githubusercontent.com/assets/6776653/20907970/47c1ee54-bb06-11e6-8d1a-3853db0dd121.png)

Is it possible to render icons beside date. Example here as you can see i have weather icons beside number.

jquense:
not at the moment but i'd happily accept a PR adding a configurable component for it. something similar to https://github.com/intljusticemission/react-big-calendar/pull/196

andresgutgon:
I think that can work for your use case ☝️

PR #402, opened 2017-05-20
Adding optional custom DateHeader component for month view

andresgutgon:
Trying to solve this issue https://github.com/intljusticemission/react-big-calendar/issues/224

### DIFF
--- a/src/BackgroundCells.js
+++ b/src/BackgroundCells.js
@@ -62,7 +62,12 @@ class BackgroundCells extends React.Component {
         {range.map((date, index) => {
           let selected =  selecting && index >= startIdx && index <= endIdx;
           return (
-            <Wrapper key={index} value={date}>
+            <Wrapper
+              key={index}
+              value={date}
+              isFirstInRange={index === 0}
+              isLastInRange={index === range.length - 1}
+            >
               <div
                 style={segStyle(1, range.length)}
                 className={cn(
--- a/src/BackgroundCells.js
+++ b/src/BackgroundCells.js
@@ -65,8 +65,7 @@ class BackgroundCells extends React.Component {
             <Wrapper
               key={index}
               value={date}
-              isFirstInRange={index === 0}
-              isLastInRange={index === range.length - 1}
+              range={range}
             >
               <div
                 style={segStyle(1, range.length)}
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -131,7 +131,6 @@ class Calendar extends React.Component {
     * ```
     */
    onSelecting: PropTypes.func,
-     
    /**
     * The selected event, if any.
     */
@@ -420,6 +419,7 @@ class Calendar extends React.Component {
      }),
      month: PropTypes.shape({
        header: elementType,
+       dateHeader: elementType,
        event: elementType
      })
    }),
--- a/src/Month.js
+++ b/src/Month.js
@@ -194,7 +194,7 @@ class MonthView extends React.Component {
     )
   }
 
-  readerDateHeading = ({ date, className, isFirstInRange, isLastInRange, ...props }) => {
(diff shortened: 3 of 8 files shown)


src/Calendar.js
```diff
@@ -412,14 +411,17 @@ class Calendar extends React.Component {
 
      day: PropTypes.shape({
        header: elementType,
+       dateHeader: elementType,
        event: elementType
      }),
      week: PropTypes.shape({
        header: elementType,
+       dateHeader: elementType,
```
jquense: This isn't relevant for Day or Week views

src/Month.js
```diff
@@ -86,6 +87,7 @@ class MonthView extends React.Component {
       rowLimit: 5,
       needLimitMeasure: true,
     }
+    this.handleHeadingClick = this.handleHeadingClick.bind(this)
```
jquense: Not necessary since you close over with an arrow function below

src/Month.js
```diff
@@ -215,16 +218,10 @@ class MonthView extends React.Component {
...
-            >
-              {label}
-            </a>
-          : <span>
-              {label}
-            </span>}
+        <DateHeaderComponent
+          label={label}
+          drilldownView={drilldownView}
+          onDateClick={e => this.handleHeadingClick(date, drilldownView, e)} />
```
jquense: Maybe onDrilldown

examples/demos/customDateHeader.js
```diff
@@ -0,0 +1,48 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import BigCalendar from 'react-big-calendar';
+import events from '../events';
```
jquense: No need for an example, but please add a story that demonstrates the functionality (in /stories)

andresgutgon:
Thanks for reviewing! I'll address your comments

andresgutgon:
Comments fixed :)

andresgutgon:
### Offtopic:
I saw double `less` command here: https://github.com/intljusticemission/react-big-calendar/blob/master/package.json#L32

It's ok?

src/BackgroundCells.js
```diff
@@ -62,7 +62,12 @@ class BackgroundCells extends React.Component {
         {range.map((date, index) => {
           let selected =  selecting && index >= startIdx && index <= endIdx;
           return (
-            <Wrapper key={index} value={date}>
+            <Wrapper
+              key={index}
+              value={date}
+              isFirstInRange={index === 0}
+              isLastInRange={index === range.length - 1}
```
jquense: ? I'd prefer to not pass arbitrary props to this component

andresgutgon: There is a reason for it. Maybe you can give me an alternative.
On our calendar when use clicks on a date we open a popover. If the day is the first one in the row popover open with arrow left aligned
image

If it's `last` cell in the row it opens in this way:
image


If it's not `first` or `last` it opens centered
image

andresgutgon: @jquense please let me know if make sense what I said. Maybe there is an easier way of do what we want.
Thanks!

jquense: it makes sense, tho its a bit specific to your needs ;) maybe just pass the range and value in and you can figure out if it's the first or not in your app code? I don't like having props that have no purpose in the code itself, it'll be really easy to break your usecase later without realizing

andresgutgon: deal

andresgutgon:
@jquense done all changes 🎉

andresgutgon:
@jquense please let me know if I need to change something else. I would like to have this code merged. Thanks 🙏

jquense:
thanks!

andresgutgon:
Thanks @jquense 🙌
Will you do a release with these changes? It would be awesome :)

PR #403, opened 2017-05-21
example will not run without setting a localizer

furiousOyster:


jquense:
? The localizer is set elsewhere for the examples

PR #410, opened 2017-05-22
Calculate segment padding according to actually visible dates

bkniffler:
I'm trying to make month view that has a filtered range (imagine defining visibleWeekDays=[1, 2, 3, 4, 5] to render all days except saturdays/sundays, or to have a shouldDayRender function that returns true/false depending on the date). These filters apply to the range that is handed to `<DateContentRow />`. 

Unfortunately, this doesn't currently work since the event padding is calculated via the difference between range start and event start. With this PR, we would get the correct padding without assuming that all dates are shown.

This is how a use-case this PR allows could look like:
ohne titel

Its simply a custom month-view rendered 4 times, that squeezes all days in one row and filters out saturdays/sundays.

### DIFF
--- a/src/utils/eventLevels.js
+++ b/src/utils/eventLevels.js
@@ -13,7 +13,7 @@ export function eventSegments(event, first, last, { startAccessor, endAccessor }
   let start = dates.max(dates.startOf(get(event, startAccessor), 'day'), first);
   let end = dates.min(dates.ceil(get(event, endAccessor), 'day'), last)
 
-  let padding = range.findIndex(x => x.getTime() === start.getTime());
+  let padding = range.findIndex(x => dates.eq(x, start, 'day'));
   let span = dates.diff(start, end, 'day');
 
   span = Math.min(span, slots)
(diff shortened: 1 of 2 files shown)


src/utils/eventLevels.js
```diff
@@ -8,12 +8,12 @@ export function endOfRange(dateRange, unit = 'day') {
...
 }
 
-export function eventSegments(event, first, last, { startAccessor, endAccessor }) {
+export function eventSegments(event, first, last, { startAccessor, endAccessor }, range) {
   let slots = dates.diff(first, last, 'day')
   let start = dates.max(dates.startOf(get(event, startAccessor), 'day'), first);
   let end = dates.min(dates.ceil(get(event, endAccessor), 'day'), last)
 
-  let padding = dates.diff(first, start, 'day');
+  let padding = range.findIndex(x => x.getTime() === start.getTime());
```
jquense: I'm not sure if this will work in all cases...is it certain that these will be exactly the same time? Perhaps you `dates.eq(x, start, 'day')` just to be safe?

Issue #407, opened 2017-05-22
Anchor hrefs hard coded to # 


tom-daly:
the following lines of code should you javascript:void(0); instead of pound. I'm using this in SharePoint Online 'Modern Pages' and the '#' is triggering navigation and it's breaking the calendar. I manually changed it to javascript:void(0); and it works.

EventEndingRow.js line 132
Month.js line 378
Timegrid.js line 332

Paul-Colucci:
This should either be a button, or the href in the anchor should be "javascript:void(0);" having an anchor element with the href set to "#" can cause postbacks in certain environments.

PR #411, opened 2017-05-22
replaced href='#' with href='javascript:void(0);' to prevent breaking navigation

tom-daly:
in response to Anchor hrefs hard coded to issue #407

### DIFF
--- a/src/EventEndingRow.js
+++ b/src/EventEndingRow.js
@@ -93,7 +93,7 @@ class EventEndingRow extends React.Component {
     return count
       ? <a
           key={'sm_' + slot}
-          href="#"
+          href='javascript:void(0);'
           className={'rbc-show-more'}
           onClick={e => this.showMore(slot, e)}
         >
--- a/src/Month.js
+++ b/src/Month.js
@@ -217,7 +217,7 @@ class MonthView extends React.Component {
       >
         {drilldownView
           ? <a
-              href="#"
+              href='javascript:void(0);'
               onClick={e => this.handleHeadingClick(date, drilldownView, e)}
             >
               {label}
--- a/src/TimeGrid.js
+++ b/src/TimeGrid.js
@@ -311,7 +311,7 @@ export default class TimeGrid extends Component {
         >
           {drilldownView ? (
             <a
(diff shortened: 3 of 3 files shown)


jquense:
Thanks for the PR. I really don't want to merge this though it's super hacky...that's not a criticism of you or the PR just an evaluation of the solution here. Can you talk to me about why this is required a bit more. SharePoint shouldn't be responding to the clicks or navigating since the event prevents the default navigation behavior

tom-daly:
Thanks for the quick reply. I think that this problem might existing regardless of SharePoint or if you are using this inside another platform. We don't have control over certain elements that are in or on the page. I'm not sure how many other followers are using this inside another platform we you don't own the page or site. 

I can't speak on if it's hacky to use that, web is full of hacky things. I haven't looked into the 'why' there is an <a> tag with a link to '#'. Is the <a> itself really needed if you aren't using it to navigate? This issue however is a common thing that I've run across before.

https://stackoverflow.com/questions/20676873/why-use-javascriptvoid0-instead-of-in-href

The second highest voted comment here states the problem. Whatever it is that someone else developed is getting caught up in the #. If you want to stick with href that's fine but it should have an onclick event that returns false. 
https://stackoverflow.com/questions/134845/which-href-value-should-i-use-for-javascript-links-or-javascriptvoid0

joperron:
This should be done [the proper way](https://facebook.github.io/react/docs/handling-events.html).

eg: every onClick handler should have a e.preventDefault

jquense:
@joperron they already have `preventDefault` called

joperron:
I guess the bug is in @tom-daly's code — Something is preventing that from firing.

That being said, `<button>` is the proper element for this context anyway.

I'll try to submit a PR soon

tom-daly:
I jumped too soon. That worked for my case but I didn't have time to dig into why there was an a tag or if it could be replaced.

Paul-Colucci:
This should either be a button, or the href in the anchor should be "javascript:void(0);"  having an anchor element with the href set to "#" can cause postbacks in certain environments.

PR #413, opened 2017-05-24
export calenar class

chrisy-lili:
I wonder whether it's okay to export the Calendar class itself, in addition to exporting the component returned from uncontrollable. This would allow me to provide a custom render function by extending Calendar.


jquense:
Hi there, thanks for the PR. I don't really want to do that since its generally considerd an anti-pattern to extend anything other than `Component`. If you insist tho :P you can access the underlying component from the uncontrollable instance via `BigCalendar.ControlledComponent`

PR #417, opened 2017-05-24
Dynamic slots

TeaBough:
The purpose of dynamic slots is to fit events start/end date. I need this feature for a personal project, thus I thought why not push this PR in case someone might find it useful.

This is what dynamic slots look like : 

image

### DIFF
--- a/examples/App.js
+++ b/examples/App.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Component } from 'react';
 import Api from './Api';
 import Intro from './Intro.md';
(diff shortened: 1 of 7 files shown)


jquense:
Hi there, thanks for the PR, I have to admit, I don't understand what this does :P can you talk about the use case?

TeaBough:
The result i'm expecting is something like that : 

image

In the example the slots are based on events and not fixed (9h05 and not 9h). This allows me to implement for instant the week view of a student's semester. Here is another example : 

image

PR #373, opened 2017-05-02
Add Resource view

sabarasaba:
Heya,

I'm back from holidays, but have a lot of stuff on my plate at the moment and I don't know how much more time I'll be able to put on this pr. I'm hopeful someone else that is in need of it will be able to pick it up and make the few small things that are needed:

- [ ] The `Resource` and `ResourceGrid` are just duped files from `Day` and `TimeGrid` files with some extra added functionality. I'd be much happier if we could somehow DRY it up.
- [ ] Ideally when having lots of resources you should be able to have a minimum amount of resources that are shown and the rest would be hidden and shown upon scrolling horizontally (pretty much like an excel grid would work). Also the resource headers should always be visible, only cells would scroll.
- [ ] It needs an proper demo in the `/examples/demos` folder and documentation.

I've added a `story` that shows how the resource view works and example data in `stories/resourceEvents.js`.

Fixes #13 and others.

PR #423, opened 2017-06-02
Add resource

TeaBough:
I've created this PR, that serve the same purpose as https://github.com/intljusticemission/react-big-calendar/pull/373. The difference is that the logic of the resource is done directly in the `TimeGrid` instead of creating a `ResourceGrid` which DRY things up and allow to enjoy resource in the week view as well.

### DIFF
--- a/src/DayColumn.js
+++ b/src/DayColumn.js
@@ -68,6 +68,7 @@ class DayColumn extends React.Component {
     dayWrapperComponent: elementType,
     eventComponent: elementType,
     eventWrapperComponent: elementType.isRequired,
+    resource: PropTypes.string
   }
 
   static defaultProps = {
@@ -367,6 +368,7 @@ class DayColumn extends React.Component {
       slots,
       start: startDate,
       end: endDate,
+      resourceId: this.props.resource,
       action,
     })
   }
--- a/src/TimeGrid.js
+++ b/src/TimeGrid.js
@@ -26,6 +26,7 @@ export default class TimeGrid extends Component {
 
   static propTypes = {
     events: PropTypes.array.isRequired,
+    resources: PropTypes.array,
 
     step: PropTypes.number,
     range: PropTypes.arrayOf(
@@ -49,6 +50,7 @@ export default class TimeGrid extends Component {
     allDayAccessor: accessor.isRequired,
     startAccessor: accessor.isRequired,
     endAccessor: accessor.isRequired,
+    resourceIdAccessor: accessor.isRequired,
 
     selected: PropTypes.object,
     selectable: PropTypes.oneOf([true, false, 'ignoreEvents']),
@@ -146,6 +148,7 @@ export default class TimeGrid extends Component {
       , width
       , startAccessor
       , endAccessor
+      , resources
       , allDayAccessor
       , showMultiDayTimes} = this.props;
 
@@ -178,10 +181,12 @@ export default class TimeGrid extends Component {
 
     let gutterRef = ref => this._gutters[1] = ref && findDOMNode(ref);
 
+    let eventsRendered = this.renderEvents(range, rangeEvents, this.props.now, resources || [{}]);
+
     return (
       <div className='rbc-time-view'>
 
-        {this.renderHeader(range, allDayEvents, width)}
+        {this.renderHeader(range, allDayEvents, width, resources)}
 
         <div ref='content' className='rbc-time-content'>
           <div ref='timeIndicator' className='rbc-current-time-indicator' />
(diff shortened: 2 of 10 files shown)


jquense:
🎉 thanks for picking this up!

src/TimeGrid.js
```diff
@@ -174,10 +176,14 @@ export default class TimeGrid extends Component {
 
     let gutterRef = ref => this._gutters[1] = ref && findDOMNode(ref);
 
+    let eventsRendered = resources ?
```
jquense: `resources && resources.length` ?

src/TimeGrid.js
```diff
@@ -189,22 +195,56 @@ export default class TimeGrid extends Component {
...
+      let daysEvents = events.filter(
+          event => dates.inRange(date,
+              get(event, startAccessor),
+              get(event, endAccessor), 'day')
+      )
+
+      return resources.map((resource, id) => {
+
+        let eventsToDisplay = daysEvents.filter(
+            event => event.resourceId === resource.id
```
jquense: Like the `events` I don't want to be prescriptive about the shape of the data in `resources` so can we add `accessor` props for these fields like we do with startAccessor, et al?

src/TimeGrid.js
```diff
@@ -189,22 +195,56 @@ export default class TimeGrid extends Component {
...
           />
-
-          {this.renderEvents(range, rangeEvents, this.props.now)}
+          {eventsRendered}
 
         </div>
       </div>
     );
   }
+  renderEventsWithResource(range, events, today, resources) {
```
jquense: can this method be joined with `renderEvents` I'd be worried the common code will diverge unintentionally

XVincentX:
@TeaBough Is this still happening?

TeaBough:
I couldn't work on it during the french summer. But I think I can work on this this weekend !

XVincentX:
I think this feature would be a significant game changer here 👍

TeaBough:
@XVincentX @jquense I've finally updated this PR

jquense:
Awesome thanks! I'll try and review asap

jquense:
looks great! Can you rebase? The one concern i see here is that with a few resources the rendering hangs for a few seconds as you switch to the tab. I'm not sure what it is that is causing it, it may just be the number of divs being rendered as the grid since I see the same thing with or without any events.

TeaBough:
I've rebased

rvignesh89:
This is great! Exactly what I've been looking for. @jquense any chance that this will be added soon?

arthurmmedeiros:
Hey guys, 
When is this feature going to be added to the master?

ksloan:
Any updates?

samlll42-github:
Looking forward to this as well. Happy to help moving this along if it requires some testing or updates....

jquense:
If anyone wants to help, the last remaining thing is the performance concerns I noted in my comment. If anyone wants to take a crack at optimizing/testing that a bit

vongdarakia:
There's a slight performance issue, but it's a not a big deal for me. It's still way faster than other calendars out there. They often do an http request. I'd really like to see this feature out though. It'd be very helpful! Thanks for all the hard work @jquense and @TeaBough!

src/DayColumn.js
```diff
@@ -64,6 +64,7 @@ class DayColumn extends React.Component {
     dayWrapperComponent: elementType,
     eventComponent: elementType,
     eventWrapperComponent: elementType.isRequired,
+    resource: React.PropTypes.string
```
frehner: Need to remove `React.` here: just `PropTypes.string`

frehner:
Here's a screenshot looking at the Chrome Dev Tool's performance tab when the resource example is created in Storybook. 

screen shot 2017-11-28 at 11 10 50 am

I'm no expert, but it appears that there isn't anything _new_ per se that this PR introduces that reduces performance, it's just the fact that you're now rendering _N_ number of `DayColumn`s (one per resource).  I may be wrong, but if performance is a worry here then we'll either have to figure out a different way to do resources, or go into Day/Time Column components and optimize there.

Thoughts?

jquense:
yeah it may just be the extra columns now. The issue is that it's gonna worse and worse the more resources you have. One solution may be to do something completely different layout wise, like full calendar: https://fullcalendar.io/scheduler/ something a bit more amenable to simpler layout.

another option is to maybe try and virtualize the TimeColumns so it's not rendering extra divs. We can also jsut try and reduce the amount of DOM elements rendered in the first place

frehner:
I'm currently looking at a way of possibly just having one DayColumn component and just using styling to put the events into different columns... does that seem like a reasonable solution?

jquense:
I'm up for whatever works and isnt too hacky :P

frehner:
On second thought, changing the CSS works well when it comes to displaying the events, but would require more work when you consider creating events. Hmm. I'm not really sure what the best way to go about doing this is then, other than what has already been done here.

XVincentX:
Will this find its way in the master somehow? @TeaBough

TeaBough:
@XVincentX I've rebased with master and removed `React.` from `React.PropTypes`. I don't know what else I can do to make this PR ready to merge...

jquense:
if we can make CI green i'm ok merging for now. We really need to address the performance issues though

TeaBough:
@jquense done

marwej:
Any thoughts on when next version (including this) will be released? (Many thanks for a great package!)

f2net:
Could you please add the new "resources" API in the documentation?
I'm looking at the source to understand it, but it would be great to have it documented.

Thank you,
Matteo

atul-vzn:
Does anyone have a working example?

Thank you
Atul

justoverclockl:
> Could you please add the new "resources" API in the documentation? I'm looking at the source to understand it, but it would be great to have it documented.
> 
> Thank you, Matteo

any chance to know if this is implemented in 1.8.1? with a sort of documentation? thx!

PR #424, opened 2017-06-04
Permit to disable hover event title

TeaBough:


jquense:
Thanks again for the PR, I think i want to take a more holistic approach to this section, perhaps provide a more robust way to replace the entire component for those that are comfortable dealing with the positioning and style requirements

PR #435, opened 2017-06-14
Add allDayEventsLimit

ozhe:


jquense:
I think this is tough, becasue in the cases where events are hidden you'll require that user have the 'agenda' view. I'd rather the default behavior be to show a popup like in the Month view.

PR #446, opened 2017-06-23
Add support for "selectable" calendar on mobile

### LINKED ISSUE #38: Touch enabled?
Nice component, any plans to make this touch enabled?  It really should have this functionality given the mobility requirements of most apps these days.  If I end up uses this, I'll make a PR if you don't have any plans to implement this yourself.

Thanks again for the work here :)

minznerjosh:
This PR adds support for touch events when "selectable" is enabled. The
selection box appears after the user presses and holds on the screen
for 250ms as to not interfere with scrolling. This is similar to how the native iOS calendar behaves.

I ended up reorganizing/renaming some things in `Selectable.js` to make things make more sense in the context that it is not necessarily _mouse_ events that are being handled.

I'm looking forward to your feedback on this, and am happy to make any changes you suggest! Cheers!

Fixes #38.

### DIFF
--- a/src/Selection.js
+++ b/src/Selection.js
@@ -168,7 +168,7 @@ class Selection {
 
   _handleInitialEvent (e) {
     const { clientX, clientY, pageX, pageY } = getEventCoordinates(e);
-    var node = this.container()
+    let node = this.container()
       , collides, offsetData;
 
     // Right clicks
@@ -208,8 +208,6 @@ class Selection {
     if (result === false)
       return;
 
-    //e.preventDefault();
-
     switch (e.type) {
       case 'mousedown':
         this._onEndListener = addEventListener('mouseup', this._handleTerminatingEvent)
@@ -235,9 +233,9 @@ class Selection {
 
     if (!this._initialEventData) return;
 
-    var inRoot = !this.container || contains(this.container(), e.target);
-    var bounds = this._selectRect;
-    var click = this.isClick(pageX, pageY);
+    let inRoot = !this.container || contains(this.container(), e.target);
+    let bounds = this._selectRect;
+    let click = this.isClick(pageX, pageY);
 
     this._initialEventData = null
 
@@ -259,10 +257,10 @@ class Selection {
   }
 
   _handleMoveEvent(e) {
-    var { x, y } = this._initialEventData;
+    let { x, y } = this._initialEventData;
     const { pageX, pageY } = getEventCoordinates(e);
-    var w = Math.abs(x - pageX);
-    var h = Math.abs(y - pageY);
+    let w = Math.abs(x - pageX);
+    let h = Math.abs(y - pageY);
 
     let left = Math.min(pageX, x)
       , top = Math.min(pageY, y)
@@ -343,7 +341,7 @@ export function objectsCollide(nodeA, nodeB, tolerance = 0) {
 export function getBoundsForNode(node) {
   if (!node.getBoundingClientRect) return node;
 
-  var rect = node.getBoundingClientRect()
+  let rect = node.getBoundingClientRect()
     , left = rect.left + pageOffset('left')
     , top = rect.top + pageOffset('top');
 
--- a/src/Selection.js
+++ b/src/Selection.js
@@ -2,10 +2,10 @@ import contains from 'dom-helpers/query/contains';
 import closest from 'dom-helpers/query/closest';
 import events from 'dom-helpers/events';
 
-function addEventListener(type, handler) {
-  events.on(document, type, handler)
+function addEventListener(type, handler, target = document) {
+  events.on(target, type, handler)
   return {
-    remove(){ events.off(document, type, handler) }
+    remove(){ events.off(target, type, handler) }
   }
 }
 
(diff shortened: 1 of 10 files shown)


jquense:
awesome, thanks for the PR :) I'm a bit short on time in the near future but i'll try and review as soon as I can. THanks for taking the time to do this!

src/Selection.js
```diff
@@ -88,15 +107,74 @@ class Selection {
...
+  // without moving their finger for 250ms.
+  _addLongPressListener(handler, initialEvent) {
+    let timer = null;
+    let touchMoveListener = null;
+    let touchEndListener = null;
+    const handleTouchStart = (initialEvent) => {
+      timer = setTimeout(() => {
+        cleanup();
+        handler(initialEvent);
+      }, 250);
```
GameHex: Can the long press timer value be made configurable?

src/Selection.js
```diff
@@ -113,92 +191,109 @@ class Selection {
...
+      x: pageX,
+      y: pageY,
+      clientX,
+      clientY,
     });
 
     if (result === false)
       return;
 
     //e.preventDefault();
```
GameHex: I know this wasn't part of your PR, but this commented line looks fairly clean-up-able.

src/Selection.js
```diff
@@ -113,92 +191,109 @@ class Selection {
...
-
-    this.selecting = false;
   }
 
-  _openSelector(e) {
-    var { x, y } = this._mouseDownData;
-    var w = Math.abs(x - e.pageX);
-    var h = Math.abs(y - e.pageY);
+  _handleMoveEvent(e) {
+    var { x, y } = this._initialEventData;
```
GameHex: I see a lot of `var` and `let`. Just being picky, but the vars could be turned into lets since I also see consts in the code base.

jquense:
@GameHex I don't think it has, please feel free to jump in and run some tests. I'd love to merge this.

minznerjosh:
Thanks for the review! I'll address your feedback on Monday. It's only been tested on an iPad. That's the only device I have available to me.

GameHex:
It works on Browserstack, however it can be enhanced to accommodate drag and drop as well by swapping the _react-dnd-html5-backend_ in **src\addons\dragAndDrop\index.js** with [react-dnd-touch-backend](https://github.com/yahoo/react-dnd-touch-backend). It doesn't have desktop capture events enabled out of the box, but per their docs you can enable them by using it like this: 
`return DragDropContext(TouchBackend({ enableMouseEvents: true }))(DragAndDropCalendar);` 
in the dnd's index.js file near the bottom and I have verified it working in both desktop and mobile.

aert:
Experimenting the same issue on iOS Safari select events are not fired..., flawlessly working on android (chrome/firefox) otherwise.

Can we hope for a merging of this PR ?

minznerjosh:
Sorry y'all. I've been swamped lately. I'll make a note to fix-up some of the issues here tomorrow morning so this can hopefully get merged soon.

aert:
Thanks for your work @minznerjosh !

minznerjosh:
Sorry for the delay everybody! This PR has been updated to

* Make the long press threshold configurable
* Replace `var`s with `let`s in the `Selection` module
* Remove some commented-out code

I have _not_ replaced the `react-dnd-html5-backend` with yahoo's `react-dnd-touch-backend`. I'd love some guidance on what to do there.

Simply changing dependencies would be a breaking change, as we require users to install the backend themselves (it's an optional dependency.) Alternatively, I could write some waterfall logic to first try to use `react-dnd-touch-backend` if it is available, and otherwise fallback to `react-dnd-html5-backend`.

But I'm not sure the added complexity is necessary seeing as we allow users to pass in their own backend anyway. The obvious tradeoff is requiring users to do some configuration to get mobile drag support, instead of offering it out-of-the-box.

I'd love to get some direction on this. Thanks!

aert:
Hi guys, I tested this on chrome/firefox latest on Android, unfortunately `onSelectSlot` event are not fired anymore,  
Here is my config:

```
        <BigCalendar
          selectable="ignoreEvents"
          events={events}
          startAccessor="start"
          endAccessor="end"
          defaultView="day"
          components={{ toolbar: CalGoLiveCalendarDisplayToolbar }}
          views={["month", "week", "day"]}
          onSelecting={this._onSelecting}
          onSelectEvent={this._onSelectEvent}
          onSelectSlot={this._onSelectSlot}
          showMultiDayTimes
        />
```

**Additional details:**

`onSelectEvent` are working fine on Android.
`onSelectSlot` are fired *only* in month view when `longPressThreshold` is reached... in day/week view no event is fired.

Thanks for the follow up :)

minznerjosh:
@aert are you able to reproduce this in the `react-big-calendar` example app? I can't.

aert:
@minznerjosh I'll test it tomorrow on the example app then feed back, thanks !

jeromevvb:
I just tried with browserStack (Ipad Pro Safari IOS 11) on this example http://intljusticemission.github.io/react-big-calendar/examples/index.html#selectable
and any event are fired, even after a long press :(

Issue #448, opened 2017-06-26
rbc-event-continues-after class calculation incorrect


peterox:
The EventCell component is supposed to add the rbc-event-continues-after class to the element if the end date of the event is after the current event row.

The problem is that for an EventRow that is showing for example the dates Monday the 5th June to Sunday the 11th of June, the 'end' prop is June 12th at midnight.

The calculation for continuesAfter is
```javascript
continuesAfter = dates.gt(end, slotEnd, 'day')
```

So if the event goes from Saturday June 10th to Monday June 12th then the above will be false as the end date day is NOT greater than the slotEnd dat day. i.e. 12 > 12.

I believe this should be !dates.lt(end, slotEnd, 'day')

jquense:
thanks for hte issue, care to make a PR? I think you can also use `datas.gte` here as well

sunlee-newyork:
Calculation for `rbc-event-continues-prior` is incorrect as well - I'm seeing this being applied to an event segment that is the first segment.

PR #450, opened 2017-06-26
Correct calculation of .rbc-event-continues-after class

peterox:
Fix for issue #448


nosnaj:
Surely this PR is straightforward and safe to merge? I have already personally tested this out and it fixes my issue. However, the same cannot be said for the events that get rendered in the popup.

Correctly rendering in day view:
image

Wrong in the popup:
image

The red and blue right/left borders are just custom visual cues for the start/end of an event.

PR #451, opened 2017-06-26
Fix for marking date as 'now'

peterox:
Default 'now' prop in Month view so that the .rbc-now class is correctly added to the current date when the 'now' prop isn't supplied to the Calendar.


jquense:
thanks!

Issue #449, opened 2017-06-26
agendaHeaderFormat date (start, end) return undefined


AgamlaRage:
react-big-calendar 0.14.4
localizer: moment 2.18.1

I cannot get the **start** or **end** date from the ```agendaHeaderFormat(date, culture, localizer)``` format.
This is returning undefined for both.
`culture` and `localizer` params are fine.

Others formats like dayRangeHeaderFormat, dayHeaderFormat return date normally.

Thx

peterox:
I believe this was broken by @jquense in  f084becac0fa83ff56062094e318e10cf02dabcb.

The change to viewLabel.js expects the View.range() methods to return an array. Day, Week and WorkWeek were changed to return an array but Month and Agenda were not. As Agenda's header is a range then the getRangeBounds is failing to return the correct range object to be localized.

PR #457, opened 2017-07-05
Issue #449 - Fixed agenda header label

peterdivvito:
- Added missing less plugin
- Fixed examples not rendering


src/utils/viewLabel.js
```diff
@@ -13,9 +13,12 @@ const Formats = {
 }
 
 function getRangeBounds(range) {
-  let start = range[0]
-  let end = range[range.length - 1]
-  return { start, end }
+  if (range instanceof Array) {
```
jquense: Can you use Array.isArray instead?

jquense:
thanks!

PR #472, opened 2017-07-14
Event prop getter updates

choruk:
- adds enzyme (which needs jsdom), mocha, chai, sinon testing frameworks
- makes eventPropGetter work for the Agenda view
- add view as another argument passed to the eventPropGetter function


choruk:
Closing this PR in favor of opening up smaller PRs to address each separate issue individually (the first of which can be seen in #540).

PR #478, opened 2017-07-20
Remove unneeded href

mgoudy91:
This should fix some routing issues

### DIFF
--- a/src/EventEndingRow.js
+++ b/src/EventEndingRow.js
@@ -93,7 +93,6 @@ class EventEndingRow extends React.Component {
     return count
       ? <a
           key={'sm_' + slot}
-          href="#"
           className={'rbc-show-more'}
           onClick={e => this.showMore(slot, e)}
         >


jquense:
href-less anchors behave differently then is desired here. I'd check the previous issues on this subject for full context, but ultimately his should be a button not a link

mgoudy91:
Ah, I missed those issues on my first look through. I'll take a look

ravihugo:
Just ran into this issue.
This was my local fix in my component that was rendering my BigCalendar:

    componentDidUpdate() {
        // After an update, we must dynamically replace all +1 more links a bit
        var calendar = ReactDOM.findDOMNode(this);
        if (calendar instanceof Element) {
            calendar.querySelectorAll('a').forEach(a => {
                a.href = 'javascript:false';
            });
        }
    }

Effectively using javascript:false instead of hash-symbol works perfectly for our use-case.

PR #496, opened 2017-08-07
Option to show multi-day events with specific start and end times in time-slots

Auchindoun:
Added the config option `showMultiDayTimes` to show multi-day events with defined start and end times to show in the time slot views as opposed to the all-day header.

However there were limitations in doing this as all day events and time range events are rendered as separate objects, so when there is an 'in-between' day in the multi-day event, this will take up the entire time slot view for the given days. I found this to be more desirable, because if it were that way, then clicking on the all day object would not link to the rest of the days in the event.

The bulk of this work is from PR #42 but I have updated it with the new styling changes from the past year, as well as made this feature a configurable option.

### DIFF
--- a/examples/demos/basic.js
+++ b/examples/demos/basic.js
@@ -12,6 +12,7 @@ let Basic = React.createClass({
         events={events}
         views={allViews}
         defaultDate={new Date(2015, 3, 1)}
+        showMultiDayTimes={true}
       />
     )
   }
--- a/examples/demos/basic.js
+++ b/examples/demos/basic.js
@@ -12,7 +12,6 @@ let Basic = React.createClass({
         events={events}
         views={allViews}
         defaultDate={new Date(2015, 3, 1)}
-        showMultiDayTimes={true}
       />
     )
   }
--- a/examples/events.js
+++ b/examples/events.js
@@ -66,5 +66,15 @@ export default [
     'title': 'Birthday Party',
     'start':new Date(2015, 3, 13, 7, 0, 0),
     'end': new Date(2015, 3, 13, 10, 30, 0)
+  },
+  {
+    'title': 'Late Night Event',
+    'start':new Date(2015, 3, 17, 19, 30, 0),
+    'end': new Date(2015, 3, 18, 2, 0, 0)
+  },
+  {
+    'title': 'Multi-day Event',
+    'start':new Date(2015, 3, 20, 19, 30, 0),
+    'end': new Date(2015, 3, 22, 2, 0, 0)
   }
 ]
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -12,6 +12,7 @@ import {
 import { notify } from './utils/helpers';
 import { navigate, views } from './utils/constants';
 import defaultFormats from './formats';
+import message from './utils/messages'
 import viewLabel from './utils/viewLabel';
 import moveDate from './utils/move';
 import VIEWS from './Views';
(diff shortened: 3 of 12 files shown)


mtt87:
@jquense @Auchindoun 
I've tried this PR and so far it works really well except for one big issue:
it has problems whenever it's the end of the month.

screen shot 2017-08-29 at 21 54 52
screen shot 2017-08-29 at 21 54 58


I suspect you are adding yourself `days = days +1` instead of using something more reliable like `moment().add(1, 'days')`

I'm going through the code now and see if I can spot the problem and fix it.

Other than this, I haven't found any other issue so far it looks great!
Thanks @Auchindoun for this contribution!

Auchindoun:
I'll get that sorted out @mtt87, thanks for the heads up. I was adding to it like you said, as I want to avoid using moment.js for the core rendering, since not everyone will be using moment, even though they should ;)

Auchindoun:
@mtt87 [I'm using Date.getDate() to get the start, end, and current date of the events to compare, so it's comparing 28/30/31 against 1, and thus failing](https://github.com/intljusticemission/react-big-calendar/pull/496/files#diff-1606d8d68f8e0307027a29cbf08fca91R133). I'll have to work around that once I find some time

Auchindoun:
Fixed @mtt87

mtt87:
Great job I'll give it a try tomorrow probably!

Thanks for your contribution

Auchindoun:
@mtt87 Haha no problem. I'll pull from my fork for my project in the meantime. Thanks!

mtt87:
Works like a charm!! Great job @Auchindoun , now hopefully it will be included in the next version 😄

tobiasandersen:
I just pulled latest from master, and this PR seems to break both the day and week view, cc @jquense @Auchindoun. 

Here's a screenshot of the day view before this PR:

screen shot 2017-09-12 at 22 
45 22

And one after with the same events:

screen shot 2017-09-12 at 22 46 40

It doesn't seem to matter what events are passed, they all show up down in the bottom.

Auchindoun:
Interesting. I’ll check that out asap

On Sep 12, 2017, at 3:50 PM, Tobias Andersen <notifications@github.com<mailto:notifications@github.com>> wrote:


I just pulled latest from master, and this PR seems to break both the day and week view, cc @jquense<https://github.com/jquense> @Auchindoun<https://github.com/auchindoun>.

Here's a screenshot of the day view before this PR:

[screen shot 2017-09-12 at 22   45 22]<https://user-images.githubusercontent.com/3599069/30347372-23f97454-980c-11e7-8f2d-f34b4a664145.png>

And one after with the same events:

[screen shot 2017-09-12 at 22 46 40]<https://user-images.githubusercontent.com/3599069/30347421-47ccf0e0-980c-11e7-9386-81a201927b01.png>

It doesn't seem to matter what events are passed, they all show up down in the bottom.

tobiasandersen:
It was just a simple mistake, I've made a PR with a fix: https://github.com/intljusticemission/react-big-calendar/pull/546.

jasonogasian:
I noticed that if there are concurrent events that wrap to at least midnight they are each 100% width, making it impossible to see one if they are the same length.

jasonogasian:
Just a follow-up on my previous comment it looks like the issue is in this line in dayViewLayout.js
return (Math.abs(start1 - start2) < 60 && start2 < end1)

It doesn't know end1 is on a different day and therefore actually _after_ start2.

jquense:
@jasonogasian care to send a  PR?

PR #498, opened 2017-08-11
Fixed scrolling issues

sceptre12:
Resolved scrolling issue. 
There probably needs to be more clean up done but this is a start.
https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView

### DIFF
--- a/src/TimeGrid.js
+++ b/src/TimeGrid.js
@@ -110,6 +110,10 @@ export default class TimeGrid extends Component {
 
     this.applyScroll();
     this.positionTimeIndicator();
+    var rbc = document.getElementsByClassName('rbc-event');
+    if(rbc.length > 0){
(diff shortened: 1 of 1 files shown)


jquense:
Have there, thanks for the PR. Which scrolling issue are you trying to resolve?

sceptre12:
Hello,

So I have a parent component “A” and  child component "B" that contained the Big Calendar.
When B is already mounted and I’m passing updated information from A to B, the scrollToTime function was never updating or working. 
It always works on initial mounting of B which also mounts the Big Calendar, but after the mounting no new updates would cause the scrollToTime function to work.

jquense:
thanks for the PR, this is probably not the best way to handle this, scroll things are fraught and this will likely call scrollIntoView more times then required

Issue #182, opened 2016-09-30
Change the background color of off range date


dncdrn:
Hi! How will I change the background color of off range date in calendar?
For example this one, I want to change the background color of July 31, Sept 1 - 4.
Thanks in advance!
![offrange](https://cloud.githubusercontent.com/assets/21171473/18981592/5a6c6f5c-8712-11e6-9369-df5f97153ad9.PNG)

jquense:
There isn't any hook for changing or styling individual background cells, but I'd accept a PR adding something similar to eventPropGetter

PR #499, opened 2017-08-11
Add `off-range-bg` in MonthView day component

v1k45:
- Add `rbc-off-range-bg` className to days which are off range
  from current month displayed in MonthView.
- Update off range date and bg colors

-----

This is similar to `off-range` css applied to date values which are not in the current month. This is also a potential fix for #182. 

The update example will look like this:

screenshot from 2017-08-11 19-58-24


jquense:
thanks!

Issue #203, opened 2016-10-27
disable "All day"  row


KnockNow:
Hi,
thank you for this library !
i want disable the display of the line "all day", how can i do pls ? 
And i have a problem with the view. I can't disable two view (day and aganda)...

mawek:
you can pass ```views``` property to the ```BigCalendar``` component
```
<BigCalendar
                        events={events}
                        views={['week', 'day']}
                        defaultView={'week'}
...
```

KnockNow:
Thank you but i want disable the row "All day", is that possible ?

jquense:
its not currently possible to disable the all day view, in part because there would be no way of showing events that span multiple days (even if they aren't marked all day), as @mawek points out you can adjust the vies you want with the `views` prop

mcmatan:
Is there a way of currently removing this line? Maybe with css?

KnockNow:
Yes, I have sucessfully done it but not with css (it's not possible because that line have a dependency). You can fork this lib and just comment the code that does this. (And why not create a parameter for enable/disable that line). Good luck !

mcmatan:
Yeah that's what I had in mind, thanks! (:

mcmatan:
Did it with CSS! Boom!
```

.rbc-time-view .rbc-row {
    min-height: 0px;
}

.rbc-time-view .rbc-row div:nth-child(1) {
    height: 0px !important;
}
```

Maxincredible52:
@mcmatan you the real mvp

tkporter:
For anyone else looking to use @mcmatan's solution, you can use that and set the `allDay` message to `''`: http://intljusticemission.github.io/react-big-calendar/examples/index.html#prop-messages

darrenklein:
I went with

    .rbc-row:nth-of-type(2) {
	    display: none !important;
    }

Anything wrong with that?

darrenklein:
Using version `0.16.1`, I found this worked:

    .rbc-time-header > .rbc-row:nth-of-type(2) {
	    display: none !important;
    }

This got rid of the 'all day' row in week and day views. Anything wrong with this fix?

drewandrew:
I used
```css
.rbc-allday-cell {
  display: none;
}
.rbc-time-view .rbc-header {
  border-bottom: none;
}
```
anything wrong with this (as far as CSS hacks go)?

[...]

PR #500, opened 2017-08-11
Add boolean allDaySlot prop to Calendar to match FullCalendar

kat3kasper:
Determines if the "all-day" slot is displayed at the top of the calendar.
When hidden with false, all-day events will not be displayed in agenda views.

Boolean, default: true

### DIFF
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -227,6 +227,12 @@ class Calendar extends React.Component {
    rtl: PropTypes.bool,
 
    /**
+    * Determines if the "all-day" slot is displayed at the top of the calendar.
+    * When hidden with false, all-day events will not be displayed in agenda views.
+    */
+   allDaySlot: PropTypes.bool,
+
+   /**
     * Optionally provide a function that returns an object of className or style props
     * to be applied to the the event node.
     *
--- a/src/TimeGrid.js
+++ b/src/TimeGrid.js
@@ -41,6 +41,7 @@ export default class TimeGrid extends Component {
     culture: PropTypes.string,
 
     rtl: PropTypes.bool,
+    allDaySlot: PropTypes.bool,
     width: PropTypes.number,
 
     titleAccessor: accessor.isRequired,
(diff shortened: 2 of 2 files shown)


jquense:
Hi there! thanks for the PR. There are quite a few issues about this particular feature already. We don't allowing hiding the AllDay slots because it hide events that are all-day or span a day boundary. Overall I try to avoid adding options that can lead to broken configuration for consumers.

kat3kasper:
@jquense thanks for the quick response. Anyway we can add a warning or something. Its great to have the react version available but this functionality is available in FullCalendar. Would prefer not to have to publish a new npm package that I will have to update everytime things change here. Any ideas how to work around that?

arecvlohe:
I imagine forking this repo and adding any additional functionality is an option

jquense:
I don't think forking is necessary. There is an issue or two in the tracker that details a css style for doing it. I can be open to the toggle in the api but the docs need to be clear the cost and potential downsides

Auchindoun:
We might be able to pair this with my PR, so that they can disable the all day column, and any days that DO span all day will take up the whole day slot. I could make the adjustment. And of course have big warning that this can cause annoyances for busy schedules. Thoughts @jquense @kat3kasper?

jquense:
Works for me

Auchindoun:
@kat3kasper do you want me to give you access to my fork, or do you want me to add your changes in to my PR?

francomendoza:
@Auchindoun @kat3kasper is there an update on this?

tobiasandersen:
This is quite old now, so I'll close it for now. If anyone wants to add the changes requested by @jquense, feel free to re-open again. Thanks everyone and especially @kat3kasper for opening the PR! 

For anyone looking for a solution, please read through https://github.com/intljusticemission/react-big-calendar/issues/203.

marwej:
This would be very useful for me too. If I can assure, in my application, that no event will ever span several days, then I don't want to show the "All day" row. With the new resource row (great feature!), suggested CSS won't work since the "All day" row can be both second row (without resources) and third row (with resources). (I try to avoid CSS and use inline styles as much as possible too.)

rizvified:
I think this is super useful. As my project implementation does not require an all day view and using CSS or inline styles to hide it is not a good solution.

SrishtiSingh1902:
when I am writing allday:false it is giving error is the start and end accessor expecting particular datetime format , Can anybody suggest ?

PR #505, opened 2017-08-16
Fix prop types validation when passing object to views

ox-michaelradionov:
When passing object to `views` property, there is an error in browser: 

```
Warning: Failed prop type: validator.apply is not a function
```

The reason is incorrect usage of [react-prop-types#all](https://github.com/react-bootstrap/react-prop-types#allvalidators), which spreads args, but does not handle arrays.

Component itself continues to work fine though.


jquense:
Thanks!

Issue #488, opened 2017-07-31
defaultView missing from api docs


olivergeorge:
I think this is a small oversite from the docs/demo page.  `defaultView` is not listed as a prop and the only mention is associated with the `view` property.

http://intljusticemission.github.io/react-big-calendar/examples/index.html

jquense:
yeah, the `default` props aren't listed seperately as they are directly tied to their related controlled/uncontrolled prop. I'm open to a PR adding them tho if anyone wants to tackle it

PR #510, opened 2017-08-17
Add defaultView prop to docs

arecvlohe:
This relates to #488. This adds the defaultView prop to the docs.


arecvlohe:
Let me know if there is anything I need to change or update 👍

PR #516, opened 2017-08-18
Fix a Potentially unhandled rejection variable

DanZeuss:
Potentially unhandled rejection [2] variable @cell-border is undefined in file /var/www/resources/assets/less/react-big-calendar/agenda.less line no. 10


jquense:
thanks!

PR #517, opened 2017-08-22
Allow specify custom argument for dnd instead of event

bkniffler:
This is useful in case event is a class instance (dnd plugin needs a plain object), so it can be converted to json before drag; or in case you want to modify the object for some reason.


bkniffler:
Any reasons to close this? This can be very helpful if events are not simple POJO but some class instance objects with non-serializable methods.

jquense:
we moved off of react-dnd. If there is an equivalent hook in the new dnd hook i'd be happy to merge something in

Issue #461, opened 2017-07-07
Overnight event doesn't render on second day


jasonogasian:
I have a use case where events can span midnight. When these events are drawn on the calendar only the portion on the first day is shown.

jquense:
Do you have a specific example? Please provide a reproduction thanks

jasonogasian:
When adding the following event to examples/events.js I would not expect to see an "all day" event as the event does not go for an entire day.

  {
    'title': 'TEST',
    'start': new Date(2015, 3, 3, 11, 0, 0),
    'end': new Date(2015, 3, 4, 0, 10, 0)
  },

PR #518, opened 2017-08-23
Render overnight events in time grid view

martynasj:
This should make it possible to render overnight events that span multiple days. The event which spans multiple days is added to the `rangeEvents`, so it appears in both "all day events" and in the "time grid" views. I haven't yet added any props to configure this behaviour, as I was not sure if this solution is good. 

References #461

### DIFF
--- a/examples/events.js
+++ b/examples/events.js
@@ -66,5 +66,15 @@ export default [
     'title': 'Birthday Party',
     'start':new Date(2015, 3, 13, 7, 0, 0),
     'end': new Date(2015, 3, 13, 10, 30, 0)
+  },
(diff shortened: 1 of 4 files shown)


jasonogasian:
@martynasj where is the "get" method defined that you're using in startsBeforeDay and endsNextDay?

mtt87:
I think https://github.com/intljusticemission/react-big-calendar/pull/496 is doing this and it's well written (including CSS to fix rounded borders)

There is just one problem when it's the last day of the month but haven't figured out where is the bug, if someone wanna help 😄

PR #523, opened 2017-08-28
Prevent popup growing up outside of calendar

gzc:
Prevent popup growing up outside of calendar. Before this change, the logic is when overlay reaches bottom it will grow up.

If there are too many events in a day, when the overlay reaches the bottom it can grow up outside calendar since the top can be a very big negative number. In this case I can't scroll to view hidden events.

bad top
bad view

After this change, the logic is when overlay reaches bottom it will grow up until it reaches the top of calendar, then it grows downward. So I can scroll down to view other events.

view2
view1

### DIFF
--- a/src/Popup.js
+++ b/src/Popup.js
@@ -48,7 +48,7 @@ class Popup extends React.Component {
       , leftOffset = (this.state || {}).leftOffset || 0;
 
     let style = {
-      top: top - topOffset,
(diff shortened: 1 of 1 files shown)


jquense:
thanks!

gzc:
@jquense Thanks for the quick response^_^

BTW, when will you release next version?

gachavez:
this would be very useful!

Issue #524, opened 2017-08-29
styles included in eventPropGetter can be overwrite at template definition 


csolarz:
Hello!
I was trying to integrate bootstrap to the event wrapper, by eventPropGetter where i added classnames and some styles, the problem was in the line [170](https://github.com/intljusticemission/react-big-calendar/blob/1ea675897aeb58268096347f7bd30fb2bbbdae95/src/DayColumn.js#L170) where overwrite width of any class of bootstrap like col-xs-*,  col-md-*, etc  

i think any possible solution is to move highlithed line [166](https://github.com/intljusticemission/react-big-calendar/blob/1ea675897aeb58268096347f7bd30fb2bbbdae95/src/DayColumn.js#L166)

![image](https://user-images.githubusercontent.com/3832225/29803630-bc40d910-8c53-11e7-8718-ff1199fc6c2b.png)

at final style props, like img:

![image](https://user-images.githubusercontent.com/3832225/29803676-f459677c-8c53-11e7-8be7-8ac1f0aabca7.png)

that's was a nice solution to adding bootstrap classes to event wrapper.  

tomorrow can be upload  a pullrequest if yours want!

saludos!!!

yury-stepanov:
Straggling with the same issue. Any chances it will be merged into master?

tobiasandersen:
Overriding those styles will mess with the positioning of events. Is that what you want?

yury-stepanov:
@tobiasandersen I am interesting in changing `width` and `left` \ `right` properties. This will help me to create more custom UI. (Overlapping events logic right now isn't the best choice for my scenario). It would leave changing of the `top` to the calendar and wouldn't mess with event position of the event in the day

tobiasandersen:
I see. While that definitely is a valid use case, I'm a bit worried people would override these styles by mistake and in turn break the layout. Another solution that's been discussed briefly is to make the layout algorithm pluggable. At the end of the day, I guess it's a decision for @jquense to make.

yury-stepanov:
Created a pull request #600 
@tobiasandersen @jquense would mind to take a look?

jquense:
Yeah, right now I think allowing this would be a footgun that I don't want to deal with support for. IN 99% of cases overriding these properties would be a bug. If we need more control over the layout there should be a proper API for it :)

skvark:
[...]

PR #526, opened 2017-08-30
#524 overwrite default props at EventWrapper with styles defined in eventPropGetter

csolarz:


jquense:
Hi there, thanks for the PR, My concern here is that now a user can break the calendar layout, by setting a style. The reason the style was spread before the overrides was so the user _couldn't_ change style settings that were structural.

csolarz:
thanks for the feedback, maybe i need another workaround for my specific issue

jquense:
Gonna close for now, but let me know if you can't figure out an additional approach here.

PR #529, opened 2017-08-30
Update Calendar.js - Add undocumented messages for localization

hudovisk:
Add undocumented messages keys for localization.


jquense:
many thanks!

Issue #316, opened 2017-03-06
changing timeGutterFormat -> props validation errors for TimeColumn and TimeSlotGroup


mtt87:
My calendar seems to work correctly but I get 2 warnings in the console.

This is my component:
```js
        <BigCalendar
          events={this.state.events}
          defaultView='week'
          views={['week', 'day']}
          step={30}
          timeslots={1}
          min={moment().minute(0).hour(7).toDate()}
          components={components}
          scrollToTime={new Date()}
          formats={{
            timeGutterFormat: (date, culture, localizer) => 
              localizer.format(date, 'H:mm', culture),
          }}
        />
```

2 errors:  

Warning: Failed prop type: Invalid prop `timeGutterFormat` of type `function` supplied to `TimeColumn`, expected `string`.  

Warning: Failed prop type: Invalid prop `timeGutterFormat` of type `function` supplied to `TimeSlotGroup`, expected `string`.

jquense:
definately supported, just a bad propType declaration. PR? :)

rrcobb:
Saw this same error, opened #530

PR #530, opened 2017-08-31
Fix timeGutterFormat propTypes

rrcobb:
#316

Uses the dateFormat propType from `utils/propTypes`, which matches the PropType for the timeGutterFormat in `Calendar.js`

Thanks for react-big-calendar!


jquense:
Thanks!

PR #540, opened 2017-09-08
Use event prop getter in agenda view

choruk:
Currently, the eventPropGetter is not used at all for the agenda view, which means the event nodes in the agenda view cannot be customized through this prop. This PR fixes that and also adds semi-colons to all the lines that are missing them in the agenda view.


choruk:
Thanks for the quick approval! I don't mean to bug you, but just curious why you didn't also merge this PR. Was that intentional or just an oversight? 😁

PR #406, opened 2017-05-21
Resizable Events

jnmandal:
This PR adds the ability to resize events to the DND plugin, as discussed in issue https://github.com/intljusticemission/react-big-calendar/issues/382. It includes:

bug fix to core:
* `eventComponent` was being destructured incorrectly in the `DayColumn` component's `renderEvents` function (so it was always undefined)

enhancements to dragAndDrop addon:
* passing prop `resizable` will have the wrapper use a new component `ResizableEvent` as the `eventComponent` for the calendar
* the `onEventResize` callback will give the ability to hook into the DND wrapping. It receives:
  * `resizeType`: 'drag' or 'drop' based on what type of interaction is causing the resize
  * `eventData`: object containing the original event as well as a new `end` for the event

gotchas:

Since only the original event object is accessible through react-dnd... In the `onEventResize` callback you will not be able to use Array.prototype.indexOf for finding the event again -- unless you continue to mutate the original event and splice it into the cloned events array... but I'm not sure if we would want to recommend that practice

PR #541, opened 2017-09-09
Feature/resizable

stevenkaspar:
Adds a `resizable` prop that appends a `=` to the bottom of the DayColumn event wrapper that allows for drag-resizing that updates events similar to DnD

```
<BigCalendar
          resizable
          onResizing={this.moveEvent} // receives: originalEvent, startDate, endDate
          onResizeInit={() => true}
```


jquense:
Hey there. Thanks a lot for taking the time to do this, contributions are always appreciated. There is actually already an effort underway to add this feature in #406 and a bunch of conversation around how we'd like to implement it. I would try and coordinate with the author there.

Issue #467, opened 2017-07-11
Broken in IE11 due to missing Object.assign


forum-is:
Commit 464af403aec464d2428220288960f6b16e01c34a by @jquense completely breaks IE11 due to removal of babel-plugin-object-assign.

Please re-add the polyfill for Object.assign and also consider fixing #464, which is another show-stopper for IE11. Thanks!

Paul-Colucci:
this would be nice!

tobiasandersen:
This should be fixed by https://github.com/intljusticemission/react-big-calendar/pull/542.

Issue #464, opened 2017-07-10
src/utils/eventLevels.js : findIndex is not supported on ie11


o-bo:
Since this [commit](https://github.com/intljusticemission/react-big-calendar/commit/2a70c3c3134b26e60997a729cc8048a8be91ee83), an error is thrown by ie11, complaining about findIndex [support](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex#Browser_compatibility).

We solved the problem by adding the [polyfill](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex#Polyfill) when initializing our project.

o-bo:
We corrected the problem by adding the [polyfill](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex#Polyfill) when initializing our project.

tobiasandersen:
This should be fixed since https://github.com/intljusticemission/react-big-calendar/pull/671 by including lodash's `findIndex` method.

PR #542, opened 2017-09-11
Fixing #464 and #467 for IE11 Compatibility

MatSeifert:
Please consider adding these two fixes for making React Big Calendar work again in Internet Explorer 11, since both of these functions are unsupported and need Polyfills.

### DIFF
--- a/src/utils/eventLevels.js
+++ b/src/utils/eventLevels.js
@@ -27,7 +27,6 @@ export function eventSegments(event, first, last, { startAccessor, endAccessor }
   }
 }
 
-
 export function segStyle(span, slots){
   let per = (span / slots) * 100 + '%';
   return { WebkitFlexBasis: per, flexBasis: per, maxWidth: per } // IE10/11 need max-width. flex-basis doesn't respect box-sizing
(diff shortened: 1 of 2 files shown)


src/utils/eventLevels.js
```diff
@@ -95,3 +94,33 @@ export function sortEvents(evtA, evtB, { startAccessor, endAccessor, allDayAcces
     || !!get(evtB, allDayAccessor) - !!get(evtA, allDayAccessor) // then allDay single day events
     || +get(evtA, startAccessor) - +get(evtB, startAccessor)     // then sort by start time
 }
+
+// Polyfill for IE11 
+if (!Array.prototype.findIndex) {
```
jquense: Instead of adding the polyfill let's make a note in the docs for users to add it instead as part of the setup process. It's not great practice to alter global objects in libraries

MatSeifert: I really understand this point, but on the other hand I think it is not a good idea to exclude a whole browser because of this. I know the IE is deprecated with the release of Edge, but at the moment there are plenty of users out there using this hell of a software, because they don't want, can't or don't know how to switch to FF, Edge or Chrome. I think a lot of devs will miss the hint to use polyfills, since they don't notice they are necessary, because they don't test the IE. In the End it is up to you how to handle this, I just wanted to point out the issue we ran into, and why. Maybe there is another compromise to fix the issues mentioned in my request.

jquense: I'm suggesting we exclude a browser and we should definately support Ie. I'm saying we should take the same approach as React which is specify what polyfills are required to use the component. I realize this may mean that folks may miss it initially but its better than being a bad citizen in the JS community by changing the global state, which can break other stuff in unexpected ways.

If you'd prefer we can just switch instances of the native findIndex to the lodash one, which is already included 👍 ?

MatSeifert: That would be ok for me, I guess :-)

anissaSanadtech:
so what are the required polyfills for the component ?

PR #546, opened 2017-09-12
fix broken day view layout

tobiasandersen:


alpfactory-alex:
Will this be merged soon? We're working on master and without this the day view obviously is broken ;)

PR #548, opened 2017-09-16
Add onSelectEvent in Agenda view

Bombe93:
Now in Agenda view there is event click like Month view.

### DIFF
--- a/src/Agenda.js
+++ b/src/Agenda.js
@@ -12,6 +12,7 @@ import { accessor as get } from './utils/accessors';
 import { accessor, dateFormat, dateRangeFormat } from './utils/propTypes';
 import { inRange } from './utils/eventLevels';
 import { isSelected } from './utils/selection';
+import { notify } from './utils/helpers';
 
 
 class Agenda extends React.Component {
@@ -25,6 +26,8 @@ class Agenda extends React.Component {
     endAccessor: accessor.isRequired,
     eventPropGetter: PropTypes.func,
     selected: PropTypes.object,
+
+    onSelectEvent: PropTypes.func,
 
     agendaDateFormat: dateFormat,
     agendaTimeFormat: dateFormat,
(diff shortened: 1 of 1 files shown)


AgamlaRage:
This is nice! I need this too since few days, to realise something.
That would be great to return the exact same params as the others views, I guess...

```javascript
onClick={(e) => onSelect(event, e)}
```

then 

```javascript
onSelect = (...args) => {
    notify(this.props.onSelectEvent, args);
}
```

Bombe93:
Thank you for appreciation.
I had tried to pass the params as in the other views but it returns errors and so I opted to do this way.
In this way for my application works very well, I always accept any improvement.

AgamlaRage:
Sorry, I changed the name you give to the callback, to follow others view too.
Imagine that your **handleSelectEvent()** is my **onSelect()**

I just changed the params from *(day, events, dayKey)* to *(event, e)*

src/Agenda.js
```diff
@@ -124,7 +131,7 @@ class Agenda extends React.Component {
       let title = get(event, titleAccessor);
 
       return (
-        <tr key={dayKey + '_' + idx} className={className} style={style}>
+        <tr key={dayKey + '_' + idx} className={className} style={style} onClick={()=>this.handleSelectEvent(day, events, dayKey)}>
```
jquense: since this is only passing `events we don't need to use an arrow function here. you can right the `handleSelectEvent` method as `notify(this.props.onSelectEvent, this.props.events)`

AgamlaRage: I don't understand.
Why retreive the whole events array, instead of just the selected event ?

jquense:
Thanks for the PR. I don't think we want to add this here. Generally for this view I've used custom components and specific buttons in those components for event click behavior. Having a "global" click handler on the table row will conflict with folks doing that, and at the moment you can't distinguish or disable the click behavior for one view.

Issue #563, opened 2017-09-28
Add double click handler


AlexMarvelo:
Hello, guys
I think that double click is very common interaction kind among calendars. So it would be useful to have:

1. Handler for double click on an event
2. Handler for double click on an empty cell/empty part of a day cell (to add new event, for example)

What do you think about it?

jquense:
sure, care to submit a PR?

AlexMarvelo:
Yes, in couple of days

PR #567, opened 2017-10-01
Add event doubleClick handler

AlexMarvelo:
This PR is part of #563 issue

New:
* `onDoubleClickEvent` handler
    Callback fired when a calendar event is clicked twice.
    `(event: Object, e: SyntheticEvent) => any`

All changes are similar to `onSelectEvent` handler

### DIFF
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -123,6 +123,15 @@ class Calendar extends React.Component {
     onSelectEvent: PropTypes.func,
 
     /**
+     * Callback fired when a calendar event is clicked twice.
+     *
+     * ```js
+     * (event: Object, e: SyntheticEvent) => any
+     * ```
+     */
+    onDoubleClickEvent: PropTypes.func,
+
+    /**
      * Callback fired when dragging a selection in the Time views.
      *
      * Returning `false` from the handler will prevent a selection.
@@ -645,6 +654,7 @@ class Calendar extends React.Component {
           onNavigate={this.handleNavigate}
           onDrillDown={this.handleDrillDown}
           onSelectEvent={this.handleSelectEvent}
+          onDoubleClickEvent={this.handleDoubleClickEvent}
           onSelectSlot={this.handleSelectSlot}
           onShowMore={this._showMore}
         />
@@ -673,6 +683,10 @@ class Calendar extends React.Component {
   handleSelectEvent = (...args) => {
     notify(this.props.onSelectEvent, args)
   };
+
+  handleDoubleClickEvent = (...args) => {
+    notify(this.props.onDoubleClickEvent, args)
+  }
 
   handleSelectSlot = (slotInfo) => {
     notify(this.props.onSelectSlot, slotInfo)
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -126,7 +126,7 @@ class Calendar extends React.Component {
      * Callback fired when a calendar event is clicked twice.
      *
      * ```js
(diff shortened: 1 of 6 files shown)


src/Calendar.js
```diff
@@ -112,6 +112,24 @@ class Calendar extends React.Component {
     onSelectSlot: PropTypes.func,
 
     /**
+     * Callback fired when empty calendar cell was clicked twice.
```
jquense: "when an empty calendar [...]"

src/Calendar.js
```diff
@@ -112,6 +112,24 @@ class Calendar extends React.Component {
...
+    onDoubleClickSlot: PropTypes.func,
+
+    /**
+     * Callback fired when empty calendar all-day cell was clicked twice.
+     *
+     * ```js
+     * (date: Date, e: SyntheticEvent) => any
+     * ```
+     */
+    onDoubleClickAllDaySlot: PropTypes.func,
```
jquense: I don't think we have a normal `click` equivalent to this one?

src/less/month.less
```diff
@@ -28,6 +28,7 @@
   height: auto;
   line-height: normal;
   white-space: nowrap;
+  pointer-events: all;
```
jquense: ?

src/Calendar.js
AlexMarvelo: I decided to separate `onDoubleClickSlot` and `onDoubleClickAllDaySlot`. Because both of them can get same `date` parameter, but handling behavior is usually different for clicks on regular cells and all-day cells.
What do you think?

src/less/month.less
AlexMarvelo: This point is important.
As `.rbc-row-content` is overlaying `.rbc-day-bg` (in month view), that is listening to `doubleClick` events, we have to add `pointer-events: none` to `.rbc-row-content`, and `pointer-events: all` to all active elements inside it (events, popup link, day number link)

src/Calendar.js
AlexMarvelo: Fixed

src/less/month.less
jquense: I think that may be a problem, first it won't work in older browsers and second I think it is going to break the drag and drop code as well, since they also mess with pointer-events. 

Maybe we can narrow the scope of this PR a bit, and only add doubleClick handlers where there is already a click handler, that will make sure we have a nice parity and we can talk about adding new handlers in a separate PR/issue?

AlexMarvelo: No, I checked. It doesn't make problems with DnD. Moreover, `pointer-events` properties are already used in some places of DnD addon. I can't see any other solutions, how to catch `doubleClick` events under `.rbc-row-content` overlay (remember, that clicked date should be calculated and passed to handler. this is a limitation)

Ok, I'll leave `onDoubleClickEvent` in this PR, and `onDoubleClickSlot` and `onDoubleClickAllDaySlot` will be moved to another one.

AlexMarvelo:
PR description was changed

src/Calendar.js
```diff
@@ -123,6 +123,15 @@ class Calendar extends React.Component {
     onSelectEvent: PropTypes.func,
 
     /**
+     * Callback fired when a calendar event is clicked twice.
+     *
+     * ```js
+     * (event: Object, e: SyntheticEvent) => any
```
jquense: should be `=> void` for the return type i believe

AlexMarvelo: @jquense done

PR #570, opened 2017-10-03
Add doubleClick handlers on empty slots

AlexMarvelo:
This PR is a part of #563 issue

New: 
* `onDoubleClickSlot` handler
    Callback fired when empty calendar cell was clicked twice.
     `(date: Date, e: SyntheticEvent) => any`
* `onDoubleClickAllDaySlot` handler
     Callback fired when empty calendar all-day cell was clicked twice.
     `(date: Date, e: SyntheticEvent) => any`

Changes are similar to `onSelectSlot` handler

### DIFF
--- a/src/BackgroundCells.js
+++ b/src/BackgroundCells.js
@@ -20,6 +20,7 @@ class BackgroundCells extends React.Component {
     longPressThreshold: PropTypes.number,
 
     onSelectSlot: PropTypes.func.isRequired,
+    onDoubleClickSlot: PropTypes.func,
     onSelectEnd: PropTypes.func,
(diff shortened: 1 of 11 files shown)


jquense:
On selectSlot already handles the click case, maybe we can extend that for dblClick without the pointer-events? I want to avoid that css if possible in the core library for browser support. I'm less concerned about it in addons since they are additions.

AlexMarvelo:
Ok, I'll look into Selection object, what we can do with it.
Note, that `pointer-events` have (good browser support)[http://caniuse.com/#search=pointer-events]

AlexMarvelo:
@jquense See another solution in #575

PR #575, opened 2017-10-04
Add doubleClick support to slot selection

AlexMarvelo:
This PR is a part of #563 issue

### Purpose:
Selection object already has support for slot selection with `action: 'click'` property passed to `onSelectSlot`. We have to add support for double click interaction, with `action: 'doubleClick'` property passed.

### Problem:
Now Selection object emits `click` event on every click, so if you click twice, it will emit 2 `click` events with a small delay.

### Solution:
To emit `doubleClick` event inside `Selection` object, we have to check, wether first click is just a single click (then we emit `click` event), or it's the first click of double click (then we should mute it):
1. User makes the first click. Don't emit `click` event yet. Create a timeout for 200ms (usual max delay between 2 clicks of double click). Remeber timestamp of the first click.
2. User makes the second click. Check current time and the timestamp of first click. If it's less than 200ms, then emit `doubleClick` event and clear created timeout, else emit `click` event.
3. If user haven't made the second click, emit `click` event inside created timeout

### Disadvantage: 
`onSelectSlot` handler for `click` events can be called only after 200ms after user click. I don't think we can make this delay less, because normal delay between 2 clicks of `doubleClick` event can be up to 170ms.

### DIFF
--- a/src/BackgroundCells.js
+++ b/src/BackgroundCells.js
@@ -91,6 +91,27 @@ class BackgroundCells extends React.Component {
       longPressThreshold: this.props.longPressThreshold,
     })
 
+    let selectorClicksHandler = (point, actionType) => {
+      if (!isEvent(findDOMNode(this), point)) {
+        let rowBox = getBoundsForNode(node)
+        let { range, rtl } = this.props;
+
+        if (pointInBox(rowBox, point)) {
+          let width = slotWidth(getBoundsForNode(node),  range.length);
+          let currentCell = getCellAtX(rowBox, point.x, width, rtl, range.length);
+
+          this._selectSlot({
+            startIdx: currentCell,
+            endIdx: currentCell,
+            action: actionType,
+          })
+        }
+      }
+
+      this._initial = {}
+      this.setState({ selecting: false })
+    };
+
     selector.on('selecting', box => {
       let { range, rtl } = this.props;
 
@@ -125,26 +146,10 @@ class BackgroundCells extends React.Component {
     })
 
     selector
-      .on('click', point => {
-        if (!isEvent(findDOMNode(this), point)) {
-          let rowBox = getBoundsForNode(node)
-          let { range, rtl } = this.props;
+      .on('click', point => selectorClicksHandler(point, 'click'))
 
-          if (pointInBox(rowBox, point)) {
-            let width = slotWidth(getBoundsForNode(node),  range.length);
-            let currentCell = getCellAtX(rowBox, point.x, width, rtl, range.length);
-
-            this._selectSlot({
-              startIdx: currentCell,
-              endIdx: currentCell,
-              action: 'click',
-            })
-          }
-        }
-
-        this._initial = {}
-        this.setState({ selecting: false })
-      })
+    selector
+      .on('doubleClick', point => selectorClicksHandler(point, 'doubleClick'))
 
     selector
       .on('select', () => {
--- a/src/DayColumn.js
+++ b/src/DayColumn.js
@@ -295,6 +295,13 @@ class DayColumn extends React.Component {
       }
     }
 
+    let selectorClicksHandler = (box, actionType) => {
+      if (!isEvent(findDOMNode(this), box))
+        this._selectSlot({ ...selectionState(box), action: actionType })
+
+      this.setState({ selecting: false })
+    }
+
     selector.on('selecting', maybeSelect)
     selector.on('selectStart', maybeSelect)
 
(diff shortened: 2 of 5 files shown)


src/DayColumn.js
```diff
@@ -295,6 +295,13 @@ class DayColumn extends React.Component {
       }
     }
 
+    let selectorClicksHandler = (box, actionType) => {
+      if (!isEvent(findDOMNode(this), box))
+        this._selectSlot({ ...selectionState(box), action: actionType })
+
+      this.setState({ selecting: false })
+    }
```
AlexMarvelo: Refactor handler functions, as they are absolutely the same for `click` and `doubleClick` events, except `action` property.

src/BackgroundCells.js
```diff
@@ -91,6 +91,27 @@ class BackgroundCells extends React.Component {
...
+            startIdx: currentCell,
+            endIdx: currentCell,
+            action: actionType,
+          })
+        }
+      }
+
+      this._initial = {}
+      this.setState({ selecting: false })
+    };
```
AlexMarvelo: Refactor handler functions, as they are absolutely the same for `click` and `doubleClick` events, except `action` property.

AlexMarvelo:
@jquense Can you take a look at this PR, please

jquense:
hey there, thanks for the PR and your patience. I think the best option is to mimic browser doubleClick behavior, which is that the first click event happens, and if the another click happens within a configurable timespan (defualt 250ms) the _second_ click is emitted as a doubleClick. So for every double click you get one click and one doubleClick event.

jquense:
sorry not sure if ytou are saying it already works like i suggested? Does the curretn code mute the first click? It shouldn't

AlexMarvelo:
Yes, it mutes the first click. I described the reason in 1. of my previous comment

jquense:
Ok that's what I thought. Yes I'm saying we should change that behavior and instead emit the first click. THat's how doubleClick works natively on the web and we should copy that behavior

AlexMarvelo:
Ok, you are right. I just [checked it by myself](https://jsfiddle.net/OcS78wNsbQ/wL97xqum)

AlexMarvelo:
@jquense Done. Now solution looks more simplified. When a user clicks an empty cell:
* if difference with current time and time of last click event is less than 250ms - emit `doubleClick` event
* else - emit `click` event

AlexMarvelo:
@jquense please take a look at this

src/Selection.js
```diff
@@ -243,17 +244,42 @@ class Selection {
...
+      return this.emit('select', bounds)
+  }
+
+  _handleClickEvent(e) {
+    const { pageX, pageY, clientX, clientY } = getEventCoordinates(e);
+    const now = new Date().getTime();
+
+    if (this._lastClickData && now - this._lastClickData.timestamp < clickInterval) {
+      // Double click event
+      this._lastClickData = {
```
jquense: shouldn't this be cleared here? It seems like if you click 3 times really fast you'll get 1 click and 2 dblClick events instead of: click, dblClick, click

AlexMarvelo:
@jquense fixed

AlexMarvelo:
@jquense please take a look at this

src/Selection.js
```diff
@@ -243,17 +244,46 @@ class Selection {
...
+      return this.emit('select', bounds)
+  }
+
+  _handleClickEvent(e) {
+    const { pageX, pageY, clientX, clientY } = getEventCoordinates(e);
+    const now = new Date().getTime();
+
+    if (
+        this._lastClickData &&
+        !this._lastClickData.isDblClick &&
```
jquense: why add this `isDblClick` instead of setting `_lastClickData` to `null` after a double click

AlexMarvelo: I don't see much difference between checking for a flag and checking for null value. `_lastClickData = null` seems obscure for me

jquense: I here you, though the opposite strikes me as more obscure, at the moment we are tracking some metadata that isn't used which isn't clear. I think i'd be clearer to clear out the last click data if there is no last click data :)

AlexMarvelo: Fixed. Clear _lastClickData on double click event

jquense:
thanks a lot!

AlexMarvelo:
Finally! :confetti_ball: Thanks

abhibansal60 (commented):
is this feature still functional?

i want to use doubleClick on onSelectSlot? how can i do that

cutterbl:
If you 'showCode' on the [onDoubleClickEvent prop documentation](http://jquense.github.io/react-big-calendar/examples/index.html?path=/docs/props--on-double-click-event) you'll get an example of how you can handle this.

dr-skot:
Actually for this behavior you need the [onSelectSlot](https://jquense.github.io/react-big-calendar/examples/index.html?path=/docs/props--on-select-slot) prop. Check for `action === 'doubleClick'` in `slotInfo` parameter.

dr-skot:
It looks like there's no access to the `MouseEvent`, or its `target`. If double-clicking on a slot adds a new calendar event, it would be nice to anchor a the popup form to the slots. But I don't see how to get the slot element. Is there a way?

PR #577, opened 2017-10-05
Feature request: resize events

AgamlaRage:
The promise PR from the issue #382 .
Also create a new example page to illustrate it (see the .gif).

---

I left behind some "TODOs" that will be great to do before release. 

**TODO CORE**
1- Do not use a **global var** to control the DnD component render

**TODO EXAMPLE**
2 - Don't use **id** do identify events (didn't get moveEvent() to work properly)
3 - Control views to disable resizing callback in **month** and **agenda** views

---

See example page in action

screen

Issue #382, opened 2017-05-07
Feature request: resize events


pnkflydgr:
When will the ability to resize events be added?

I don't mind attempting to write it if someone could point me in the right direction. I've been coding for a long time but I'm fairly new to react.

jquense:
No plans right now to add it but it would make a good add-on like drag and drop if you want to give it a shot

pnkflydgr:
Can you just tell me which component I would need to override to control events?  Is there already something there like for customView.js.

I would love to write it so it could be used like this (the mouse resize cursor would occur if(onResize()):

```
<DragAndDropCalendar
                    events={this.props.events}
                    style={{height: 800}}
                    selectable='ignoreEvents'
                    defaultView='week'
                    scrollToTime={new Date(1970, 1, 1, 8)}
                    onResize={event => { //stuff }}
                />
```

AgamlaRage:
+1 for this feature !
I didn't dig too much in the core yet, but if I will start somewhere, it will be in the **selection system**.
Maybe it can be used to acheive that.

![screen](https://cloud.githubusercontent.com/assets/15155827/25795630/83e87828-33d6-11e7-9e95-e11b8350af4b.gif)

gazab:
Or should it be implemented as part of the drag and drop addon? Resizing and moving are rather similar.. 

Either way, I would also love this feature!

jnmandal:
This is something I currently need so I'm taking a stab at it now. My plan is to add it as part of the dnd addon, taking advantage of the `components` prop to pass in a gcal-like resize anchor.

pnkflydgr:
Thanks @jnmandal for picking this up.  I had got to the point where I had decided to just make another component for dnd.  Don't have a lot of time right now.

AgamlaRage:
I quickly did something about that. That is working but there is a too heavy render lifecycle in my project to be fully usable. (Sorry, I worked on the compiled files, because I need to live test it on my app, then I didn't do a PR for now)

## react-big-calendar modifications

**[lib/addons/dragAndDrop/index.js](https://github.com/intljusticemission/react-big-calendar/blob/master/src/addons/dragAndDrop/index.js)**

1 - Create and pass through props the new **onEventResize** callback (following the same road as **onEventDrop**) callback


[...]

PR #578, opened 2017-10-05
Resize Month Events and Resize Week/Day Events on Both Ends

arecvlohe:
## Description

This PR adds the ability to resize month events as well as resize week/day events on both ends, top and bottom. I wasn't able to get this to perform correctly on hover which I suppose is ideal. We might find inspiration from #577. 

This is a long standing issue so it would be good to see some movement on this.

Also, I think this could be a good first implementation that is refined on going forward.

### Month
resize-month-event

### Week
resize-event-week

## Issues
#382 

## PRs
#577, #541, #406


arecvlohe:
I did not mean to push this to master. My mistake.

arecvlohe:
Actually, I think I will keep this open

svanetten1976:
@arecvlohe @jnmandal any idea if this excellent feature you guys added is going to get merged into the project?

arecvlohe:
@svanetten1976 I don't have a good idea. @jquense might know better than me. I don't really know what/how things get vetted and merged.

src/EventCell.js
```diff
@@ -49,6 +50,10 @@ class EventCell extends React.Component {
     if (eventPropGetter)
       var { style, className: xClassName } = eventPropGetter(event, start, end, selected);
 
+    if (this.props.resizable) {
```
jquense: These bits should be encapsulated in the addon code, the core libra should require things from addons (so that they are optional)

src/DayColumn.js
```diff
@@ -142,10 +142,8 @@ class DayColumn extends React.Component {
       , showMultiDayTimes
       , culture
       , eventPropGetter
-      , selected
+      , selected, eventTimeRangeFormat, components: { eventComponent }
       , messages
-      , eventComponent
-      , eventTimeRangeFormat
```
jquense: Seems like a rebate error

examples/App.js
```diff
@@ -101,7 +102,7 @@ const Example = React.createClass({
                 <strong><i className='fa fa-code'/>{' View example source code'}</strong>
               </a>
             </div>
-            <Current className='demo' />
+           <Current className='demo' />
```
jquense: ?

examples/demos/dnd.js
```diff
@@ -35,13 +35,40 @@ class Dnd extends React.Component {
     alert(`${event.title} was dropped onto ${event.start}`);
   }
 
+  resizeEvent = (resizeType, { event, start, end }) => {
+    const { events } = this.state;
+
+    // if we want to update the event while dragging we need to find the event
+    // by an identifier. here we use the title and the start time, but depending
+    // on use case you may want have a truly unique id attribute in the event
+    const nextEvents = events.map(existingEvent => {
```
jquense: Why doesn't indexOf work here?

jnmandal: @jquense this was due to a quirk in react-dnd, copying my explanation from #406 

Since only the original event object is accessible through react-dnd... In the onEventResize callback you will not be able to use Array.prototype.indexOf for finding the event again -- unless you continue to mutate the original event and splice it into the cloned events array... but I'm not sure if we would want to recommend that practice

src/DayColumn.js
jnmandal: Actually this is because of a fix I made in #406. eventComponent was being destructured incorrectly in the DayColumn component's renderEvents function (so it was always undefined). We need to be able to specify custom eventComponent for the addOn to work properly.. But its possible this has since been fixed some other way.

examples/demos/dnd.js
jquense: Ideally in the next iteration of the callback the new event will be there? Why can't we get react dnd to update the thing it's referencing. I don't love this logic is fairly flakey even for an example and likely to lead folks to copying it in their own code with potential issues. Last ditch we should at least add an id to events in the example and use that vs title and date instances

src/EventCell.js
jquense: ping

src/addons/dragAndDrop/ResizableEvent.js
```diff
@@ -0,0 +1,56 @@
...
+        {Top}
+        {title}
+        {Bottom}
+      </div>
+    );
+  }
+}
+
+const eventSourceTop = {
+  beginDrag: ({ event }) => ({ ...event, type: 'resizeT' })
```
jquense: 'resizeTop'

src/addons/dragAndDrop/ResizableEvent.js
```diff
@@ -0,0 +1,56 @@
...
+}
+
+const eventSourceBottom = {
+  beginDrag: ({ event }) => ({ ...event, type: 'resizeB' })
```
jquense: 'resizeBottom'

src/addons/dragAndDrop/backgroundWrapper.js
```diff
@@ -72,6 +72,20 @@ class DraggableBackgroundWrapper extends React.Component {
...
+  componentWillReceiveProps(nextProps) {
+    const { isOver: wasOver} = this.props;
+    const { isOver } = nextProps;
+    if (isOver && !wasOver) {
+      const { onEventResize, dragDropManager } = this.context;
+      const { value } = this.props;
+      const monitor = dragDropManager.getMonitor()
+      if (monitor.getItemType() === 'resize') {
+        // This was causing me performance issues so I commented it out. Thoughts? - Adam Recvlohe
+        // onEventResize('drag', {event: monitor.getItem(), end: value});
```
jquense: what's going on here?

src/addons/dragAndDrop/compose.js
```diff
@@ -0,0 +1,14 @@
+// Exact copy over from recompose library
+// https://github.com/acdlite/recompose/blob/master/src/packages/recompose/compose.js
+
+export default function compose(...funcs) {
```
jquense: if this is only used once can we just compose by hand and avoid vendoring this?

jquense:
looking great!

src/EventCell.js
arecvlohe: I see what you mean here. In our implementation at work I ended up removing this.

src/addons/dragAndDrop/backgroundWrapper.js
arecvlohe: I honestly don't know. This was carried over I think from what @jnmandal was doing I think.

src/addons/dragAndDrop/compose.js
arecvlohe: Yeah, sounds good!

arecvlohe:
@jquense Thanks for taking a look at this some more and making comments! I will try and work on it this weekend and push some updates. Anyone else feel free to take this up and push it forward if you have time.

examples/App.js
arecvlohe: It looks like a space was removed.

src/DayColumn.js
arecvlohe: Instead of using event component. I updated the addon to pass the ResizeEvent to `components: { day: { event }, week:  { event } }` as opposed to creating a new prop inside of components. I think this is how the library is intended to be used.

arecvlohe:
@jquense Howdy! I made changes based on your comments. Please let me know if this works and if I need to make any more changes!

src/addons/dragAndDrop/backgroundWrapper.js
jquense: is think broken? If it's not needed lets remove it, otherwise don't we need to fix this?

src/EventCell.js
```diff
@@ -4,6 +4,7 @@ import cn from 'classnames';
 import dates from './utils/dates';
 import { accessor, elementType } from './utils/propTypes';
 import { accessor as get } from './utils/accessors';
+import ResizableMonthEvent from './addons/dragAndDrop/ResizableMonthEvent';
```
jquense: rempve

src/Calendar.js
```diff
@@ -609,7 +609,7 @@ class Calendar extends React.Component {
     let names = viewNames(this.props.views)
 
     let viewComponents = defaults(
-      components[view] || {},
+      components[view],
```
jquense: put the `|| {}` back

examples/App.js
jquense: can we fix it?

arecvlohe:
@jquense I made the requested changes. Let me know if there is anything else I need to do!

src/addons/dragAndDrop/backgroundWrapper.js
jquense: Ping

AlexMarvelo:
Any progress here?

jquense:
Nope, waiting on updates

AlexMarvelo:
There are only 2 small conflicts in `examples/events.js`. @arecvlohe please, can you take a look?

AlexMarvelo:
I'm wondering, why DnD behavior for top dragging and bottom dragging is different (see second "week" gif in description)?
1) If you drag top side of event, new position will be set to highlighted cell exactly
2) But if you drag bottom side, new position will be set to a previous cell before highlighted one
Why?

arecvlohe:
I will take a look @AlexMarvelo at what you are referring to.

carnesik:
Thanks for the work guys! We are really looking forward to building this into our app here at DNSFilter once changes are approved.

arecvlohe:
I have been busy the past few weekends so I should be able to get to it this weekend (12/09/2017). Or during the week if I feel up for it 😄

src/addons/dragAndDrop/backgroundWrapper.js
arecvlohe: Removed

arecvlohe:
Can someone show me how to merge these conflict and get them into my local branch?

arecvlohe:
Everytime I think I have resolved them it still keeps showing them as conflicts

examples/App.js
arecvlohe: You mean put the space back in?

arecvlohe:
@AlexMarvelo I think I addressed your issue with the dnd bottom. I think the information coming back from the hover event is off by 30 min so I added 30 min.

arecvlohe:
Okay, I think I got it

arecvlohe:
Let me know what else I need to do.

AlexMarvelo:
:thumbsup: Bottom DnD looks good for me now

AlexMarvelo:
But all-day event's behavior sometimes seems unexpectable.
1. All-day items can be resized horizontally only. Vertical resize is disabled 
1. Right/left side icon should be a horizontal arrow, not vertical

Do you agree, that all-day items should be more similar to items on month tab?
@arecvlohe I can help and open a PR to your fork if you want

arecvlohe:
@AlexMarvelo Ah I did not notice that behavior and I agree it is unexpected. Sure, If you can work on this feel free to fork and make changes!

arecvlohe:
@AlexMarvelo I made an update based on your recommendation but I am not sure if it is a good solution. I think I am missing some things. Let me know what you think!

arecvlohe:
@jquense @AlexMarvelo How are things looking?

arecvlohe:
The changes look good to @jquense so if I don't hear from @AlexMarvelo in the next couple days I will merge this on Thursday (12/21/2017) before the holiday break.

jquense:
@arecvlohe can you also please update the Changelog (you can put the entries in an unreleased heading)

AlexMarvelo:
Hello
Unfortunately, I still see the same issue
1. Sometimes vertical arrow icon appears on hover for all-day items. To change the duration of all-day item (vertical resize) seems not clear to me
1. A short draggable line sometimes appears on all-day items. I think it shouldn't

I guess in some cases vertical arrow icon should be replaced with horizontal. Will check out the code and give more details tomorrow 

dec-19-2017 00-48-12

arecvlohe:
Ah I see, I will check it out some more tonight @AlexMarvelo

arecvlohe:
@AlexMarvelo I am a little lost in terms of trying to fix this. DayColumn is not passing any information related to the events. I am not sure why this is happening.

arecvlohe:
I could use some help with this but I will try and look at it this weekend.

AlexMarvelo:
@arecvlohe Check the PR I made to your fork with header all-day event fixes https://github.com/jnmandal/react-big-calendar/pull/5

arecvlohe:
@AlexMarvelo mentioned creating a separate PR for removing ResizableMonthEvent. This can be handled at a separate time.

arecvlohe:
I will review this more later today and push any needed changes

arecvlohe:
@jquense When I try to `npm run examples` I get an error in the console saying:

screen shot 2018-01-06 at 1 58 54 pm

I believe the reason for this is `markdown-jsx-loader` is using `createClass`, but `react-big-calendar` is using `React@^16.2.0`. This version of react does not have `createClass` as part of its API as per the [docs](https://reactjs.org/blog/2017/09/26/react-v16.0.html#packaging). I am a bit confused as to why the examples would be working, maybe because the server build is using React 15. I don't really know. All I do know is that I updated the project to include `create-react-class` and replace all instances of `createClass` with `createReactClass`. Until I can get the markdown component `Intro` working I won't be merging this in.

arecvlohe:
Also, I think the error with the deployment comes from not being able to resolve the correct dependency for the `require-from-string` module. I am able to resolve this issue locally but I don't think it's being handled properly on TravisCI. See below:

<a href="https://cl.ly/21052r1c2D2R" target="_blank"></a>

arecvlohe:
Other than these issues above everything seems to be working fine.

tobiasandersen:
Sorry, I haven't looked at the code for this until just now, so I've already fixed the `createClass` issues on master by updating to es6 (rather than adding the extra package). Would you mind updating your PR?

arecvlohe:
@tobiasandersen I have merged in master and your ES6 changes but still face a similar issue of the examples not working that you found in #676

AlexMarvelo:
@jquense updated React to v16 in #671, but it conflicts with markdown-jsx-loader. We don't need extra `create-react-class` dep, it should be removed. The problem with loader I will try to fix in separate PR

arecvlohe:
@AlexMarvelo I removed the `create-react-class` dep already.

arecvlohe:
@AlexMarvelo Are you then saying this PR is ready?

AlexMarvelo:
#681

arecvlohe:
For those scrolling through here, this PR will not be merged until dependencies are updated for React 16.

AlexMarvelo:
Yes. Major version update of React is a big deal. I don't know why it was updated and not tested. Maybe it's easier to revert back to v15 and make the update later, with all necessary fixes.

tobiasandersen:
> I don't know why it was updated and not tested.

I'd assume this was because it was forgotten or somehow missed. Maybe it was tested, but with older node_modules or something.  No need for the harsh tone. 

> Maybe it's easier to revert back to v15 and make the update later, with all necessary fixes.

I'm not sure what's easiest here, but we definitely have to fix this before releasing a new version. I guess it's as easy as reverting back to v15.3.0 for `react` and `react-dom`? Ideally we update the deps so that we can use react 16 though :) 

I believe there're some other problems on master as well (events not showing up in day view etc.). Let's get master to a good state, and then we can merge this!

AlexMarvelo:
Now let's fix the problem with markup loader, described in #681

AlexMarvelo:
master is ok now. @arecvlohe try to merge master and check, please

arecvlohe:
@AlexMarvelo everything works as expected on my local branch!

tobiasandersen:
👍 Awesome! But there's quite a lot of noise here. Are all those commits and line changes necessary/relevant to this PR? cc @jquense

arecvlohe:
@tobiasandersen Are you saying to rebase or remove things?

tobiasandersen:
Rebase (or squash merge) :) I saw now that the heavy line changes was due the example bundle, which I guess makes sense!

arecvlohe:
I will rebase some changes to remove some fluff from the commit history @tobiasandersen

arecvlohe:
@tobiasandersen Doing a rebase is too insane, at least for me, with all the commits that have been merged over the past couple days. But I also think I missed what you were saying. Did you mean I can merge, just choose either the squash merge or rebase merge option?

tobiasandersen:
**Squash and merge** would combine all your 47 commits into one. I usually prefer having 1 (or a few maximum) commits per PR. It keeps the history clean, and makes it easier when going back to find when and why things were introduced. However, it also removes the actual history. And I see that you've been a few people working on this – squashing would make it look like you wrote all of it.

I'd prefer a squash over the 47 commits if I were to choose from the two. But it's @jquense's project, so I think it's better if he decides what's best :)

arecvlohe:
@tobiasandersen I hear you. Let's see what @jquense says...

jquense:
@arecvlohe You do whatever makes your life a bit easier to handle the rebase. Generally I think it's worth leaving the branch unsquashed and then squashing via the github butten whe nwe merge so that master ends up with one commit, and the original history is maintained. In this case tho i can imagine that the rebase is super painful without squashing so do what makes it's easier for you. We'll be sure to shout to out everyone who contributed to this PR

arecvlohe:
I ran into another hiccup. After merging in master, in the DnD example I am not able to drag and drop events in the week and day view. However, month view works fine.

<a href="https://cl.ly/2l473d05223V" target="_blank"></a>

I will have to dig into this at some point but there is a lot to dig so any help is appreciated.

arecvlohe:
@tobiasandersen @AlexMarvelo @jquense  ^^^

tobiasandersen:
I wouldn't be surprised if it's due to my latest PR (although I did test the DnD example before merging). I'll have look and get back.

arecvlohe:
I will take a look this evening and see what I can find out

tobiasandersen:
Cool! If it's anything from my last PR, it's probably this extra div that I added: https://github.com/intljusticemission/react-big-calendar/blob/0b2c2347efa8fe4fb77f7931634de9a8ee1362ac/src/DayColumn.js#L132.

If you don't find the problem I'll have a check when I wake up tomorrow!

arecvlohe:
@tobiasandersen From what I can tell it looks like the event container is the culprit. Unfortunately in `react-dnd` there is not a way to drop on a div and have the drop be handled by a div underneath. So basically it just covers where it can be dropped and that is why nothing is happening. One way to maybe get around this is to set the z-index of the date cells to > 1 when there is a drag event.

tobiasandersen:
Sorry about that, should have tested it more than I did. I'm thinking `pointer-events: none` on that div should be enough?

arecvlohe:
I will check that out and see

arecvlohe:
It works! I will wait for a successful build to complete first and then move forward with merging into master

arecvlohe:
It is done 😄

tobiasandersen:
Awesome! Great work! 🎉

jadczyk:
Hi, after your merge it seems you cannot select events in weekly view anymore when not using drag'n'drop. You added `    pointer-events: none;` to `.rbc-events-container` but didn't add `pointer-events: all;` to `.rbc-event`

arecvlohe:
Good catch @jadczyk!

arecvlohe:
@jadczyk Mind creating a PR to fix this?

jadczyk:
Just did #690

ehahn9:
Just started using the resize/DnD stuff - thanks @arecvlohe & others - hugely useful!

tl;dr: I'm looking for advice on how to best marry DnD with custom components.

When creating the underling calendar, it looks like the addon's `withDragAndDrop` HOC overwrites a bunch of `component:` props to add DnD. Alas, I have a custom `Event` component in day/week/month views (and in a custom view), so this is a bit of a challenge.

Would something like this make sense:

```
   const DnDCalendar = withDragAndDrop(BigCalendar)
   <DnDCalendar
     components={{
       week: {
         event: myCustomEventComponent
       }
     }}
   />
```
becomes instead:

```
   <DnDCalendar
     components={{
       week: {
         event: withDragAndDrop.customComponent(myCustomEventComponent)
       }
     }}
   />
```
The new HOC would wrap the user's components with one based on the current `ResizableEvent` (yes, there would have to be a month view variant, too).

I'm keen to get folks' other ideas/approaches (or validation of this one). I'm probably misunderstanding the DnD philosophy, but it was a surprise to see `examples/dnd` basically overwrite custom component props.

thx for any guidance/advice.

AlexMarvelo:
I think that no more additional work should be required to enable DnD except wrapping the calendar with `withDragAndDrop`. It's  the best practice. I will take a look, what can be done. I see, that DnD addon does not really take care of custom event renderer now.

PR #589, opened 2017-10-18
Give to the dateCellWrapper a base className

GGGGino:
I start from my needs: I want to add some classes in the cell of the month view.

**Specific**
When I override the default dateCellWrapper and I want to add some classes classes, now I have to return `this.props.children` otherwise I lose the base classes names(rbc-off-range-bg, rbc-selected-cell)

**Solution proposed**
I put the result of the 'cn' function in a variable and I pass it as prop in the `Wrapper`

### DIFF
--- a/src/BackgroundCells.js
+++ b/src/BackgroundCells.js
@@ -62,22 +62,25 @@ class BackgroundCells extends React.Component {
     return (
       <div className='rbc-row-bg'>
         {range.map((date, index) => {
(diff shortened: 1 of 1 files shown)


jquense:
Hi there, you don't need to pass them here in the wrapper you can use React.cloneElement and use classnames there to add classes

GGGGino:
You're right, I thought this was more readable but i'll try with your solution

PR #590, opened 2017-10-19
Added day param to the cell wrapper

GGGGino:
Hi,

my need is to customize the cell wrapper from a given array of object(array of specialDays), but when I override the cell Wrapper I can't retrieve the exact item I need from this array.

In this PR i create a new prop `specialDays` that work similar to the `event` prop. 

```javascript
const specialDays = [
        {
          'day': new Date(2015, 3, 0),
          'classNameToAdd': "try"
        },
        {
          'day': new Date(2017, 3, 0),
        }
      ];

<BigCalendar
        events={events}
        specialDays={specialDays}
      />
```
_The required property in the specialDays item is `day`._

The prop `specialDays` is passed to the custom cellWrapper as prop so i can customize the cell in according to the`specialDays` array.

```javascript
const CustomMonthCell = React.createClass({
  render() {
      let { range, day } = this.props;


      const baseClasses = this.props.children.props.className,
        defaultStyle = segStyle(1, range.length);

      if( day ){
        defaultStyle.background = 'red';
      }

      return (
          <div style={defaultStyle}
               className={baseClasses}>
              {this.props.children}
          </div>
      )
  }
});
```

## Result
schermata 2017-10-19 alle 15 12 35

### DIFF
--- a/src/BackgroundCells.js
+++ b/src/BackgroundCells.js
@@ -56,18 +56,28 @@ class BackgroundCells extends React.Component {
   }
 
   render(){
-    let { range, cellWrapperComponent: Wrapper, date: currentDate } = this.props;
+    let { range, cellWrapperComponent: Wrapper, date: currentDate, specialDays } = this.props;
     let { selecting, startIdx, endIdx } = this.state;
 
     return (
       <div className='rbc-row-bg'>
         {range.map((date, index) => {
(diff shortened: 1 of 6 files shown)


jquense:
thanks for the PR, I don't want add arbitrary props here, as it's not likely to scale well. You can already accomplish this by using `context` like the Drag and drop addon does, check that out.

Issue #592, opened 2017-10-20
Last row of events in month view not rendered correctly


peterdivvito:
I noticed that sometimes the last events for a day in the month view are not rendered correctly.
 
![calendarscreenshot](https://user-images.githubusercontent.com/29786286/31799698-0d74f380-b57f-11e7-8028-d28bb56c4cd7.jpg)

I tracked this down to DateContentRow not passing on the start and end accessors to EventEndingRow.

I'll try and create a pull request to fix this when I get a chance.

Thanks

tobiasandersen:
I take it that this is solved by https://github.com/intljusticemission/react-big-calendar/pull/593. Feel free to re-open otherwise!

PR #593, opened 2017-10-20
Issues/592 event ending row fix

peterdivvito:
Fix for issue #592 

I added some extra birthday events to the examples folder to prove the bug fix.


jquense:
thanks!

PR #599, opened 2017-10-29
fix agendaHeaderFormat in Calendar

cymen:
The documentation is showing the wrong information for `agendaHeaderFormat` -- it is `dateRangeFormat` not `dateFormat`. I tracked it down to it being set wrong in `src/Calendar.js`.


jquense:
thanks!

PR #600, opened 2017-10-29
Closes #524

### LINKED ISSUE #524: styles included in eventPropGetter can be overwrite at template definition

yury-stepanov:
Added ability to override `right`, 'left' and 'width' properties

### DIFF
--- a/src/DayColumn.js
+++ b/src/DayColumn.js
@@ -201,11 +201,13 @@ class DayColumn extends React.Component {
         <EventWrapper event={event} key={'evt_' + idx}>
           <div
             style={{
+              width: `${width}%`,
+              [isRtl ? 'right' : 'left']: `${Math.max(0, xOffset)}%`,
+              // We allow override all styles, apart from top and height,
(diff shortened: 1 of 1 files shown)


jquense:
see my comment in #524 Thanks for the effort

Issue #587, opened 2017-10-17
Drag-and-drop issuse


vasilevich:
## There is an issue with the drag and drop addon.
in basic calendar
when you click on a day number like this:
![image](https://user-images.githubusercontent.com/1217224/31667891-610b5c8c-b359-11e7-9f39-6c34763af9ad.png)

it goes to the Day time of that particular date:
![image](https://user-images.githubusercontent.com/1217224/31667923-7468c1fc-b359-11e7-934e-0110800867b0.png)

this feature does not seem to work in drag-and-drop,
the day number is unclickable.
![image](https://user-images.githubusercontent.com/1217224/31667961-906026b6-b359-11e7-8c6e-14dc90c08cc5.png)

this seems not to happen unless you add drag/drop.

I suspect the reason for that is, because the cell is being overidden not properly in [here](https://github.com/intljusticemission/react-big-calendar/blob/master/src/addons/dragAndDrop/index.js#L76).

## and while you are at it, may I suggest an useful feature in my opinion:
in addition to providing a "onEventDrop" event, please provide, "onEventDrag" and "onEventMove" which is activated after the drag, and sends x/y etc, durning movement. this will let us create more visually appealing drag and drop functionality.

thank you so much for your hard work!
good luck!

jquense:
Does seem like a bug, unfortunately we don't provide great support for addons since they are community donated. AS such the best way to get it fixed is to summit a patch :)

vasilevich:
Dear jquense.
While I agree that this is a community addon and I will definetly try to provide a patch for this bug.
in my opinion you should leave the case open until fixed so that the community is aware and can recognize the bug, since. because even tho it is a community addon it is still a part of the codebase(in the same repository). please consider leaving the issuse open or maybe it should be a sub-git of the relevant repository rather than the code direcctly.
just an opinion.
regardless thank you for your time and best of luck.

vasilevich:
I made a simple fix:

PR #606, opened 2017-11-03
fixed click when clicking on date in month when drag and drop active.…

vasilevich:
Please see issuse:
https://github.com/intljusticemission/react-big-calendar/issues/587#issuecomment-341447496

this is a patch,

### DIFF
--- a/src/Agenda.js
+++ b/src/Agenda.js
@@ -3,7 +3,6 @@ import React from 'react'
 import classes from 'dom-helpers/class'
 import getWidth from 'dom-helpers/query/width'
 import scrollbarSize from 'dom-helpers/util/scrollbarSize'
-import cn from 'classnames'
 
 import localizer from './localizer'
 import message from './utils/messages'
@@ -110,13 +109,12 @@ class Agenda extends React.Component {
     events = events.filter(e => inRange(e, day, day, this.props))
 
     return events.map((event, idx) => {
-      const _isSelected = isSelected(event, selected)
       const { className, style } = eventPropGetter
         ? eventPropGetter(
             event,
             get(event, startAccessor),
             get(event, endAccessor),
-            _isSelected
+            isSelected(event, selected)
           )
         : {}
       let dateLabel =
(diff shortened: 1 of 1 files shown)


src/addons/dragAndDrop/styles.less
```diff
@@ -2,7 +2,6 @@
 
 .rbc-addons-dnd {
   .rbc-row-content {
-    pointer-events: none;
```
tobiasandersen: I agree that `pointer-events: none` is a bit problematic here, but removing this introduces new problems when doing DnD. Could it be added only when dragging is active?

src/addons/dragAndDrop/styles.less
```diff
@@ -12,10 +11,10 @@
 
   .rbc-addons-dnd-over {
     background-color: rgba(
-      red(@date-selection-bg-color),
-      green(@date-selection-bg-color),
-      blue(@date-selection-bg-color),
-      .3
+            red(@date-selection-bg-color),
```
tobiasandersen: Why the extra indentation?

vasilevich: probably by accident.

src/addons/dragAndDrop/styles.less
vasilevich: can you specify which problems occur? I didn't notice any so far.

tobiasandersen: Part of the day slot in the month view becomes undroppable:

demo

vasilevich: ahh, I see, I'll look into this shortly with the latest calendar update.

arecvlohe:
@vasilevich How is this coming along? Need any help?

vasilevich:
@arecvlohe hi sorry about the delay, I just been very busy, if you would like you can try to implement this, I haven't done much about it ever since, but I will get to it once I can.

jquense:
sorry we didn't get to merge this! Happy to look again if you want resend it with conflicts addressed

Issue #609, opened 2017-11-06
Add optional onDrillDown to the Calendar


kurenov:
Currently experiencing problems with `onDrillDown()`, which is fired when header date is clicked. `onDrillDown()` first changes view and then changes date, this kind of order causes problems in my application (events for wrong date are being fetched for custom views).

I propose to have optional `onDrillDown` prop, which will override built-in behavior.
Moreover, there is related issue #503 which was not address and consequently closed

Issue #503, opened 2017-08-15
Make onDrillDown a prop of Calendar


mprobus:
It'd be great to be able to handle an onDrillDown event with a handler that could process change in both view and date in one pass.   Currently, I get two callbacks when user clicks on a TimeGrid header: first, the onView, and then the onNavigate.

jquense:
Sounds good! Feel free submit a PR :)

PR #610, opened 2017-11-06
Added 'onDrillDown' optional prop

kurenov:
Resolves https://github.com/intljusticemission/react-big-calendar/issues/609 & https://github.com/intljusticemission/react-big-calendar/issues/503

### DIFF
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -96,6 +96,12 @@ class Calendar extends React.Component {
     onView: PropTypes.func,
 
     /**
+     * Callback fired when date header, or the truncated events links are clicked
+     *
+     */
+    onDrillDown: PropTypes.func,
+
+    /**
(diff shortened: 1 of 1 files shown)


src/Calendar.js
```diff
@@ -693,10 +699,16 @@ class Calendar extends React.Component {
   };
 
   handleDrillDown = (date, view) => {
-    if (view)
-      this.handleViewChange(view)
+    const { onDrillDown } = this.props;
+    if (onDrillDown) {
+      onDrillDown(date, view, this.drilldownView)
```
jquense: `return` early here then you can avoid the `else` block around the rest of the function

kurenov: @jquense thanks for a quick response and review =) Done

jquense:
thanks!

kurenov:
@jquense you're welcome! Btw I did not bumped the version. Is it ok?

PR #611, opened 2017-11-06
Bump patch version (0.17.1)

kurenov:
Since the Calendar was patched, bumped version to 0.17.1


jquense:
thank, there is no need to PR version bumps i handle it when i push releases 👍

Issue #539, opened 2017-09-08
Lack of test coverage


choruk:
I love this project and I really appreciate @jquense for creating/maintaining it (and also everyone who has contributed to it so far). You guys rock!

That being said, the lack of test coverage really makes me uneasy, especially when I see so many different people making contributions. It feels like the likelihood of someone inadvertently introducing new bugs is pretty high.

What do others think about this test coverage issue? Am I the only one that finds it a bit unsettling?

Also, and this question is probably better suited for @jquense since he is the maintainer, but if one were to add test coverage, what would be the desired libraries to use?

In past React projects, I have worked with enzyme ([which needs jsdom to use the mount API](http://airbnb.io/enzyme/docs/guides/jsdom.html)), mocha, chai, and sinon. These libraries have worked well for me, but I am open to other suggestions if there are better options out there.

jquense:
I think we all find the lack of test coverage short of ideal. Historically it was less important for us because we the app this component is used in had test coverage. The reason it still doesn't have tests is that I don't have time, and frankly it's just always lower on the priorities list than other stuff i need to get done. 

I would *LOVE* if folks wanted to contribute tests, it would be greatly appreciated and would almost certainly increase the ease and speed of merging PR's.

In terms of libraries, I tend to use Jest, Chai, Sinon, and [teaspoon](https://github.com/jquense/teaspoon) (but feel free to swap that out for enzyme if you want). Thanks for taking interest here :) It's very needed

choruk:
Thanks for the quick response! That definitely makes sense and I can totally understand just not having the time and having other, more pressing priorities. I was mostly curious because I wanted to make sure any efforts to add test coverage (either by myself or others) would not be in vain (e.g. in case you were against it or had strong preferences about which testing libraries to use).

Two more questions:
1. What do you think about adding partial test coverage?
  There is a lot of code to test, so it would be quite the uphill battle to test it all in one go, but I could also understand the line of thinking that it is not that worth it unless it is all covered. Honestly, I could go either way on that one, but leaning more towards partial test coverage since there is just so much code to test. To be clear, when I say partial test coverage, I am thinking maybe starting at the lowest level components or the util functions and fully testing those first.

2. What do you think about using a continuous integration tool like Travis CI / CircleCI? Do you have any preferences there?
  I have worked with both and would be happy to play around with setting that up on my fork to get it working right before setting that up on this repo.

jquense:
[...]

PR #615, opened 2017-11-08
Add test coverage for event levels utils

choruk:
Beginning to address comments raised in #539. I definitely dropped the ball on getting around to doing this after opening that issue, but I finally got some time and was able to add full test coverage of the event levels utility functions. I have only added Jest for now, as that is the only testing library needed at this stage. I also haven't played around with the Travis setup yet, but I noticed you have a `.travis.yml` file, so not sure if you already have something going. Let me know and I can try to look into that over the weekend, but I wanted to open this PR so that you can start reviewing it in the mean time at least.


jquense:
this is awesome thanks!

jquense:
thanks a ton @choruk this was super helpful :)

Issue #614, opened 2017-11-08
onSelectSlot callback is called instead of onSelectEvent (sometimes)


remituveri:
Sometimes, clicking on an event will execute the onSelectSlot callback (instead of onSelectEvent)
This is caused by the document.elementFromPoint (in isEvent function of Selection.js) which will return the span contained in the rbc-slot-selection div...
Removing the following code from DayColumn.js fixed that issue but I'm not sure 100% that it's the better way to do it. rbc-slot-selection does not seems to be used

&& _react2.default.createElement(
        'div',
        { className: 'rbc-slot-selection', style: style },
        _react2.default.createElement(
          'span',
          null,
          _localizer2.default.format(selectDates, selectRangeFormat, culture)
        )
      )

jquense:
Sorry folks I can't reproduce on the examples, perhaps it was inadvertently fixed recently

arximughal:
@jquense it's not fixed I guess. I am facing the same issue! 
It's on the demo page as well! 
**Reproduce:**

* Go to [DEMO](http://intljusticemission.github.io/react-big-calendar/examples/index.html)
* Navigate to the **Selectable** event tab
* Click on the events there

Events in **all day** group seems to be working just fine. But the events underneath aren't working! 
Click a couple of times on different parts of the events e.g., Click on Time, Click on any place other than the title...

The solution by SamHurne in https://github.com/intljusticemission/react-big-calendar/issues/630 seemed to have fixed the issue previously. But we updated the package recently and that fix is not working either!

arximughal:
![rbc](https://user-images.githubusercontent.com/8762441/39239195-99119e68-4899-11e8-9e03-4f294ebcdf13.png)
@jquense the issue occurs on some events in different scenarios!

noagani:
I had this problem. Resolved it very simple: I gave the event class z-index:999 so when clicking an event it will catch the event and not the slot below.

llewellyn1411:
This issue is still happening for me. 

@noagani that fix did not work for me

arximughal:
@noagani that was the first thing I tried. Did work for sometime. But recently, after upgrading to the latest version, the `z-index` stopped working for me.

ash-PS:
Also still happening for me.

PR #624, opened 2017-11-17
onSelectSlot callback is called instead of onSelectEvent (sometimes) …

remituveri:
…#614

Sometimes, clicking on an event will execute the onSelectSlot callback (instead of onSelectEvent)
This is caused by the document.elementFromPoint (in isEvent function of Selection.js) which will return the span contained in the rbc-slot-selection div...
Removing the following code from DayColumn.js fixed that issue but I'm not sure 100% that it's the better way to do it. rbc-slot-selection does not seems to be used

&& _react2.default.createElement(
'div',
{ className: 'rbc-slot-selection', style: style },
_react2.default.createElement(
'span',
null,
_localizer2.default.format(selectDates, selectRangeFormat, culture)
)
)

### DIFF
--- a/src/DayColumn.js
+++ b/src/DayColumn.js
@@ -122,14 +122,6 @@ class DayColumn extends React.Component {
         step={step}
       >
         {this.renderEvents()}
-
-        {selecting &&
-          <div className='rbc-slot-selection' style={style}>
-              <span>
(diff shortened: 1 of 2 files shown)


jquense:
this will remove the visual feedback of the selection state.

PR #628, opened 2017-11-28
Migrated to storybook3 with the necessary webpack config changes

frehner:
In order to help profile performance issues that are blocking #423 , it is incredibly helpful to [upgrade storybook](https://github.com/storybooks/storybook/pull/623) so that you can use [`react_perf` url parameter](https://reactjs.org/blog/2016/11/16/react-v15.4.0.html#profiling-components-with-chrome-timeline)


jquense:
thanks!

PR #642, opened 2017-12-12
Fix agenda title

gabrielsanterre:
Pass missing `length` option to the `VIEW.title()` function.

The toolbar title in the agenda view was always using the default `length` (30) from the Agenda component.
example

Now when we pass the length to the Calendar component it will also change the agenda title accordingly.


jquense:
thanks!

PR #652, opened 2017-12-19
Fix event listing in month view for FF and Edge

hawkgs:
This PR introduces a fix for a bug which is reproducible in month view of the calendar. Basically, the events in the day boxes are not displayed properly due to the `overflow: hidden` of `rbc-month-row` class.

The screenshots below show how the calendar looks with and without `overflow: hidden`

fix

The issue can be reproduced in:
 - MS Edge v41
 - Firefox Quantum (v57)


jquense:
Thanks for the PR, I'm always a bit wary when we change css since it can easily cause a regression elsewhere. From your screenshots it doesn't look like your styles are correct even in the "after" shot. Perhaps you have other conflicting styles on the page?

hawkgs:
Yeah, it looks like it doesn't look the same way. Haven't noticed this to be honest. Anyway, we are pretty much using the component straight out of the box with only few minor style changes. The issue is still reproducible with or without them. Maybe something else is affecting the component. I'll dig into it.

jquense:
Imma close this for now since it seems like there are other issues at play with your styling. Feel free to resubmit something if there is still a bug

PR #658, opened 2017-12-27
Fix CustomView

arkgast:


examples/demos/customView.js
```diff
@@ -44,18 +51,23 @@ MyWeek.navigate = (date, action) => {
...
-      defaultDate={new Date(2015, 3, 1)}
-      views={{ month: true, week: MyWeek }}
-      test="io"
-    />
-  </div>
-)
-
-export default CustomView
+
+let CustomView = React.createClass({
```
tobiasandersen: Could you update this to use es6 instead?

arkgast: yeah, no problem

arkgast:
Hello @tobiasandersen, I think the example was about to set a custom start/end of a week.

As you can see [here][code] the purpose was to get a range but the [date-arithmetic] api right now doesn't have that method.

[code]: https://github.com/intljusticemission/react-big-calendar/pull/658/files#diff-c25d8e64b4c528fe1e1e86385af983ffR20
[date-arithmetic]: https://www.npmjs.com/package/date-arithmetic

tobiasandersen:
Okay cool, I'll just let @jquense have a look at it as well before merging!

PR #663, opened 2017-12-30
Avoid unecessary anonymous function

jdeniau:
Avoid creating anonymous function, specially in the render method.


jquense:
Thanks!

PR #671, opened 2018-01-04
Fixup resource PR

jquense:


AlexMarvelo:
You've updated React to v16 here. Did you test the app after that? Because it's not working for me anymore after this commit. Perhaps React v16 conflicts with markdown-jsx-loader v2

vongdarakia:
I just tested this on master, and although it works for month an week view, it doesn't work for the Day and Agenda views.

No events are shown for the Day and Agenda views.

vongdarakia:
I know what the problem is. It was from this update (#638). The inRange function doesn't work when the start and end date are the same.

PR #673, opened 2018-01-06
Added ability to hide gutter and time indicator

paustint:
Added the ability to pass in two additional optional props  and
hideGutter=true will hide the gutter in an view where there is a time sidebar (except agenda view)
hideTimeIndicator=true will hide the time indicator in views where this is shown

If the gutter is hidden, then the time indicator will also be hidden

Impacted Views:
**week**, **work_week**, **day**

image

image

### DIFF
--- a/src/TimeGrid.js
+++ b/src/TimeGrid.js
@@ -53,6 +53,9 @@ export default class TimeGrid extends Component {
     selectable: PropTypes.oneOf([true, false, 'ignoreEvents']),
     longPressThreshold: PropTypes.number,
 
+    hideGutter: PropTypes.bool,
+    hideTimeIndicator: PropTypes.bool,
+
     onNavigate: PropTypes.func,
     onSelectSlot: PropTypes.func,
     onSelectEnd: PropTypes.func,
@@ -148,6 +151,8 @@ export default class TimeGrid extends Component {
       resources,
       allDayAccessor,
       showMultiDayTimes,
+      hideGutter,
+      hideTimeIndicator,
     } = this.props
 
     width = width || this.state.gutterWidth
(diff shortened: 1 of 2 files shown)


jquense:
Sorry for the delay :P This looks good can you update and rebase on the current master?

src/TimeGrid.js
```diff
@@ -188,15 +193,17 @@ export default class TimeGrid extends Component {
...
-          <div ref="timeIndicator" className="rbc-current-time-indicator" />
-
-          <TimeColumn
-            {...this.props}
-            showLabels
-            style={{ width }}
-            ref={gutterRef}
-            className="rbc-time-gutter"
-          />
+          {!hideTimeIndicator && <div ref="timeIndicator" className="rbc-current-time-indicator" />}
```
jquense: If we aren't showing the indicator we should also not be calculating it's position on an interval. Can you also disable that when this is true?

paustint:
@jquense - sorry, I have been on vacation for a bit. I am going to close this PR since there has been a lot of other changes since this was first opened and I need to get up to speed on those other changes and will revisit the changes in this PR and probably open another PR with your suggested changes as well.

snehjogani:
i'm unable to hide the time indicator by passing the above mentioned options. 

```
<BigCalendar
            localizer={localizer}
            events={set}
            popup
            step={60}
            hideTimeIndicator
            showMultiDayTimes
            defaultDate={new Date()}
            views={{ month: true, week: true, day: true }}
            startAccessor={'startDate'}
            endAccessor={'endDate'}
            tooltipAccessor={'status'}
            eventPropGetter={event => ({
              style: { backgroundColor: event.colorCode },
            })}
            onSelectEvent={({ id }) => history.push(`Dispatches/${id}`)}
          />
```

Remark if i'm doing something wrong here.

asashay:
would be awesome to have this one in the lib :)
@paustint do you still plan to add it?

iamgbayer:
Some update about this feature?

angela318:
This is not merged?

AndresFSalcedo:
Hi guys,
Any update on this matter?

craftzneko:
@paustint @jquense  was this ever merged assume it was not, has there been later work to do this?, would be very useful for all day events. Have any others found a good solution @AndresFSalcedo @angela318  @iamgbayer @asashay @snehjogani

cutterbl:
@craftzneko I don't think anyone ever updated/recreated the PR with current code, so it never made it into master.

craftzneko:
Is there any hope for this feature to be added, I am using a plugin for grafana that has this as part of their code and hiding the time is critical in my use case as it doesnt make sense having times listed that will confuse users. If this just is not going to happen is there a way i could make it work on my side @paustint dont have much experience with this codebase but i can take instruction if someone with more knowledge than me can explain how to achieve

cutterbl:
@craftzneko Someone would need to recreate the PR with current code

PR #674, opened 2018-01-06
Current timeslot indicator does not update if page is left open #636

### LINKED ISSUE #636: Current timeslot indicator does not update if page is left open
The [time indicator updating code](https://github.com/intljusticemission/react-big-calendar/blob/483534f2798e66c46550af3da7c572624baa63e0/src/TimeGrid.js#L431) manually changes the position of the indicator, but doesn't re-render the component or change the [`now` prop which is set by default at page load time](https://github.com/intljusticemission/react-big-calendar/blob/483534f2798e66c46550af3da7c572624baa63e0/src/TimeGrid.js#L79). This leads to the issue below if the page is left open:

screen shot 2017-12-05 at 3 30 39 pm

My suggestion would be to remove the repositioning code entirely, relying on React to do that instead, put `now` as state instead of as a prop, and simply calling `setState()` in the timeout.

paustint:
If `now` is not passed to `TimeGrid`, then dynamically update `now` upon component changes
If a caller passes in `now` using props, then no automatic changes will occur to state and the caller must manage updates

resolves #636

cc @sciyoshi

@jquense - I went ahead and implement #636 as my project could benefit from this change.  Please let me know if the changes are satisfactory or if you would like me to change anything.

Thank you!

### DIFF
--- a/src/TimeGrid.js
+++ b/src/TimeGrid.js
@@ -78,12 +78,15 @@ export default class TimeGrid extends Component {
      * There is a strange bug in React, using ...TimeColumn.defaultProps causes weird crashes
      */
     type: 'gutter',
-    now: new Date(),
   }
 
   constructor(props) {
     super(props)
-    this.state = { gutterWidth: undefined, isOverflowing: null }
+    this.state = {
+      gutterWidth: undefined,
+      isOverflowing: null,
+      now: props.now || new Date(),
+    }
     this.handleSelectEvent = this.handleSelectEvent.bind(this)
     this.handleDoubleClickEvent = this.handleDoubleClickEvent.bind(this)
     this.handleHeaderClick = this.handleHeaderClick.bind(this)
@@ -114,7 +117,13 @@ export default class TimeGrid extends Component {
     if (this.props.width == null && !this.state.gutterWidth) {
       this.measureGutter()
     }
-
+    if (
+      !this.props.now &&
+      dates.diff(new Date(), this.state.date, 'seconds') > 60
+    ) {
+      // Update now() at most every 60 seconds if changes occur
+      this.setState({ now: new Date() })
+    }
     this.applyScroll()
     this.positionTimeIndicator()
     //this.checkOverflow()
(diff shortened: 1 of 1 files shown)


paustint:
@jquense - looks like there were some 404's on Travis CI - I don't think this was anything I did, as I did not modify dependencies.  **Thoughts?**

src/TimeGrid.js
```diff
@@ -114,7 +117,13 @@ export default class TimeGrid extends Component {
     if (this.props.width == null && !this.state.gutterWidth) {
       this.measureGutter()
     }
-
+    if (
```
jquense: maybe stick this in `componentWillReceiveProps` instead?

paustint: @jquense - I have made the requested change. I have also changed when the state update is made, as I realized that I was not actually solving the problem in my first commit...

State is now updated each time `triggerTimeIndicatorUpdate()` is called.

Also, I noticed another bug, but probably not a big deal.
One one minute intervals, the time Indicator will always be on the timeslot prior to the actual current minute.  I doubt many people would have this many intervals.

image

sciyoshi:
Nice @paustint - note that the 404s might be related to the npm packages that suddenly disappeared: https://github.com/npm/registry/issues/255

src/TimeGrid.js
tobiasandersen: You seem to have used `componentWillMount` instead of `componentWillReceiveProps`.

paustint: @tobiasandersen - take a look at the more recent commit here: https://github.com/intljusticemission/react-big-calendar/pull/674/commits/bec269b8aa74ca6bd3d64aebf2b03968ca19acb8

I changed the strategy a bit because I realized the componentWillReceiveProps wouldn't work.
Details of implementation:
- If user passes in the undocumented prop 'now', that is passed to the `TimeGrid` (via `{...this.props}` on `Calendar` )
    - In this case, we use `now` by setting the `now` prop on the `TimeGrid` state and keep the state up to date  with `componentWillReceiveProps`
- If `now` was not passed in, then we initialize `now` on the state to `new Date()`
- When the `triggerTimeIndicatorUpdate()` is invoked it starts the 60 second recursive timer, this will now update the `now` property on state.

tobiasandersen: Gotcha, I was too quick and only looked through the changed code.

jquense:
I think this is more robustly handled by #707 so gonna close this. Thanks for the PR!

PR #676, opened 2018-01-07
use es6 instead of React.createClass for examples

tobiasandersen:
React 16 was introduced with https://github.com/intljusticemission/react-big-calendar/pull/671 , so we can't use `React.createClass` any more. This means neither stories or examples can be run at this point. I updated all the examples to use es6 instead, but (I think) the `markdown-jsx-loader` still uses `React.createClass`, so that package will also have to be updated before stories and examples work again.

### DIFF
--- a/examples/Api.js
+++ b/examples/Api.js
@@ -14,7 +14,7 @@ let cleanDocletValue = str =>
     .replace(/^\{/, '')
     .replace(/\}$/, '')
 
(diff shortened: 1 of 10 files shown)


jquense:
Oops sorry about that should have been more thorough on that pr

arecvlohe:
This is an issue in #578 as well

Issue #649, opened 2017-12-18
Multiple events at the same time


zackify:
If multiple events are in the same time slot, only the top most one shows, and the others are rendered under it on the day view.

tobiasandersen:
That sounds like a bug, but the examples page seems to be working. Can you post a screenshot, along with the events data?

zackify:
<img width="1021" alt="screen shot 2017-12-18 at 9 43 53 am" src="https://user-images.githubusercontent.com/449136/34111535-ff89d60a-e3d7-11e7-92e0-c0decbf84fee.png">
<img width="909" alt="screen shot 2017-12-18 at 9 43 49 am" src="https://user-images.githubusercontent.com/449136/34111536-ff952578-e3d7-11e7-9ba4-cb12b021658b.png">

I was thinking maybe its a css issue, but the percentages are all the same

zackify:
![image](https://user-images.githubusercontent.com/449136/34111790-bc1b9eca-e3d8-11e7-8c36-94624da71037.png)
here's the data, had to delete sensitive stuff lol

sandysandajan:
I also encountered the same issue that the events overlap when the time slots are the same.  Awaiting for reply from the developers.

jquense:
@zackify can use actually paste the cleaned events that cause the issue so we can copy it here? Or just put something together we can run

zackify:
yeah ill try to make a little CRA demo soon

jquense:
Thanks!

zackify:
1. create-react-app cal
2. npm install react-big-calendar moment
3. Paste code below into App.js
4. npm start

```import React, { Component } from 'react';
import moment from 'moment';
import BigCalendar from 'react-big-calendar';
import 'react-big-calendar/lib/css/react-big-calendar.css';
BigCalendar.momentLocalizer(moment);

class App extends Component {
  render() {
    return (
      <div>
        <BigCalendar
          popup
          events={[
            {
              end: new Date('2017-12-27T21:45:00.000Z'),
              patient_id: 1,
              selectable: true,
              start: new Date('2017-12-27T21:45:00.000Z'),
              title: 'test',
            },
            {
              end: new Date('2017-12-27T21:45:00.000Z'),
              patient_id: 2,
              selectable: true,
              start: new Date('2017-12-27T21:45:00.000Z'),
              title: 'test 2',
            },
            {
              end: new Date('2017-12-27T21:45:00.000Z'),
[...]

PR #677, opened 2018-01-07
New algorithm for day view layout

### LINKED ISSUE #512: Event shown on top of another Event (hiding the one below)
Version 0.12.3 didn't have this issue. I only see this issue in the latest 0.14.4. Notice that we have are giving events 1-6 (with titles 0 to 5). 4 & 5 are displayed on top of each other. If I do display:none on #5 you will see #4 show up. Attaching the start and end times along with order of events array we are passing. I JSON.stringified it so you guys can easily look at it. Seems like an offset calculation issue.

[json.stringify.txt](https://github.com/intljusticemission/react-big-calendar/files/1232684/json.stringify.txt)

screen shot 2017-08-17 at 2 28 42 pm
screen shot 2017-08-17 at 2 28 54 pm
screen shot 2017-08-17 at 2 29 17 pm

tobiasandersen:
This PR replaces the old day view layout algorithm with a new one. There are some problems with the old one that are quite hard to solve (e.g. #512). The render order is also quite messed up. I did my best to understand how Google render events in their (old) calendar. I think I got pretty close (they have since updated Google Calendar with a new layout, along with a new algorithm). 

This new layout should fix #512 and #649. It's also more space efficient, and renders events in a more logical order (no need to keep track of different z-indexes for overlapping events, which was the only solution in the old version). The whole algorithm also makes a lot more sense than the old one, and is easier to follow/read (at least to me). 

I've also introduced a 10px wide empty space column so that new events can always be created on a slot (same way google does it):

empty-space

I'm sure this is far from perfect and still has bugs, but it should be a pretty big improvement to the old one. The master branch is currently a bit broken though, so I haven't tested this with the latest code, so we should probably hold off on merging this until master is fixed and this has been tested. I'd welcome anyone to try it out before we merge though! 

For anyone trying and finding problems, please post the relevant data set here!


jquense:
🎉🎉 awesome! This looks great. What's wrong with master at the moment? (did I forget to fix something??)

src/DayColumn.js
```diff
@@ -129,8 +129,9 @@ class DayColumn extends React.Component {
         max={max}
         step={step}
       >
-        {this.renderEvents()}
-
+        <div className={cn('rbc-events-container', { rtl: this.props.rtl })}>
+          {this.renderEvents()}
+        </div>
```
jquense: We may need to handle a tweak to DnD to account for the new div? maybe not

src/less/time-column.less
```diff
@@ -31,6 +31,19 @@
 .rbc-day-slot {
   position: relative;
 
+  .rbc-events-container {
+    border-right: 10px solid transparent;
+    height: 100%;
+    position: absolute;
+    top: 0;
+    width: 100%;
```
jquense: any particular reason for this set of styles to accomplish this? as opposed to like 

```css
position: absolute;
top: 0, bottom: 0: left: 0; right 10px;
```

i'm not saying the above is better/worse just curious

src/utils/dayViewLayout/event.js
```diff
@@ -0,0 +1,184 @@
...
+export function positionFromDate(date, min, total) {
+  if (startsBefore(date, min)) {
+    return 0
+  }
+
+  const diff = dates.diff(min, dates.merge(min, date), 'minutes')
+  return Math.min(diff, total)
+}
+
+export class Event {
```
jquense: it seems like many of the getters here are essentially `readonly` at object construction? maybe it'd be worth doing some of the calculations then so they aren't run more than once? Does `data` change? Maybe it's over optimization, but it might simplify the compiled code if nothing else

jquense:
Also @taion  if you have second I'd appreciate your expertise, if you could take a quick look on the layout algorithm code w/r/t complexity, efficiency, etc. The scale here is possibly 100s of events but probably mostly <50

src/less/time-column.less
tobiasandersen: Honestly don't remember, I think I added this a few months back. I'll look over it...

src/DayColumn.js
tobiasandersen: Good point, I'll test the DnD stuff before merging.

src/utils/dayViewLayout/event.js
tobiasandersen: That's definitely possible, I haven't really done any sort of optimizations yet...

tobiasandersen: I should probably do this before merging though since it's an easy change 👍

taion:
Well, for dev at least, we'd need to upgrade R-B to work with React 16

src/utils/dayViewLayout/index.js
```diff
@@ -0,0 +1,128 @@
...
+  // Group overlapping events, while keeping order.
+  // Every event is always one of: container, row or leaf.
+  // Containers can contain rows, and rows can contain leaves.
+  const containerEvents = []
+
+  for (let i = 0; i < eventsInRenderOrder.length; i++) {
+    const event = eventsInRenderOrder[i]
+
+    // Check if this event can go into a container event.
+    const container = containerEvents.find(c => c.endSlot >= event.endSlot)
```
taion: this seems like it might be inefficient

taion:
Assuming I didn't mess something up, this doesn't work quite right for overlapping events. For:

```js
    [
      {
        title: 'test',
        start: new Date(y, m, d, 10, 0, 0, 0),
        end: new Date(y, m, d, 11, 0, 0, 0),
      },
      {
        title: 'test 2',
        start: new Date(y, m, d, 10, 30, 0, 0),
        end: new Date(y, m, d, 11, 30, 0, 0),
      },
    ],
```

I get:
image

GCal gives:
image

taion:
The Google algorithm is a bit simpler, as far as I can tell:
image

It goes something like:

1. Sort all events by start time and then end time
2. For each event in order, render it in the first "column" where it would fit
3. After laying everything out, go through each column and increment the X offsets such that event titles don't get occluded too much

Some more examples:
image
image
image
image
image

taion:
As an edge case example, see
image

Note the "***" event is rendered in column 4. A non-greedy algorithm could probably render it "better" (i.e. to the left of its actual parent), but this is pretty straightforward.

taion:
To demonstrate the greedy nature:
image

In this case it would  definitely be better to render the "***" event to the left, but the greedy algorithm isn't cost-sensitive

src/utils/dayViewLayout/index.js
```diff
@@ -0,0 +1,128 @@
...
+
+function sortByRender(events) {
+  // Sort events according to:
+  // 1. start time
+  // 2. duration
+  const sortedByTime = [...events] // clone to keep original order
+    .sort((a, b) => {
+      if (a.start === b.start) {
+        if (a.end === b.end) {
+          return events.indexOf(a) - events.indexOf(b)
```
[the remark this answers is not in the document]
tobiasandersen: Good point!

src/utils/dayViewLayout/index.js
```diff
@@ -0,0 +1,128 @@
...
+        return b.end - a.end
+      }
+
+      return a.start - b.start
+    })
+
+  const sorted = []
+  while (sortedByTime.length > 0) {
+    const [event] = sortedByTime.splice(0, 1)
+    event && sorted.push(event)
```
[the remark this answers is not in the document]
tobiasandersen: It can't, I'll remove that check...

src/utils/dayViewLayout/index.js
```diff
@@ -0,0 +1,128 @@
...
+
+    for (let i = 0; i < sortedByTime.length; i++) {
+      const test = sortedByTime[i]
+
+      // Still inside this event, look for next.
+      if (event.end > test.start) {
+        continue
+      }
+
+      // We've found the first event of the next event group.
```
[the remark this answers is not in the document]
tobiasandersen: Agreed it seems a bit weird, even more so now when their new algorithm doesn't do it like this, but up until a few weeks ago google actually did it this way (and I tried to just copy their behavior).

src/utils/dayViewLayout/index.js
tobiasandersen: Yeah, I'll see if I can simplify it.

src/utils/dayViewLayout/index.js
```diff
@@ -0,0 +1,128 @@
...
+
+        return b.end - a.end
+      }
+
+      return a.start - b.start
+    })
+
+  const sorted = []
+  while (sortedByTime.length > 0) {
+    const [event] = sortedByTime.splice(0, 1)
```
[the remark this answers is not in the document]
tobiasandersen: Haha yes should definitely use `shift()` instead (it didn't always splice from index 0 before).

tobiasandersen:
Hey thanks a lot @taion, this is awesome! 

I've barely looked at how google's new algorithm works, but I'll see if it seems straightforward enough for me to update. 

This test case you provided:

```
    [
      {
        title: 'test',
        start: new Date(y, m, d, 10, 0, 0, 0),
        end: new Date(y, m, d, 11, 0, 0, 0),
      },
      {
        title: 'test 2',
        start: new Date(y, m, d, 10, 30, 0, 0),
        end: new Date(y, m, d, 11, 30, 0, 0),
      },
    ],
```
shouldn't be rendered like that though, so I'll have a look at it later today. 

Again, really appreciate your help!

zackify:
Can't wait for this to be merged, amazing work :)

src/DayColumn.js
tobiasandersen: I only tested some basic DnD in the examples, and it worked as far as I can tell.

src/less/time-column.less
```diff
@@ -31,13 +31,29 @@
...
+      left: 10px;
+      right: 0;
+    }
+  }
+
   .rbc-event {
     border: 1px solid @event-border;
     display: flex;
     max-height: 100%;
+    min-height: 20px;
```
tobiasandersen: Events smaller than this looks really weird. Is it okay to add this here?

tobiasandersen:
I've addressed most of your and @taion's comments. But I kept the same algorithm (where I try to mimic google's old one) - I don't have the time or energy to do it all over again at this point. If anyone else wants to take a stab at the newer algorithm, please do! 

As far as I can tell, this should be good enough to merge now. And as stated above, there're probably (definitely) problems with this one as well, but I'm pretty confident in this being better than what we have now.

src/utils/dayViewLayout/event.js
tobiasandersen: This was a really good call, I even ended up removing the extra MultiDayEvent class.

taion:
I don't understand. The classic Google calendar uses the same layout algorithm:
image

It handles event widths a little differently, and it doesn't "de-offset" things that don't overlap a title, but it follows (1) and (2) in the greedy algorithm I describe in https://github.com/intljusticemission/react-big-calendar/pull/677#issuecomment-355898724 just the same.

tobiasandersen:
Yes their algorithm definitely is better than what I've done here. For this example I think mine renders it ok though:

screen shot 2018-01-11 at 10 35 01

But I'm sure one can find other examples where the differences are more striking and problematic. 

I know you know what you're talking about, so I'm not arguing here, but to me (2) isn't obvious. I'd love to fix/improve it, but I'm not going to have a lot of time the following weeks, and it seems like a pretty fundamental change is required. If anyone wants to pick this up from here, I really don't mind. And if @jquense wants to leave this here until I, or someone else, have the time to get back to it, that's also fine!

jquense:
I think this strictly an improvement over what we have so I say we merge it. For anyone that hits edge cases and may want to improve it this thread is a good resource. Thanks y'all

jadczyk:
Hi,
I ran into an issues with this version: the algorithm considers 2 events as overlapping if one ends when the other starts. I changed the comparison here for my needs:
```
src/utils/dayViewLayout/index.js
    const container = containerEvents.find(c => c.endSlot > event.startSlot)
```
Not sure how you wanted to treat this case.

tobiasandersen:
You're definitely right! Care to make a PR? :)

zackify:
Any plans for a new release on npm soon? Also will this fix events that have the exact same start and end time being not viewable? or only fix ones that are slightly staggered. thanks again :)

tobiasandersen:
That should be fixed here! About the release, I personally feel like there're a few minor issues left to fix on master before we're ready for a release. E.g:
- https://github.com/intljusticemission/react-big-calendar/pull/677#issuecomment-357525261
- https://github.com/intljusticemission/react-big-calendar/pull/690#issuecomment-357544038
- (possibly) https://github.com/intljusticemission/react-big-calendar/issues/688

But once those are fixed, I think we should publish a new one.

PR #681, opened 2018-01-08
Update react-related dependencies

AlexMarvelo:
In #671 React was updated to v16. But the app is crashing after that. We have 2 ways:
1) Revert v16 update
2) Fix crashes. This PR is about it

`uncontorllable` and `markdown-jsx-loader` should be updated, as they use React.createClass in v3

But we have another problem - `markdown-jsx-loader` v3.0.0 is braking. It failes webpack build with `BabelLoaderError: SyntaxError: Unexpected token, expected ;` error (even for empty markdown file!). @jquense look into that. It's critical

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,7 +105,7 @@
     "prop-types": "^15.5.8",
     "react-overlays": "^0.7.0",
     "react-prop-types": "^0.4.0",
-    "uncontrollable": "^3.3.1 || ^4.0.0",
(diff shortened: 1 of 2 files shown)


tobiasandersen:
I had to patch the `markdown-jsx-loader` to get it working. But the examples should run on master again now. Closing this, but thanks for the PR!

PR #683, opened 2018-01-09
update broken dependencies

tobiasandersen:
This should get the examples running properly again.

### DIFF
--- a/examples/App.js
+++ b/examples/App.js
@@ -124,7 +124,9 @@ class Example extends React.Component {
           </div>
         </div>
         <div className="docs">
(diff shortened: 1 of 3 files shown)


examples/App.js
```diff
@@ -124,7 +124,9 @@ class Example extends React.Component {
           </div>
         </div>
         <div className="docs">
-          <Intro className="contain section" />
+          <div className="contain section">
```
tobiasandersen: `className` was dropped `markdown-jsx-loader`, so had to do this...

PR #684, opened 2018-01-09
fix inRange check

tobiasandersen:
Currently no events are rendered in the day view, this should fix that.


tobiasandersen:
Guess I was a bit too quick here, since some tests are failing...

PR #690, opened 2018-01-14
fixed non selectable events on weekly view

jadczyk:
After some recent changes, events became unselectable on weekly view if not using drag and drop.

### DIFF
--- a/src/less/event.less
+++ b/src/less/event.less
@@ -6,6 +6,7 @@
   background-color: @event-bg;
   border-radius: @event-border-radius;
   color: @event-color;
+  pointer-events: all;
 
   &.rbc-selected {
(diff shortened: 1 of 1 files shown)


arecvlohe:
Hopefully this is all that needs to be changed. Need to have more testing around this.

jquense:
Let's get two contributor approvals for PRs please  :)

I'm not sure this the best fix. We really want the standard calendar to work without needing pointer-events css. If this is only needed for resizing then it should probably go in the addons style

arecvlohe:
@jquense My mistake! I will create a PR to fix this PR :)

tobiasandersen:
@arecvlohe @jquense is this considered resolved?

jquense:
Yup :thumbsup:

Issue #470, opened 2017-07-13
Change event-cell hoover behavior


gryzik:
I need to change event hoover behavior.
I want to remove existing tooltip and add my hoover function.
Is it possible ?

GhislainRoy04:
I manage to do something using eventPropGetter, I passed a component that had hover events on it and I displayed the info there. Not sure on how to remove the exiting tooltip tho

skhamoud:
Any updates on this ? I can't find a way to neither hide nor change the behavior of the tooltip .

jquense:
No updates. PR's welcome!

aappddeevv:
I think I responded to  a closed issue, but we need to create our own event renderings. Something I have seen that is useful in other libraries is a `onEventRender` type property that takes the properties and default rendering function. Then someone can just use the default rendering and wrap the default rendered component easily with something that does tooltip/hover behavior with your own hover content of course. 

onEventRender would get a default rendering function that renders an EventCell object  so you can really do what you want easily. This also fits into the renderprops pattern that is popular.

PR #701, opened 2018-01-23
Add tooltipAccessor prop to calendar to disable tooltip if needed

### LINKED ISSUE #226: Event title tooltip should stringify
It would appear that in case where we want to stylize (bold, italic, etc) some parts of an event title, it shows up properly on the page but the tooltip contains `[object Object]` as values.

A solution would be to use the `jsx-to-string` library to stringify the value or better yet add a tooltipAccessor method.

ginifizz:
Use a "tooltipAccessor" prop for the tooltip instead of the "titleAccessor" prop. Can show a tooltip different from the title, or not display it at all if the accessor returns null.

Fixes #226, #470

### DIFF
--- a/src/DayColumn.js
+++ b/src/DayColumn.js
@@ -233,12 +233,11 @@ class DayColumn extends React.Component {
               [isRtl ? 'right' : 'left']: `${Math.max(0, xOffset)}%`,
               width: `${width}%`,
             }}
-            {...(tooltip
-              ? {
-                  title:
-                    (typeof label === 'string' ? label + ': ' : '') + tooltip,
-                }
-              : {})}
+            title={
+              tooltip
+                ? (typeof label === 'string' ? label + ': ' : '') + tooltip
+                : undefined
+            }
             onClick={e => this._select(event, e)}
             onDoubleClick={e => this._doubleClick(event, e)}
             className={cn('rbc-event', className, {
--- a/src/DayColumn.js
+++ b/src/DayColumn.js
@@ -39,6 +39,7 @@ class DayColumn extends React.Component {
 
     rtl: PropTypes.bool,
     titleAccessor: accessor,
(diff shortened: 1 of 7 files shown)


wlalele:
:+1:

src/DayColumn.js
```diff
@@ -230,7 +233,12 @@ class DayColumn extends React.Component {
               [isRtl ? 'right' : 'left']: `${Math.max(0, xOffset)}%`,
               width: `${width}%`,
             }}
-            title={(typeof label === 'string' ? label + ': ' : '') + title}
+            {...(tooltip
+              ? {
+                  title:
+                    (typeof label === 'string' ? label + ': ' : '') + tooltip,
+                }
+              : {})}
```
jquense: lets make this a bit less ugly please :P i'd just set title to `undefined` in the one case that will be fine

src/EventCell.js
```diff
@@ -75,7 +78,7 @@ class EventCell extends React.Component {
           onClick={e => onSelect(event, e)}
           onDoubleClick={e => onDoubleClick(event, e)}
         >
-          <div className="rbc-event-content" title={title}>
+          <div className="rbc-event-content" {...tooltip ? { title: tooltip } : {}}>
```
jquense: ditto here

jquense:
thanks!

RusselFourteen:
Can you please tell me how to off the built it Tooltip? I already tried many times but I still failed. Thanks and Sorry, newbie here

RusselFourteen:
I already added toolTipAccessor but what is my next move? Where can I defined tooltip as undefined or null?

ginifizz:
It works like the others accessors. Assign a function who returns NULL to the tooltipAccessor and it will be removed.
Like this : 
`<Calendar tooltipAccessor={() => null} ... />`

PR #703, opened 2018-01-24
Update display of time indicator

alexrass:
Update time indicator to only show the bar on the current day instead of the full week.
This behavior is more consistent with gcal and is less confusing when looking at the calendar

screen shot 2018-01-24 at 4 08 34 pm

screen shot 2018-01-24 at 4 08 40 pm

### DIFF
--- a/examples/events.js
+++ b/examples/events.js
@@ -91,4 +91,10 @@ export default [
     start: new Date(2015, 3, 20, 19, 30, 0),
     end: new Date(2015, 3, 22, 2, 0, 0),
   },
+  {
+    id: 14,
+    title: 'Today',
+    start: new Date(new Date().setHours(new Date().getHours() - 3)),
+    end: new Date(new Date().setHours(new Date().getHours() + 3)),
+  },
 ]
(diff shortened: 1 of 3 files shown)


tobiasandersen:
I also like this behavior more! But couldn't we adjust the line that's already there rather than creating another one? To me it looks a bit off having that circle in the gutter all the time, especially if "today" is not in range and I can't see the line...

alexrass:
That works as well -- I was thinking it would be nice to know what time it is even if you aren't on the current day, but if you want something closer to what gcal does, I'm happy to update it.  Let me know what your preference is and I will update.

tobiasandersen:
Ah I see 👍  It's just that when I looked at the demo page I thought something was wrong. But that could also just be that I was expecting something else. I'd prefer what Google does, but it's up to @jquense in the end.

jquense:
I like this, i'd just remove the little circle, its superfluous for a default anyway

longsleep:
Mhm this change is problematic since now it uses javascript to compute a width. This width does not change until the component is rerendered (or every minute). This creates issues when the size of the component changes because the container in it changes width (e.g. on window resize).

longsleep:
There is also a issue for mis-alignment of the time indicator at #836 - will continue discussion there.

PR #705, opened 2018-01-29
Added an example for Resource view

akeroyd:
I've added an example using the Resource settings for a multi-column daily view.


jquense:
thanks

Issue #704, opened 2018-01-24
feature request: nowAccessor (or equiv)


ehahn9:
The `timeIndicator` is always placed by `positionTimeIndicator()` at the time corresponding to `new Date()` - e.g. the browser's local time. 99% of the time this is perfect.

... but if you're displaying the calendar in another time zone (e.g. using any of the great ideas from #118), then you also need to adjust (e.g. "fake out") the notion of 'now' to match.

tl;dr - a `nowAccessor` would be lovely.

If I was more confident in my open source manners, I'd submit a PR, but alas, I'm a newb. :-(

ehahn9:
Also would impact the determination of `rbc-today` styling.

... and `nowAccessor` wouldn't really be an accessor insofar as it requires a fn(), and can't take a sting (there's no named property option to reference). So, to keep consistent with the api, maybe `getNow` or `nowGetter` is a more accurate name?

jquense:
Sorry closed accidentally!

ehahn9:
Made some progress on this - might issue a PR shortly (my first ever!). Please accept my apologies if I'm doing any of this incorrectly... Would love some feedback if that's not unreasonable...

I named the property `now`. It can be either a date or (more usefully) a fn which returns a date, since that way the time-indicator bar will correctly poll for changes as it currently does. The default prop value is therefore `() => new Date()`. I created a new `PropType.dateGetter` (which is either `instanceOf(Date)` or `func`). `now` is the only prop using it, tho.

The behavior affects the `.rbc-today` shading, the `.rbc-time-indicator` horizontal line and the behavior of `navigate.TODAY` - all of which get the notion of now (or today) from the new prop.

Turns out it also impacted the calculations for `min`, `max` and `scrollToTime` which were all hardcoded to `new Date()` (they use the new prop, too).

Finally, I changed the `rendering` demo to illustrate a customized `now` fn, but I suspect that's not worth it since it's sort of a trivial api change. Thoughts?

Worth issuing a PR or am I the only person likely to ever use this? (in fairness, I think anyone wanting to display calendars in different time zones will need something like this?).

tobiasandersen:
Hey @ehahn9! This sounds great! 

About the name, I actually think I like your original naming better (i.e. `nowAccessor`). It's more consistent with the rest of the api. Having said that, I think it makes sense if that prop only accepts functions. I don't really see the value of passing just a date, but maybe I'm missing something? It sounds like something that people will do by mistake quite easily.

[...]

PR #707, opened 2018-01-30
nowAccessor to allow control over "current" date/time (#704)

ehahn9:
New, optional, `nowAccessor` prop which accepts a function to return the current (now) time.

This result is used to indicate which days/times are highlighted (.rbc-today and .rbc-time-indicator). It also informs navigation to "today" via `nagivate.TODAY`. Finally, it does impact the `min`, `max` and `scrollToTime`.

The `rendering` example shows usage and a new storyboard is included.


src/Month.js
```diff
@@ -32,7 +32,7 @@ let propTypes = {
   max: PropTypes.instanceOf(Date),
 
   step: PropTypes.number,
-  now: PropTypes.instanceOf(Date),
+  nowAccessor: PropTypes.func.isRequired,
```
jquense: lets move the default up to Calendar, so it doesn't need to be set a bunch of places (like TimeGrid)

src/BackgroundCells.js
```diff
@@ -62,8 +64,10 @@ class BackgroundCells extends React.Component {
       cellWrapperComponent: Wrapper,
       dayPropGetter,
       date: currentDate,
+      nowAccessor,
     } = this.props
     let { selecting, startIdx, endIdx } = this.state
+    let current = get(nowAccessor)
```
jquense: the use of `get` is a bit much for this, since it's a required function, `nowAccesser()` is fine

src/BackgroundCells.js
```diff
@@ -18,6 +19,7 @@ import Selection, { getBoundsForNode, isEvent } from './Selection'
 class BackgroundCells extends React.Component {
   static propTypes = {
     date: PropTypes.instanceOf(Date),
+    nowAccessor: PropTypes.func.isRequired,
```
jquense: i missed a bit of the naming discussion but I think `accessor` is misleading since it's not, its only a function that returns a date. `getNow` sounds a bit more accurate to my ears

src/Calendar.js
```diff
@@ -696,10 +719,19 @@ class Calendar extends React.Component {
...
+      max,
+      scrollToTime,
       length,
       ...props
     } = this.props
 
+    current = current || get(nowAccessor)
+    min = min || dates.startOf(current, 'day')
+    max = max || dates.endOf(current, 'day')
+    scrollToTime = scrollToTime || dates.endOf(current, 'day')
```
jquense: this all seems like logic change? what is the rational for this, I think it makes more sense to leave these empty.

I'm also not sure defaulting to scrolling to the end of a day is a good choice...i wouldn't expect that behavior

src/utils/accessors.js
```diff
@@ -9,6 +17,7 @@ export function accessor(data, field) {
     field in data
   )
     value = data[field]
+  else if (typeof data === 'function' && field === undefined) value = data()
```
jquense: I'd remove this and just call the function directly (see my comment above)

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "react-big-calendar",
+  "name": "ehahn9-react-big-calendar",
```
jquense: fix

examples/demos/rendering.js
```diff
@@ -20,6 +20,14 @@ function EventAgenda({ event }) {
   )
 }
 
+const customNow = () => {
```
jquense: i don't think we need this addition to the example here?

src/utils/dates.js
```diff
@@ -80,8 +80,8 @@ let dates = {
     return dates.eq(dateA, dateB, 'month')
   },
 
-  isToday(date) {
-    return dates.eq(date, dates.today(), 'day')
+  sameDay(dateA, dateB) {
+    return dates.eq(dateA, dateB, 'day')
```
jquense: i think maybe this is doesn't add a ton of value over using `dates.eq` each time, i'd remove the helper personally

examples/demos/rendering.js
ehahn9: agreed! esp since there's a storyboard. Will drop.

package.json
ehahn9: yeah, my inexperience showing... had to npm publish this branch but will revert the name in the next attempt.

src/BackgroundCells.js
ehahn9: Especially since we're dropping the `get()` semantics. Will rename to `getNow()`

package.json
jquense: you shouldn't need to do anything with npm here :) whatever is on npm is unrelated to what's in the PR here. you can commit a change to your branch and push and it'll update without needing to publish anything

src/Calendar.js
ehahn9: Rationale: the current behavior has `min`, `max` and `scrollToTime` set as `defaultProps` to `new Date()`. I think they all do need to be based on the `getNow()` time. My change of `scrollToTime` is just plain **wrong** - it should have been `startOf(current, 'day')`.

I was just mirroring (well, attempting to :-)) the current behavior using the new accessor instead of the hardwired `new Date()` in the default props.

Having said all of that, I can see an argument for leaving all of this alone - let me give it some more thought so I don't waste anyone's time.

src/Month.js
ehahn9: Sounds right. I was a bit concerned about backward-compatibility with someone using a naked `<TimeGrid>` etc. but that seems YAGNI-ish.

src/utils/dates.js
ehahn9: was copying the `sameMonth` example just above. I agree - will remove. I thought it wise to remove `isToday` (no longer used) since it encourages bad (future) behavior. Agree?

src/Month.js
jquense: yeah, i appreciate you making the most minimal change, though now is a good time to clean up my messy defaults :P

src/utils/dates.js
jquense: 👍

ehahn9:
hope that does it. Thanks for the great feedback and thanks to you & the community for building and enhancing this terrific component! I'm *so* happy to retire `fullCalendar` (and `jQuery`)!

PR #714, opened 2018-02-05
Small stuff

jquense:


jquense:
I tried to check that it all still worked, seemed to not cause issues :thumbsup:

PR #715, opened 2018-02-05
DnD addon: support for custom event components

ehahn9:
I've been trying to use the drag and drop addon with a custom `Event` component. Alas, the addon needs to use its own `ResizeEvent` component and thus overwrites mine.

I have no idea if this PR is useful to anyone else, but I thought I'd submit it just in case in inspires someone more experienced than me to work on this further :-)

This PR changes the behavior to use a HOC (`withEventResizability()`) to compose a new component which effectively wraps any caller-supplied `Event` component. (To mimic the base `Calendar` behavior it wraps a new `TitleComponent` as the default).

I also tried to DRY up the original `ResizeEvent` and `ResizeMonthlyEvent` into a single wrapper which takes a `direction` (horizontal or vertical) to determine where the anchors go, etc.

And I changed a bunch of other gratuitous stuff like adding more docs, using static class props, etc. I also added two new storybooks: one showing drag+resize, the other drag+resize with a custom event component.

Feel free to discard/close this if it isn't useful, but also feel free to put me to work on edits/updates - I'll give it my best shot, time-permitting.

(ps: I know addons aren't really part of react-big-calendar - apologies if I've put this issue in the wrong place).


ehahn9:
apologies - closing this PR and will re-issue once I correct my newb probs with two outstanding PR's.

PR #717, opened 2018-02-06
fix(dayViewLayout): container event check

tobiasandersen:
In response to https://github.com/intljusticemission/react-big-calendar/pull/677#issuecomment-357525261. I think this is what you'd expect...


tobiasandersen:
I'd like to get this in the next release so I'll just merge it right away :)

Issue #656, opened 2017-12-27
Droping over itself, does not fire trigger onDrop. 


sergioshev:
Hello guys.
   We are using your calendar in our project and running in an issue. If I drag and drop an event over itself onDrop trigger is not fired. I tried adding z-index: -1 in react-big-calendar/lib/addons/dragAndDrop/styles.css with no luck. In all other cases the calendar is working ok.

Could you helpme figure out what is gooing on?

tobiasandersen:
It seems to be working for me on the example site, could you show the problem using e.g. http://recordit.co/ or in a [codesandbox](https://codesandbox.io/s/l9jwl0kj6m)?

arecvlohe:
I know what this issues is:

<a href="https://cl.ly/203w1n2J2U3n" target="_blank"><img src="https://dha4w82d62smt.cloudfront.net/items/0O0G3a2V3k0g1T3Q2c1l/Screen%20Recording%202018-01-12%20at%2005.23%20PM.gif" style="display: block;height: auto;width: 100%;"/></a>

jquense:
what's the status on this one?

arecvlohe:
I think there is a somewhat simple way to handle this on resize. I will try and take a stab at it this weekend.

arecvlohe:
Deployed in #724

PR #724, opened 2018-02-12
Fix dnd style preventing event from dropping on itself

arecvlohe:
## Description

This PR is to fix the issue related to not being able to drop an event onto itself when resizing an event to be smaller than its current length. I removed the related styling that was preventing this and the app now works as expected.

## Screentshot

<a href="https://cl.ly/3Z3c0I0N1225" target="_blank"></a>

## Issue
#656

### DIFF
--- a/src/addons/dragAndDrop/styles.less
+++ b/src/addons/dragAndDrop/styles.less
@@ -5,7 +5,7 @@
     pointer-events: none;
 
     & .rbc-show-more,
-    & .rbc-event:not(.rbc-addons-dnd-dragging) {
+    & .rbc-event {
       pointer-events: all;
     }
   }
@@ -32,11 +32,6 @@
     pointer-events: none;
     opacity: .50;
   }
-
-  // .rbc-addons-dnd-dragging {
-  //   pointer-events: none;
-  //   opacity: .35;
-  // }
 
   .rbc-addons-dnd-resize-anchor {
     width: 100%;


src/addons/dragAndDrop/styles.less
```diff
@@ -5,7 +5,7 @@
     pointer-events: none;
 
     & .rbc-show-more,
-    & .rbc-event:not(.rbc-addons-dnd-dragging) {
+    & .rbc-event {
       pointer-events: all;
```
jquense: i'm not sure how this would work...It seems like this would prevent dropping the event at all because drop target is always obscured by the event being dragged, preventing it from detecting `isOver`

arecvlohe: It's not seen by this commit but there is another style that allows this to work on line 31.

```
  &.rbc-addons-dnd-is-dragging .rbc-event {
    pointer-events: none;
    opacity: .50;
}
```

Few things going on here. I think the dragging style is now called `rbc-addons-dnd-is-dragging` instead of `rbc-addons-dnd-dragging`. Second, `rbc-event` does not get the the `rbc-addons-dnd-is-dragging` class. That class in added to the top level div along with `rbc-calendar rbc-addons-dnd`. So I think line 31 handles things correctly.

jquense: 👍 great LGTM

arecvlohe:
Oh, you mean it's approved. I thought there needed to be two approvals? @jquense

jquense:
two contributors should sign off, but you count as one ;) so only one other needed in this case.

PR #753, opened 2018-03-04
use recompose, use unix timestamps

bkniffler:
Hey, I'm using react-big-calendar and I'm seeing performance hits when using many events and/or many resources.

I noticed that for example a change to the events array will cause a lot of components to update, even though their props don't really change (e.g. TimeSlot); or any prop change will cause all events to recalculate and rerender. This is caused by missing shouldComponentUpdate/PureComponent as well as declaration of arrays and dates inside render functions and handing those to child components as props.

There is a few things that could be optimized:
- Outsource expensive calculations out of render (use withPropsOnChange from recompose?) to avoid unnecessary recalculations and unnecessary rerendering of child components
- Avoid date types (use unix timestamps +new Date() since they are much easier to compare)
- Use PureComponent
- Small stuff like avoiding inline functions, object declarations and the likes in render functions

This PR introduces recompose withPropsOnChange, unix date and other small optimizations, but it is only a start. There is many other places that would benefit, but will need to be revisited.

I'd like to use this as a discussion start, so let me know what you think of this.

I've published this packaged as `react-big-calendar-opt`, so feel free to try it, especially with 4x throttling enabled in chrome devtools.

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,6 +106,7 @@
     "prop-types": "^15.5.8",
     "react-overlays": "^0.7.0",
     "react-prop-types": "^0.4.0",
+    "recompose": "^0.26.0",
     "uncontrollable": "^4.0.0",
     "warning": "^2.0.0"
   }
--- a/src/DayColumn.js
+++ b/src/DayColumn.js
@@ -18,6 +18,7 @@ import getStyledEvents, {
 } from './utils/dayViewLayout'
 
 import TimeColumn from './TimeColumn'
+import { compose, withPropsOnChange } from 'recompose'
 
 function snapToSlot(date, step) {
   var roundTo = 1000 * 60 * step
@@ -36,6 +37,7 @@ class DayColumn extends React.Component {
     min: PropTypes.instanceOf(Date).isRequired,
     max: PropTypes.instanceOf(Date).isRequired,
     getNow: PropTypes.func.isRequired,
+    totalMin: PropTypes.number.isRequired,
 
     rtl: PropTypes.bool,
     titleAccessor: accessor,
@@ -106,7 +108,6 @@ class DayColumn extends React.Component {
       ...props
     } = this.props
 
-    this._totalMin = dates.diff(min, max, 'minutes')
     let { selecting, startSlot, endSlot } = this.state
     let slotStyle = this._slotStyle(startSlot, endSlot)
 
@@ -168,6 +169,7 @@ class DayColumn extends React.Component {
       timeslots,
       titleAccessor,
       tooltipAccessor,
+      totalMin,
     } = this.props
 
     let styledEvents = getStyledEvents({
@@ -176,7 +178,7 @@ class DayColumn extends React.Component {
       endAccessor,
       min,
       showMultiDayTimes,
-      totalMin: this._totalMin,
+      totalMin,
       step,
       timeslots,
     })
@@ -264,8 +266,8 @@ class DayColumn extends React.Component {
   }
 
   _slotStyle = (startSlot, endSlot) => {
-    let top = startSlot / this._totalMin * 100
-    let bottom = endSlot / this._totalMin * 100
+    let top = startSlot / this.props.totalMin * 100
+    let bottom = endSlot / this.props.totalMin * 100
 
     return {
       top: top + '%',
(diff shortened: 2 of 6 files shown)


jquense:
Thanks for taking the time to put this together. I can't look at it in detail right now but it's appreciated! Honestly I'm always a bit anxious about performance optimizations generally since they often include a bunch of hidden costs but it's always good to explore! Are you noticing any significant improvements or performance with this? Is it solving a specific concern your having? It's helpful to get baseline for starting the discussion :)

bkniffler:
Hey, yes, it's making a big difference as soon as you have many cells rendered. Here you see few events, lots of cells; after I click on save, one event is moved to another resource:

Before the changes:
before

Around 6 seconds UI freeze with console.log on rerender of components (`why-did-you-update`), 2.5 seconds freeze without logs.

After 
after

Around than 1.5 seconds freeze with console.log, no noticeable freeze without the logs.

I can feel the anxiousness with perf optimizations, especially when implementing shouldComponentUpdate. I found recompose withPropsOnChange very useful since it serves more as a memoization for prop calculations instead of limiting component rerender to specific changes. For example, the calendar gets events but needs to separate allday events from non-allday and filter visible events. Currently this is done inside the render cycle of `TimeGrid` component. Instead, you could have this:

```jsx
const enhance = compose(
  withPropsOnChange(
    [
      'events',
      'range',
      'startAccessor',
      'endAccessor',
      'allDayAccessor',
      'showMultiDayTimes',
    ],
    ({
      events,
      range,
      startAccessor,
      endAccessor,
      allDayAccessor,
      showMultiDayTimes,
    }) => {
      let start = range[0],
        end = range[range.length - 1]

      let allDayEvents = [],
        rangeEvents = []

      events.forEach(event => {
        if (inRange(event, start, end, { startAccessor, endAccessor })) {
          let eStart = get(event, startAccessor),
            eEnd = get(event, endAccessor)

          if (
            get(event, allDayAccessor) ||
            (dates.isJustDate(eStart) && dates.isJustDate(eEnd)) ||
            (!showMultiDayTimes && !dates.eq(eStart, eEnd, 'day'))
          ) {
            allDayEvents.push(event)
          } else {
            rangeEvents.push(event)
          }
        }
      })

      allDayEvents.sort((a, b) =>
        sortEvents(a, b, { startAccessor, endAccessor, allDayAccessor })
      )
      return { allDayEvents, rangeEvents }
    }
  )
)
export default enhance(TimeGrid)
```

There is nothing to loose really, the only downside is the added recompose dependency, but we can always cherry pick the actually used functions instead of importing the whole thing (recompose/withPropsOnChange), so the size difference shouldn't really be noticeable.

I think it would be beneficial to have a stress test in storybooks, so we can actually measure performance differences.

jquense:
> There is nothing to loose really, the only downside is the added recompose dependency

Well and knowledge of how recompose works, which is not something i know, so there is a lot of addition mental overhead for me having to learn a new library (which may not be a bad tradeoff)

I'm still unsure tho what the benefit is practically apart from the metric "the components don't update as much", that isn't necessarily a problem if the "slow updates" are still to fast to notice a delay. I'm not trying to be difficult here, but i want to make sure we are solving a real bottleneck that folks experience.

in terms of the actual changes, it seems like `withPropsOnChange` isn't much more than a props -> state reducer in CWRP? It might be more accessible for maintainers/contributors to do it the old skool way instead of bringing in a new library (maybe not).

I would also check, btw with changes i've got staged in https://github.com/intljusticemission/react-big-calendar/pull/733 which does add some memoization to the TimeGrid component in places I think suffer the most from over re-calculation. We'd want to make srue the two PR here can work together nicely!

Thanks again for taking the effort here, i'm glad to have more time spent on the perf concerns!

arecvlohe:
@bkniffler I don't think you _need_ recompose if you are just going to use 1 utility function. Since `withPropsOnChange` is just a higher order component, there is no reason not to either copy the implementation or create your own for this library. After looking at the function myself, it simply does a shallow check of `props` and `nextProps` inside `componentWillReceiveProps`. I think this is certainly doable and acceptable to add into the project I would think. I haven't experienced the same performance issues as you have but maybe with some revisions and alignment with other open PRs this might become a suitable PR itself.

porfirioribeiro:
Personally i am not a big fan also of using lot's of hoc's in all components, but i'm also concern about performance.

Has i had to implement my own agenda view i also took performance as a big point so i end up using [getDerivedStateFromProps](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops)
I do all my calculations there and store it on state, and only recalculate it again when the needed props change.

WebspressoCo:
I'm having similar issues for calendars with 5 minute slots and more than 5 resources. There is a significant delay on every select / resize / drag & drop action which only gets worse when more resources are added. 
Recompose is certainly an option, however, the author mentioned that the library will no longer be maintained, as the functionality is present in React hooks. https://github.com/acdlite/recompose

@jquense 
Would you be open to a pull request addressing the re-rendering of slots without the use of external libraries? I'll probably have time to work on this after about 2 weeks.

PR #754, opened 2018-03-05
Add title to headers

vongdarakia:
For the resource view, sometimes the label is too small for the header cell and cuts off. It's impossible to tell what the header is when the labels have the same prefixes.

Example header

- Guitar Private Lesson Level I
- Guitar Private Lesson Level II
- Guitar Private Lesson Level III

Those get cut off like this

- Guitar Private Les...
- Guitar Private Les...
- Guitar Private Les...

So you don't know what level they are. Having a title for the resources so that you can hover over and see the full label helps solve this issue. I also added a title to the date header for consistency.

### DIFF
--- a/src/Header.js
+++ b/src/Header.js
@@ -2,7 +2,7 @@ import PropTypes from 'prop-types'
 import React from 'react'
 
 const Header = ({ label }) => {
-  return <span>{label}</span>
+  return <span title={label}>{label}</span>
 }
 
 Header.propTypes = {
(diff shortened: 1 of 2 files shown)


src/Header.js
```diff
@@ -2,7 +2,7 @@ import PropTypes from 'prop-types'
 import React from 'react'
 
 const Header = ({ label }) => {
```
arecvlohe: I am confused. Is this not handled with a custom header component?

vongdarakia: Thanks, now I feel like an idiot. Can we pretend this PR never existed? 😅

arecvlohe: Move along, nothing to see here :)

vongdarakia: @arecvlohe So, you were right about the label, but I just realized now that wasn't for resource headers. It was for the date headers.

I can create another PR so that resource headers can use a customer header.

PR #769, opened 2018-03-19
Add box and bound to onSelectSlot callback

gregeinfrank:
I want to render an in-line popup at the `{x,y}` coordinates of the selection event, so passing back the `box` (for clicks) and the `bounds` (for select) makes this possible.


gregeinfrank:
@jquense I added the doc comment -- I'm not crazy about having a different `bounds` vs `box` for click events vs select events, but the data is different (at least the way it's already written) I'm open to other ideas but at least it's documented properly as is now.

jquense:
Thanks! Yeah the type difference isn't great but it's ok for now.

PR #778, opened 2018-03-26
fix dates collisions

witkowsky:
There is an issue where If you have events where one ends when another is started, you will get different containers. 
Example (in minutes):
```javascript
[
    {start: 540, end: 550},
    {start: 550, end: 560},
    {start: 560, end: 570},
]
```
Right now. They are 3 different event containers.
After the fix, They are nested.
```javascript
[
    {
        start: 540,
        end: 550,
        rows: [
            {
                start: 550,
                end: 560,
                leaves: [
                    {
                        start: 570,
                        end: 580
                    }
                ]
            }
    },
]
```
Just like in google calendar.


tobiasandersen:
Hi, thanks for the PR. Maybe I misunderstood you, but to me it looks like Google Calendar treats them as different containers?

screen shot 2018-03-26 at 12 33 48

witkowsky:
googlecalendar
I checked that. You are right with those dates, but If event ends when another start AND this event is under 30minutes They would be nested.

witkowsky:
You are right with those dates. But I checked some possibilities. As far as I understand.

If a difference between eventA.start and eventB.start is under 30 minutes They are a container with row. If We have also with eventC.start difference between eventA.start under 30 minutes it can become leaf.

tested

Check events at 7pm

tobiasandersen:
Ah, I see! I don't mind a change as long as the current examples still work.

witkowsky:
@tobiasandersen
I understand You, but please consider some changes, because If you have 3 short events (for example 10 minutes) one after another They are quite hard to read.

witkowsky:
I change the code. For me, it behaves like in Google Calendar. Please check that. I only tested that manually. Did not write tests.

witkowsky:
There was the new release at 23 Mar 2018 which change code structure so I moved my change to the new place. 

Examples:
Google Calendar:
google_calendar

v0.19.0 before fix
v19_before_fix

v0.19.0 after fix
v19_fixed

tobiasandersen:
Looking great! 🎉

PR #786, opened 2018-03-29
Add resource id to drag and drop event

kidustiliksew:
`{
  allDay,
  end,
  event,
  start,
 resourceId 
}`

So that we could change the resource of an event as well


tobiasandersen:
Hi! There's quite a lot going on here, with lots of changes and (old?) commits, so it's a bit hard to review. You think you could clean it up a bit? :)

Issue #788, opened 2018-03-29
Event sorting inside of a day


hansena:
#### Do you want to request a _feature_ or report a _bug_?
Feature
#### What's the current behavior?
Events are not always ordered correctly inside of a day
#### What's the expected behavior?
Events are sorted by start time inside of a day

Example: https://codesandbox.io/s/yp1jz5xz7x
![screen shot 2018-03-29 at 5 53 10 pm](https://user-images.githubusercontent.com/15346521/38115620-af6c4cb4-337a-11e8-9ec0-6642f038b08e.png)

hansena:
This was sorted out by using a function in the accessor to convert the incoming string to a date object:
startAccessor={((e)=>new Date(e.start))}

PR #789, opened 2018-03-29
Support sorting events more accurately

hansena:
Pull request for issue: #788


hansena:
using a function as an accessor resulted in proper sorting without this code

PR #798, opened 2018-04-08
Add box arg to _selectSlot on BackgroundCells.js

gregeinfrank:
I think I missed this in my first PR, so some click events are not passing through the `box` properly


gregeinfrank:
@jquense thanks! Do you think you can push a new release soon with this bug fix?

PR #808, opened 2018-04-17
Enable support per-event dragability/sizability

dbuchet-bandc:


src/addons/dragAndDrop/DraggableEventWrapper.js
```diff
@@ -135,7 +135,7 @@ class DraggableEventWrapper extends React.Component {
 /* drag sources */
 const makeEventSource = anchor => ({
   beginDrag: ({ event }) => ({ event, anchor }),
-  // canDrag: ({ event }) => event.draggable === undefined || event.draggable - e.g. support per-event dragability/sizability
+  canDrag: ({ event }) => event.draggable === undefined || event.draggable // support per-event dragability/sizability
```
jquense: I think i'd prefer to avoid another special property on the event objects, can we use a function prop that is passed the event and returns true/false. thoughts?

arecvlohe: I kind of agree with this. A function that takes the event and determines if it can be dragged I think is a better option. I suppose call this `draggableAccessor`?

dbuchet-bandc: Hi there,
I just moved quickly as I saw the code was already implemented.
An accessor would be a good idea but I'm afraid we'll be lost with DnD context.
https://react-dnd.github.io/react-dnd/docs-drag-source.html

For me it's easier to keep the `draggable` key in the event object for now :)

jquense: the props to the component are passed in here, i don't think you need to mess with context at all, we just get the function the same way we get the event. Also since it's an accessor like the other sorts (it's just a function not a string or func) lets call it `isDraggable` sort of like `getNow`

arecvlohe: @dbuchet-bandc Yes, it's just a function that is passed to the calendar so context is not an issue, just passing it down from props should suffice.

dbuchet-bandc:
Ok sorry, misunderstood what you asked for.
What about this update? :)

arecvlohe:
@dbuchet-bandc Nice work! What do you think @jquense?

src/EventCell.js
```diff
@@ -67,7 +69,9 @@ class EventCell extends React.Component {
         selected
       )
 
+    let _draggable = get(event, draggableAccessor)
     let wrapperProps = {
+      draggable: _draggable === null ? true : _draggable,
```
jquense: `null` is a bit wonky here. I'd say go with `draggableAccessor ? !!get(event, draggableAccessor) : true`

src/addons/dragAndDrop/DraggableEventWrapper.js
```diff
@@ -81,22 +83,27 @@ class DraggableEventWrapper extends React.Component {
...
-        <div className="rbc-addons-dnd-resize-ns-anchor">
-          <div className="rbc-addons-dnd-resize-ns-icon" />
-        </div>
-      )
-      StartAnchor = !continuesPrior && connectTopDragSource(anchor)
-      EndAnchor = !continuesAfter && connectBottomDragSource(anchor)
+    /*
+      *  We add anchors only if the event is indeed draggable
+     */
+    if (draggable) {
```
jquense:
```
if (!draggable) {
  return children
}
```

maybe?

src/EventCell.js
dbuchet-bandc: I've used it because your `get` (from accessors) returns `null` in this case.

Speaking for the logic, if `get(event, draggableAccessor)` is `null`, draggable should be `true` (to enable dragging by default). So if I do `!!` , it will result as `false`, which is not what we want here :)

If you want to keep your syntax, I can define default `draggableAccessor: null` in Calendar `defaultProps`, 
or I can set it to `draggableAccessor: () => true`,  then we'll just need 
```
let wrapperProps = {
  draggable: get(event, draggableAccessor),
```
But everyone will have to set the props `draggableAccessor`  as a function  to use it

Let me know what you prefer :)

jquense: > So if I do !! , it will result as false

Yeah i think that makes sense if you provide a draggableAccessor the intent is you want to determine if it's draggable, i'd prefer to split out the distinction between "I returned a falsy value from draggableAccessor" and "I did not provide a draggableAccessor", I think it makes sense to default true if they didn't provide it, and if they did assume the return value indicates  something.

jquense: sounds good!

src/Calendar.js
```diff
@@ -175,6 +175,17 @@ class Calendar extends React.Component {
...
+    /**
+     * Determine if an event is draggable. Must resolve to a Boolean
+     *
+     * ```js
+     * string | (event: Object) => Boolean
+     * ```
+     *
+     * @type {(func|string)}
+     */
+    draggableAccessor: accessor,
```
jquense: This should go in https://github.com/dbuchet-bandc/react-big-calendar/blob/f2c9cf454b795064076b879c272192897ab2b4ec/src/addons/dragAndDrop/withDragAndDrop.js#L65 since it's not relevant to the core calendar only the addon

src/DayColumn.js
```diff
@@ -219,8 +221,10 @@ class DayColumn extends React.Component {
 
       let { height, top, width, xOffset } = style
 
+      let _draggable = get(event, draggableAccessor)
       let wrapperProps = {
         event,
+        draggable: _draggable === null ? true : _draggable,
```
jquense: I didn't notice this before sorry, but logic specific to the DnD addon should live completely in the addons code and not here or in Event. Instead you want to add the `draggableAccessor` to context in DraggableCalendar, like we do for the other accessors: https://github.com/dbuchet-bandc/react-big-calendar/blob/f2c9cf454b795064076b879c272192897ab2b4ec/src/addons/dragAndDrop/withDragAndDrop.js#L95 and then use it off of context in the DnD specific wrapper components

dbuchet-bandc: I there,
Sorry, I'm afraid i'm not familiar enough with DnD to figure how to retrieve this context in DragSource.
I've searched in DnD documentation, and I don't see how we can get context before a `beginDrag` (which occurs too late). CanDrag doesn't allow to retrieve `Component` (so its context)

Thanks for your time anyway!

jquense: ah looked a bit more closely i see how `canDrag` is like the only method that doesn't provide the component :/

It looks like we need to wrap the DragSource in another component that injects the draggable from context...

dbuchet-bandc: Yes that's what I tought too.
I'll check what I can do :)

dbuchet-bandc: Mmmmmh, in fact, I don't know why I wanted to have context in DragSource... if event is not draggable, I should not return a connected Wrapper, so I just have to get context in DraggableEventWrapper.
I'll do some test and submit an update

arecvlohe:
Looking good!

PR #813, opened 2018-04-19
Fix to prevent scrollToTime from being recalculated when it shouldn't

rmanalan:
`date-arithmentic`'s `eq()` method [doesn't actually accept a third param](https://github.com/jquense/date-math/blob/master/index.js#L85). Instead it does a full dttm compare. This causes the `scrollToTime` to be recalculated whenever any prop in the `TimeGrid` component changes. To fix this, we just use `moment().isSame()` instead.

### DIFF
--- a/src/TimeGrid.js
+++ b/src/TimeGrid.js
@@ -123,8 +123,8 @@ export default class TimeGrid extends Component {
     const { range, scrollToTime } = this.props
     // When paginating, reset scroll
     if (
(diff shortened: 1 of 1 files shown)


jquense:
Hey there, thanks for the PR. It's not immediate obivous but `eq` (as well as the other operators, do accept a third argument, see here: https://github.com/jquense/date-math/blob/master/index.js#L235 Even if it didn't tho, we want to fix the problem there, instead of introducing a dependency on `moment` for just the one function.

PR #823, opened 2018-04-27
Fixing TimeSlots getSlotMetrics

krijoh92:
I was about to make an issue, but figured I could create a pull request instead.

The issue I was about to post:

> #### Do you want to request a _feature_ or report a _bug_?
> Bug
> 
> #### What's the current behavior?
> The new TimeSlots metrics generate the wrong number of timeslots, so the last timeslot is not included (the one that goes to 23:59), making it unselectable. You can see it here:
> 
> - [CodeSandbox](https://codesandbox.io/s/13kqjp5o43)
> - http://g.recordit.co/Sy5qUDTpya.gif
> 
> I've also tested this with any given `min` and `max` value, the same happens, so it isn't just 23:59 timeslot that is excluded. 
> 
> #### What's the expected behavior?
> The last timeslot in any given day should be selectable.

This change should fix this issue. I've tested it with multiple configurations of `min`, `max` and `timeslots`, and the issue was present for all of them and fixed with this change for all of them

### DIFF
--- a/src/utils/TimeSlots.js
+++ b/src/utils/TimeSlots.js
@@ -28,7 +28,7 @@ export function getSlotMetrics({ min: start, max: end, step, timeslots }) {
   for (let grp = 0; grp < numGroups; grp++) {
     groups[grp] = new Array(timeslots)
 
-    for (let slot = 0; slot <= timeslots; slot++) {
+    for (let slot = 0; slot < timeslots; slot++) {
       const slotIdx = grp * timeslots + slot
       const minFromStart = slotIdx * step
       // A date with total minutes calculated from the start of the day
@@ -43,6 +43,19 @@ export function getSlotMetrics({ min: start, max: end, step, timeslots }) {
       )
     }
   }
+  
+  const lastSlotMinFromStart = slots.length * step
+  slots.push(
+    new Date(
+      start.getFullYear(),
+      start.getMonth(),
+      start.getDate(),
+      0,
+      minutesFromMidnight + lastSlotMinFromStart,
+      0,
+      0
+    )
+  )
 
   function positionFromDate(date) {
     const diff = dates.diff(start, date, 'minutes') + getDstOffset(start, date)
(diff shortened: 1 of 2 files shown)


src/utils/TimeSlots.js
```diff
@@ -28,7 +28,7 @@ export function getSlotMetrics({ min: start, max: end, step, timeslots }) {
   for (let grp = 0; grp < numGroups; grp++) {
     groups[grp] = new Array(timeslots)
 
-    for (let slot = 0; slot < timeslots; slot++) {
+    for (let slot = 0; slot <= timeslots; slot++) {
```
tobiasandersen: This will add an extra timeslot to every step, which probably isn't what you want.

krijoh92: Ah yes, I see that now :) I didn't really dive into what all that code did, only tried it out and found it working. What do you suggest doing to give us that last slot? Only add an extra timeslot to the last group? As far as I could see it, adding an extra group would skewer the whole day a bit.

tobiasandersen: Sorry, I don't know and I don't have time to dive into it right now.

krijoh92: Or we could push an extra slot into the `slots` array, that way the last group won't have an extra slot, and it won't affect the rendering of the group. And then `closestSlotToPosition` would at least return the correct slot instead of `undefined` :)

krijoh92: Oh sorry, didn't see the previous comment you made. But as far as I can see it, it works with just pushing an extra timeslot at the end of the slots array. I can push those changes, and you can review whenever you have time :)

krijoh92:
I think this codechange is ready for reivew @tobiasandersen, if you or anybody else has time to review :)

src/utils/TimeSlots.js
```diff
@@ -43,6 +43,19 @@ export function getSlotMetrics({ min: start, max: end, step, timeslots }) {
       )
     }
   }
+  
+  const lastSlotMinFromStart = slots.length * step
```
jquense: is this used?

jquense: lets add a comment as to why this needs to be added? and also check on DST that it didn't break that day. 

A test as well would be appreciated :)

krijoh92: There wasn't any tests for this util yet, so I wasn't sure how extensive you wanted me to be. Hope the tests I pushed will suffice!

PR #829, opened 2018-05-04
Allow Infinite Scroll

breadbaker:
This adds a new prop infinityScroll when applied allows to navigate between months via scroll.

### DIFF
--- a/src/utils/dates.js
+++ b/src/utils/dates.js
@@ -1,6 +1,7 @@
 /* eslint no-fallthrough: off */
 import dateMath from 'date-arithmetic'
 import localizer from '../localizer'
+import moment from 'moment'
 
 const MILLI = {
   seconds: 1000,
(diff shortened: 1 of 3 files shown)


src/utils/dates.js
```diff
@@ -1,6 +1,7 @@
 /* eslint no-fallthrough: off */
 import dateMath from 'date-arithmetic'
 import localizer from '../localizer'
+import moment from 'moment'
```
jquense: can't use moment, it's not a direct dependency. use the functions from dateMath, instead (it's got the same api)

Bentopi:
This is an awesome feature I'd love to see implemented.

jquense:
Sorry for the delay :/

I've just had a cahnce to pull this down and try it out. Its really cool, but still feels a biut rough. Scrolling is very janky on the storybook i've added it too, I think that there are perhaps some assumptions about the size of the viewport. There is also a misalignment with the header due to the scrollbar

PR #830, opened 2018-05-04
Add declaration file to dnd in addons

Melocaster:
- If you want to use dnd from addons folder in typescript, current @types/react-big-calendar does not provide any declarations for using it. That's why this declaration file can help people using typescript.


jquense:
I'd prefer to keep types in DT as i don't want to maintain them thanks!

Issue #832, opened 2018-05-06
DnD drop not allowed for events in same row on different days


SLYtiger16:
#### Do you want to request a _feature_ or report a _bug_?
bug
#### What's the current behavior?

When using withDragAndDrop, the events are not droppable while on the same "row" as another event for a week (in month view). i.e. - if a day is full of events and you wish to drop some other event on a day in the same week as the 'full' day, you have to find space below the 'rows' of events on the 'full' day - see video image

"react-big-calendar": "^0.19.1"
"react": "^16.2.0"

[video image](http://recordit.co/SSDfOKDWMm)

SLYtiger16:
It appears that the `EventRow` component is preventing the `droppableDayWrapper` from returning canDrop = true. Hopefully this helps...

SLYtiger16:
Correct me if I am wrong but the `DroppableDayWrapper` is a dropTarget placed on the `BackgroundCells` under `div.rbc-row-bg`? The `div.rbc-row-content` which contains the actual event elements is rendered as a sibling to the `BackgroundCells` element, and therefore this issue would require a complete layout rewrite, no?

SLYtiger16:
Okay after some deeper investigation I have found that the underlying issue is when I do this: `withDragAndDrop(BigCalendar, {backend: false});` The event drop issue exists otherwise it performs as the examples and as expected. Unfortunately I must leave the backend as false because I need to place a DnD context provider in a higher component to allow for dnd labels from the outside to drop on each event. This creates an error with having 2 html5dnd backends and so I have to render the nested one as false.. my js-fu is not top notch but I will keep seeking a solution.

PR #833, opened 2018-05-07
Fixed DnD Add-on to use resourceID #807

### LINKED ISSUE #807: Can't use the next resourceId when item is dropped
_bug_

It seems that a next resource id is not used in the Drag and Drop addon.
https://github.com/ehahn9/react-big-calendar/blob/f730f2fee7cf4b99274542df37e0b9a446c17e8e/src/addons/dragAndDrop/DropWrappers.js#L151

I expect it should be use next resource ID like this commits demo.
https://github.com/intljusticemission/react-big-calendar/commit/0c0b17c045a65abaa84f14e91254e04cf34a1873#diff-4f25ee19436cd5ed38d0bce39dbd7f8aR114

mura-:
I tried to fix #807 .
Please check this PR 🙏

### DIFF
--- a/src/addons/dragAndDrop/DropWrappers.js
+++ b/src/addons/dragAndDrop/DropWrappers.js
@@ -131,7 +131,7 @@ function createDropWrapper(type) {
 
       const item = monitor.getItem()
       const { event, anchor } = item
-      const { value } = props
+      const { value, resource } = props
       const {
         onEventDrop = noop,
         onEventResize = noop,
@@ -151,6 +151,7 @@ function createDropWrapper(type) {
           onEventDrop({
             event,
             ...getEventDropProps(start, end, value, droppedInAllDay),
+            resourceId: resource,
           })
           return // all the other cases issue resize action...
 
@@ -194,7 +195,13 @@ function createDropWrapper(type) {
 
       // fall here for all of the resize cases
       // note: the 'drop' param is here for backward compatibility - maybe remove in future?
-      onEventResize('drop', { event, start, end, allDay: droppedInAllDay })
+      onEventResize('drop', {
+        event,
+        start,
+        end,
+        resourceId: resource,
+        allDay: droppedInAllDay,
+      })
     },
   }
 


SLYtiger16:
I have installed and tested this update ... aside from a pre-existing issue about dropping events [#832](https://github.com/intljusticemission/react-big-calendar/issues/832) this works well amd as expected ... thank you!

mura-:
@SLYtiger16 Thanks for your testing!

src/addons/dragAndDrop/DropWrappers.js
```diff
@@ -151,6 +153,7 @@ function createDropWrapper(type) {
           onEventDrop({
             event,
             ...getEventDropProps(start, end, value, droppedInAllDay),
+            [resourceAccessor]: resource,
```
jquense: `resourceAccessor` may be function, i would just call this `resource` since it's not hte event object.

mura-: @jquense Thanks!
Every accessors seem to be String.
https://github.com/intljusticemission/react-big-calendar/pull/833/files#diff-6cf28b0e3b0b32fa18e19efed78ee8b8R80


And... I expect `start` and `end` should support Accessor however, I didn't fix this PR. I think we have to fix on another issue.

jquense: the defaults are strings but the prop type allows functions. The other ones are already correct, and don't assume it's a string. It's just your usage here that's not right.

mura-: @jquense Sorry for replying so late. 
I'll check and solve that on this week.

mura-: @jquense 

I have a question.
Do I have to support resourceAccessor on DnD?

`start` and `end` seem to be fixed.
https://github.com/mura-/react-big-calendar/blob/614f01704c828f0f500ba166d13308daff75650c/src/addons/dragAndDrop/DropWrappers.js#L32

If I don't have to support resourceAccessor, What should I use the resource key name?

jquense: you only need it if you actually need to get the value from the user's event object. the returned value here is **not** an event object iso you don't need it here. I'd just call this `resourceId` if that's what it is

mura-: @jquense Thanks a lot. I will fix it.

mura-:
@jquense 
I have fixed resource's key. please take a look 🙏

PR #843, opened 2018-05-14
DnD support for custom DropWrapper components (dayWrapper and dateCellWrapper)

amkoehler:
When testing out the dnd addon I found this note in a comment in `withDragAndDrop.js`: 

```
 * Note: you cannot use custom `EventWrapper`, `DayWrapper` or `DateCellWrapper`
 * components when using this HOC as they are overwritten here.
```

This was a huge bummer for us because we are wanting to use custom `DayWrapper` or `DateCellWrapper` components for contextual styling on day and date cells while also using drag and drop. 

With this PR I took a stab at adding support for custom `DayWrapper` and `DateCellWrapper` components. I reviewed the discussion on #716 (initial implementation) to make sure I wasn't missing why this wouldn't be possible. 

- `DropWrapper` now receives the `components` prop. It will still use the default `BackgroundWrapper` if no custom `dayWrapper` or `dateCellWrapper` component is provided
- Added a few stories showing custom `dayWrapper` and `dateCellWrapper` (both for dnd and non-dnd)
- Cleaned up some code in `stories` and `examples` for consistency

### DIFF
--- a/src/addons/dragAndDrop/withDragAndDrop.js
+++ b/src/addons/dragAndDrop/withDragAndDrop.js
@@ -52,8 +52,9 @@ try {
  * If you care about these corner cases, you can examine the `allDay` param suppled
  * in the callback to determine how the user dropped or resized the event.
  *
- * Note: you cannot use custom `EventWrapper`, `DayWrapper` or `DateCellWrapper`
- * components when using this HOC as they are overwritten here.
+ * Note: you cannot use custom `EventWrapper` components when using this HOC as it
+ * is overwritten here.
+ *
  *
  * @param {*} Calendar
  * @param {*} backend
--- a/src/addons/dragAndDrop/withDragAndDrop.js
+++ b/src/addons/dragAndDrop/withDragAndDrop.js
@@ -78,6 +78,7 @@ export default function withDragAndDrop(
 
     static defaultProps = {
       // TODO: pick these up from Calendar.defaultProps
+      components: {},
       startAccessor: 'start',
       endAccessor: 'end',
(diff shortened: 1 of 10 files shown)


src/addons/dragAndDrop/withDragAndDrop.js
```diff
@@ -52,8 +52,9 @@ try {
  * If you care about these corner cases, you can examine the `allDay` param suppled
  * in the callback to determine how the user dropped or resized the event.
  *
- * Note: you cannot use custom `EventWrapper`, `DayWrapper` or `DateCellWrapper`
- * components when using this HOC as they are overwritten here.
+ * Note: you cannot use custom `EventWrapper` components when using this HOC as it
```
amkoehler: @ehahn9 I think it was you who did this initial implementation. Did you try implementing custom `EventWrapper` support?

ehahn9: Thanks for the ping... Indeed I took a look at it but wimped out due to lack of time. This was on my wish list (*), so thanks for working on it! Our app doesn't require it, so I probably won't have time to review your excellent PR, tho. Anyone else?

(*) my other two hot buttons were per-event draggability and much better handling to the click geometry for dnd (right now the drag tracking is truly funky because it isn't based on the mousedown location.. others had spent time on the initial DnD implementation but we all gave up...)

amkoehler:
Hi @jquense and @tobiasandersen, there's a good number of approved PRs right now including this one. Any ideas when you'll be cutting a release? No worries if you're waiting on other stuff.

PR #844, opened 2018-05-15
Dnd month date header fix

amkoehler:
When using DnD, the date header on month views is not clickable despite there being an `<a>` tag that should take you to the drilldown view when clicked. I traced this down to the `pointer-events` styling done in the DnD addon's styles.less.


amkoehler:
Apologies. This contained code from my other PR that has been approved. I'll make a different one.

PR #847, opened 2018-05-18
Option resourceHeaderPropGetter

marwej:
Added support for an option `resourceHeaderPropGetter` to enable custom styles on the resource headers. My use case was to categorize the resources into "resource groups" with colors.


jquense:
Can you rebase this? sorry for the delay!

healthycola:
@marwej can you please rebase this? This seems really useful to have.

marwej:
@healthycola Hm, I'm not really sure how to do this. Would you mind to explain how?

PR #850, opened 2018-05-21
Fixed determining events for a given day, regardless of current time

nunodelisboa:


jquense:
can we get some context for this?

nunodelisboa:
@jquense sure. 

I came across an issue where some events failed to appear to the Agenda view.
After some investigation I observed the following:

The `render()` method of the Agenda component starts by creating a `range` which is nothing but an array of `Date` objects, one for each day within the start and stop dates for the current page of the agenda. As it happens, these `Date` objects are created with the corresponding day/month/year, but they all keep the same time, which is their time of creation.

Further down, the `renderDay()` methods, which is responsible for rendering the events of each day, uses these `Date` objects to select the right events for the day. In order to do so, it relies on `inRange` method from `/utils/eventLevels`, by passing the same object as both the `start` and `end` parameters.

It turns out that if an event is indeed from a given day, it will not be selected if its end time is not after the start time of the day object passed in the parameters. I don't think this is the intended behaviour, so my suggestion is not the pass the same `Date` object but rather the a `Date` object representing the start and end of the corresponding day.

alwick:
Any idea when this is going to be released?

andersonba:
Any news?

PR #852, opened 2018-05-24
Add support for onSelectEvent and onDoubleClickEvent in Agenda view

silentsakky:
Add support for onSelectEvent and onDoubleClickEvent in Agenda view

### DIFF
--- a/src/Agenda.js
+++ b/src/Agenda.js
@@ -3,6 +3,7 @@ import React from 'react'
 import classes from 'dom-helpers/class'
 import getWidth from 'dom-helpers/query/width'
 import scrollbarSize from 'dom-helpers/util/scrollbarSize'
+import cn from 'classnames'
 
 import localizer from './localizer'
 import message from './utils/messages'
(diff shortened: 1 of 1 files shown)


jku6:
up

morgs32:
@jquense have any thoughts on this?

jquense:
thanks for the work @silentsakky my concern here is that this only really makes sense with the default simple rendering. In most apps i've used RBC in i've use more complicated custom agenda components that contain their own buttons and clickable chrome. That would be really hard to handle with this addition because the event bubbling would fire for clicks inside the component.

This is already potentially a problem for the day view event clicks but i've not personally had that issue. In any case I think there should be a way to opt out of hte default behavior per view, so I may want the clicks for month or day views but not agenda. It's probably enough to pass the current view into the handler so a user can switch off of it

m1nd:
Up

jquense:
Hey y'all I'm closing this for now per my comments above, if someone wants to take a stay at a more robust addition I'd be happy to take it but folks are unhelpfully pinging ppl on this issue at this point so let's close it out

PR #865, opened 2018-06-07
Pass Events to DateHeader Component

pl12133:
**Do you want to request a *feature* or report a *bug*?**

Feature

**What's the current behavior?**

The `DateHeader` component is aware of it's own `date`, but it is not aware of the events occurring on that `date`.

**What's the expected behavior?**

I would like to pass the events corresponding to the `date` of the `DateHeader` so the `DateHeader` is able to **apply styles based on the events of that day**.

This use case has arisen from the following requirement: "When a day has at least 1 event, the date header should be **bold**."

---

Let me know what you think, and thank you for considering my feature request.

### DIFF
--- a/src/Month.js
+++ b/src/Month.js
@@ -209,6 +209,7 @@ class MonthView extends React.Component {
 
   readerDateHeading = ({ date, className, ...props }) => {
     let {
+      events,
       date: currentDate,
       getDrilldownView,
       dateFormat,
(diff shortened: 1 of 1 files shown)


jquense:
I'm a bit concerned about potential perf issues for doing the extra filtering for calendars with many dates, since in most cases the events are used. I think this can be worked around without a change by passing the events you're interested in to a custom date header component

pl12133:
@jquense Thanks for taking a look, performance concerns sound reasonable so I can accomplish this with a custom date header like you mentioned.

PR #866, opened 2018-06-08
Fix resourceId not receive in onEventDrop for dnd with resources

open-digitalp:


src/DayColumn.js
```diff
@@ -59,7 +59,7 @@ class DayColumn extends React.Component {
     timeSlotWrapperComponent: elementType,
     eventComponent: elementType,
     eventWrapperComponent: elementType.isRequired,
-    resource: PropTypes.string,
+    resource: PropTypes.number,
```
jquense: this should probably be `any`

jquense:
this should be fixed now

PR #875, opened 2018-06-13
per event resizability

sandorvasas:
added $resizable to event fields which controls whether the resize anchors should be visible or not

### DIFF
--- a/src/addons/dragAndDrop/DraggableEventWrapper.js
+++ b/src/addons/dragAndDrop/DraggableEventWrapper.js
@@ -97,49 +97,50 @@ class DraggableEventWrapper extends React.Component {
      * in the middle of events when showMultiDay is true, and to
      * events at the edges of the calendar's min/max location.
      */
+    if (typeof event.$resizable === 'undefined' || event.$resizable === true) {
+      if (isRow || allDay) {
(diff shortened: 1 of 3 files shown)


arecvlohe:
@sandorvasas We should use an accessor here so that user can name the prop however they want. Otherwise it's looking good.

src/addons/dragAndDrop/DraggableEventWrapper.js
```diff
@@ -97,49 +97,50 @@ class DraggableEventWrapper extends React.Component {
...
-          <div className="rbc-addons-dnd-resize-ew-icon" />
-        </div>
-      )
-      StartAnchor = !continuesPrior && connectLeftDragSource(anchor)
-      EndAnchor = !continuesAfter && connectRightDragSource(anchor)
-    } else {
-      const anchor = (
-        <div className="rbc-addons-dnd-resize-ns-anchor">
-          <div className="rbc-addons-dnd-resize-ns-icon" />
+    if (typeof event.$resizable === 'undefined' || event.$resizable === true) {
```
arecvlohe: Use an accessor here instead of naming the prop explicitly.

PR #876, opened 2018-06-13
per event moveability

sandorvasas:
added $moveable to event fields which controls whether the event is moveable by dragging or not


jquense:
Hey there, thanks for the PR. I believe this has already been added? https://github.com/intljusticemission/react-big-calendar/pull/808

Any additional PR's should follow the lead set in that PR. RBC tries not to specify specific new  properties to events, they should be configureable via an accessor function/field like they others. thanks!

arecvlohe:
I suspect this won't be pursued in favor of #808

PR #877, opened 2018-06-13
per event selectability

sandorvasas:
added $selectable to event fields which controls whether the event is selectable with a single click


jquense:
Hey there again thanks for all the PR's. event selecting is already a controllable prop, meaning if you don't want to allow certain events to be selected you can directly control the `selected` value and prevent changing the value for the events you wish to not select.

```js
onSelect(event) {
  if (!event.$selectable) return
  this.setState({ selected: event })
}
render () {
  return <Calendar selected={this.state.selected} ... />
}
```

PR #880, opened 2018-06-18
DnD support for custom EventWrapper

amkoehler:
This is a continuation of the work I did on #843. 

When testing out the dnd addon I found this note in a comment in `withDragAndDrop.js`: 

```
 * Note: you cannot use custom `EventWrapper`, `DayWrapper` or `DateCellWrapper`
 * components when using this HOC as they are overwritten here.
```

With #843 we added support for `DayWrapper` and `DateCellWrapper` components. This PR continues that work and adds support for custom EventWrapper components. In hindsight I wish I had done this as part of that PR because it was a lot simpler than I thought it'd be. 

Added stories for custom eventWrapper and custom eventWrapper with DnD.

### DIFF
--- a/examples/App.js
+++ b/examples/App.js
@@ -22,6 +22,7 @@ import Popup from './demos/popup'
 import Rendering from './demos/rendering'
 import CustomView from './demos/customView'
 import Resource from './demos/resource'
+import DndResource from './demos/dndresource'
 import Timeslots from './demos/timeslots'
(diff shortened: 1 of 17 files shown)


ehahn9:
Thanks, @amkoehler for your great improvements the early DnD work! Much appreciated!

amkoehler:
@ehahn9 anytime. You did the hard part.

PR #884, opened 2018-06-24
fix selecting events in chrome

### LINKED ISSUE #630: Bug with selecting events in Chrome
When trying to select an existing event in Chrome, a new event is created instead.

As indicated by the popups on the Demo page:

- the error occurs most of the time when clicking on the left part of the event (title or background)
- sometimes works fine when clicking close to the right border
- works fine in FireFox (also when clicking on the background)
- full day event selection seems to work fine

### Steps to reproduce:

Open the official Demo page in Chrome, navigate to 'Selectable' and click on an event.

shundy:
Fixes #630 #614 

The issue can be reproduced in Google Chrome on Windows (but not on macOS).
In this browser, mouseMove event is called after mouseDown without moving mouse cursor.
So this fixes not to emit SelectEvent until cursor moved from initial position.

### DIFF
--- a/src/Selection.js
+++ b/src/Selection.js
@@ -322,6 +322,10 @@ class Selection {
       top = Math.min(pageY, y),
       old = this.selecting
 
+    if (!old && !(w || h)) {
+      return
+    }
+
     this.selecting = true
     this._selectRect = {
       top,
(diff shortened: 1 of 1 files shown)


jquense:
Thanks!

Issue #887, opened 2018-06-26
Drag and drop an event from one month to another


rmanalan:
#### Do you want to request a _feature_ or report a _bug_?
Feature

#### What's the current behavior?
I'd like to be able to drag and drop an event from one month to the next. This doesn't seem to be possible.

arecvlohe:
@rmanalan Can you create an example using codesandbox? You should already be able to do this.

arecvlohe:
This seems more like a bug if you are not able to do this already

rmanalan:
@arecvlohe here you go https://codesandbox.io/s/m5vr0wq5ox

The current version of `react-big-calendar` only allows you to drag and drop to a visible day on the month view. It would be great if there was a hot zone that could allow a user to initiate a next or previous navigation while dragging.

jquense:
yeah that isn't currently possible but I think we'd be more than open to making it so, if someone wants to work on it. I'd be interested what other solutions do for this.

ehahn9:
+1 for this excellent enhancement. Just for completeness (these might be YAGNI cases!):

* this would ideally work for any view (e.g. in day view, you could drop an event on days in the future or past). Not sure of what it means in each case, tho. BTW, I punted on DnD for agenda view altogether.

* I presume we would issue `onNavigate` (sort of like the toolbar does). Would need to think through what this means for controlled instances?

* If navigation has occurred, not exactly sure what happens if the user abandons the dnd explicitly (esc) or implicitly (invalid drop target): do we scroll back to where they started or stay at the scrolled-to date?

I hope these comments don't dissuade anyone from tackling this! It would be a wonderful addition, IMO!

arecvlohe:
@ehahn9 I agree, seems like a nice enhancement. In the project I am working on, I modified RBC to render a two month view to move things around from the current month to the next.

rmanalan:
PR here... looking for feedback. Thx! #896

PR #896, opened 2018-07-05
Add support for dragging to next or prev dates within view

rmanalan:
This is an implementation for #887. I took the approach of turning the left and right edges of the month and week views as the previous and next drag targets. Hovering over these targets for at least 1s will initiate the next or previous `onNavigate`.

month

week

For the most part, this copies what GCal does minus the animation and navigating multiple months/weeks by continuing to hover over the targets. I'd like to implement these two as well, but wanted to get early feedback on what y'all think about this approach.

### DIFF
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -715,6 +715,16 @@ class Calendar extends React.Component {
     getNow: () => new Date(),
   }
 
+  static childContextTypes = {
+    onNavigate: PropTypes.func,
+  }
+
+  getChildContext() {
+    return {
+      onNavigate: this.handleNavigate,
+    }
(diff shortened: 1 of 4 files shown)


jquense:
Looking great, Im very short on time for reviews but I think this is awesome and thanks a ton for your work!

arecvlohe:
I am going to pull this down over the weekend hopefully and check it out. Thanks!

jquense:
@rmanalan excellent work here! unfortunately we've moved the target underneath you and completely rewrote the DnD addon. Are you interested in updating to match the current code, otherwise i can give it a shot when i have time.

PR #900, opened 2018-07-10
Improve toolbar responsiveness for mobile devices

miltonmc:
### Before
rbc-issue

### After
rbc-fix

### Demo
I've added all necessary changes into ```src/calendar.css```
https://codesandbox.io/s/5x6x7y8wlp


jquense:
thanks!

PR #906, opened 2018-07-18
Update to create events on select slot

rhuanbarreto:
This will help the people on using the onSelectSlot event.


jquense:
thanks!

Issue #909, opened 2018-07-20
Bug on events width when changing the step or timeslots argument


bgelineau:
#### Do you want to request a _feature_ or report a _bug_?

bug

#### What's the current behavior?

The computation to see if events are on the same row is broken when someone changes the `step` or `timeslots` arguments. Linked to #886 but different, as the current problem arises from the hard coded minutes implemented in the computation.

![bug](https://user-images.githubusercontent.com/1941277/43008811-48460a28-8c3c-11e8-952f-2eaa68116e1e.png)

Full sample implementation on https://codesandbox.io/s/j39nx6x793

#### What's the expected behavior?

The events should each take the whole width.

arecvlohe:
Fixed in #910

Issue #886, opened 2018-06-26
Horizontal event alignment


sandorvasas:
#### Do you want to request a _feature_ or report a _bug_?
bug
#### What's the current behavior?
See screenshot: appointment indented further after the middle, 15 min appointment.
![rbc_bug](https://user-images.githubusercontent.com/9200077/41904992-f6c0a868-798d-11e8-896d-c63e3cb464f4.png)
#### Expected behavior:
The two events in the middle should be aligned horizontally to start at the same x position.

#### Suggested fix:
DayEventLayout.js
```
function onSameRow(a,b) {
  return (
    // Occupies the same start slot.
    (Math.abs(b.start - a.start) < 15) ||
    // A's start slot overlaps with b's end slot.
    (b.start > a.start && b.start < a.end)
  )
}
```

At the moment a,b events in the second part of the condition are swapped, and the first part, the difference is set to 30 mins. Should I create a PR?

arecvlohe:
Fixed in #910

PR #910, opened 2018-07-20
fix minimum start difference for same row computation (#886) (#909)

bgelineau:
Fix for issues #886 and #909. The computed `minimumStartDifference` corresponds to a half section, when using the default CSS of 40px minimum height for a section and 20px minimum height for an event.

### DIFF
--- a/src/DayColumn.js
+++ b/src/DayColumn.js
@@ -165,6 +165,8 @@ class DayColumn extends React.Component {
       rtl: isRtl,
       selected,
       startAccessor,
+      step,
+      timeslots,
       titleAccessor,
       tooltipAccessor,
     } = this.props
@@ -174,6 +176,7 @@ class DayColumn extends React.Component {
       startAccessor,
       endAccessor,
       slotMetrics: this.slotMetrics,
+      minimumStartDifference: Math.ceil((step * timeslots) / 2),
     })
 
     return styledEvents.map(({ event, style }, idx) => {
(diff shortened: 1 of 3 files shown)


jquense:
looks good can you please add some tests thanks!

bgelineau:
@jquense : I can, but I have not seen any automated tests for this part of the code. Do you mean adding a story ?

jquense:
yes there are few tests, so we want to be proactive about adding them going forward. There isn't an existing set for this section. but you can add the test file and just add tests relevant to your change. We'll be working on backfilling more in general in the future 👍

bgelineau:
No problem, I'll do it early next week.

jquense:
Awesome thanks a ton @bgelineau

amkoehler:
@bgelineau would you mind taking a look at #916? I think we're fixing different problems but I'm not 100% sure.

bgelineau:
Hi @amkoehler, your problem is fixed with my commit. There may be some combination of events where it's important to check the dates both ways but I've no example yet. However, you still have a problem if event B does not overlap with event C but the fix will be more complex.
@jquense : could you pull this commit https://github.com/bgelineau/react-big-calendar/commit/05829b3d8935864b007932ed7c76dc03345e7e3c ? It adds a test for @amkoehler and fixes my own code on how to check the output of getStyledEvents for complex combinations of events.

amkoehler:
@bgelineau awesome! I'll revoke #916. Funny that we both were working on that on the same day. It looks like the fix for my issue was swapping the overlap comparison from 

```js
a.start > b.start && a.start < b.end
```

to 

```js
b.start > a.start && b.start < a.end
```

jquense:
@bgelineau I can't really pull the commit easily at this point, can you submit a follow up PR i can merge? thanks ya'll

PR #911, opened 2018-07-23
Propose better captions for the messages

rhuanbarreto:
Having uncapitalized messages turns the calendar ugly in the presentation.

I also propose putting font awesome chevrons on the previous and next. What do you think?


arecvlohe:
Seems a bit unnecessary at least from my point of view.

jquense:
I think this is fine, it does look better as a default.

PR #915, opened 2018-07-24
Fix rtl prop has to be passed to the DateContentRow

fingerpich:
I have made [jalali version of this component](https://github.com/fingerpich/jalali-react-big-calendar) and I have this problem to make it rtl.


jquense:
Thanks!

Issue #899, opened 2018-07-09
Day layout covering up entire events for certain sets of times


amkoehler:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

I think I found an edge case with the day view layout algorithm. For a certain subset of times we found that an event can get covered up entirely. I've created a small repo using the latest version of RBC which reproduces the issue: https://github.com/amkoehler/rbc-event-layout-bug. 

Summary: I have a set of 5 events on a given day. Based on the times of those events, the event algorithm groups those into 2 "container" events. The first container event contains 2 "row" events and the second container event contains 1 row event for the total of 5:

| Name  | Time | Type |
| ------------- | ------------- | ------------- | 
| Event A  | 8:15am - 1:15pm | Container A
| Event B  | 8:45am - 3:00pm | Row (Container A)
| Event C  | 10:45am - 5:00pm | Row (Container A)
| Event D  | 2:00pm - 8:00pm | Container B
| Event E  | 3:30pm - 8:30pm | Row (Container B)

The issue, which you can see by in my example repo, is that Event E is almost entirely covered up by Event C. If Event C were a bit longer in duration (past 8:30), Event E would be entirely covered up. 

![](https://i.imgur.com/IyabmzL.png)

I was hoping to create a PR for this but got hung up on a couple things in the event layout algorithm (though it was a great learning experience!). Looking through @tobiasandersen 's work on #677, where the algorithm was implemented, I think something needs to change with how the event widths are calculated. `sortByRender` appears to be returning an order consistent with Google Calendar, so I don't think it's that function. My best guess is that row & leaf event widths need to be aware of other row events to make sure they don't cover each other up. 

#### What's the expected behavior?

The event in question does not get covered up 😃

amkoehler:
Hi @tobiasandersen, any thoughts on this?

amkoehler:
This has been resolved with #910. Will close it up after next release if that's ok.

PR #916, opened 2018-07-24
Day layout bugfix

amkoehler:
I think I've found a fix for #899. This allows for more events to be classified as 'leaves' instead of 'rows' in the day layout algorithm, which I think was the intended approach. @tobiasandersen sorry I keep bugging you about this. Hoping this is the fix. 

Before:

before

After:

after


amkoehler:
Closing this PR in favor of #910. That resolves the same issue.

PR #917, opened 2018-07-24
Revamp Drag and drop

jquense:
ugh so i know i said i wasn't going to work on this, but I had some time this weekend and felt inspired at a simpler approach than what we've been doing.

A lot of the code changes here are unrelated core hygiene improvements, but they do help make the DnD implementation simpler. 

Ultimately i realized that we already had most of the hard work done for doing a custom DnD implementation thanks to the event Selection logic. Rather than use a new library i've leveraged that to do direct calculations on where things should be in relation to the mouse


ehahn9:
you are a rockstar @jquense!

arecvlohe:
😲 Awesome!

dmitrykrylov:
@jquense when will that be released?

sendajob:
How can we npm update the change?

miltonmc:
@sendajob  You can change your package.json to directly point to the repository on Github, to get the latest updates on it: https://docs.npmjs.com/files/package.json#github-urls

````json
...
dependencies: {
    ...
      "react-big-calendar": "intljusticemission/react-big-calendar",
    ...
}
````

**Comment by miltonmc** [comment]
@sendajob can you provide some additional information, like if you are using yarn or npm, your OS, the exact error msg, a screenshot and/or your package.json?

**Comment by sendajob** [comment]
When i set dependency 
"react-big-calendar": "intljusticemission/react-big-calendar",

the react-big-calendar/ in node_modules only has 
 node_modules/

no src no dist

**Comment by dmitrykrylov** [comment]
I tried to fetch dependency from github in the way @miltonmc proposed. It was successfully imported, but I got `accessors.start is not a function` error. I tried to debug it but haven't managed to fix the error and switched back to the current npm-indexed version.

**Comment by jquense** [comment]
we still need #921 which has a few bug fixes to this initial PR

**Comment by jquense** [comment]
> is that the events lose their height while you're dragging them.

hmm that doesn't sound right, in what context?

**Comment by amazingmarvin** [comment]
@jquense I am using the "day" Calendar.

```javascript
        // Probably not that useful - let me know if there's something
        // that would be helpful to include
        <Calendar
          localizer={localizer}
          defaultView="day"
          defaultDate={moment(date).toDate()}
          views={["day"]}
          toolbar={false}
          step={dayIncrement}
          timeslots={timeslots}
          events={items}
          selectable={false}
          onEventResize={this.resize}
          onEventDrop={this.drop}
          resizable={true}
          allDayAccessor={this.getTaskWithoutTime}
          startAccessor={this.getTaskStartTime}
          endAccessor={this.getTaskEndTime}
          tooltipAccessor={this.getTooltip}
          components={{
            event: DayTask,
          }}
        />
```


calendardrag

amazingmarvin:
OK, the problem is that I'm using `startAccessor` and `endAccessor`, whereas `EventContainerWrapper` doesn't:

```javascript
    let end = dates.add(
      currentSlot,
      dates.diff(event.start, event.end, 'minutes'),
      'minutes'
    )
```

mitchdoesgit:
@jquense Could you do a publish now that #921 has been merged? Thanks!

nnals:
hello.
i would like to try the new drag'n'drop/resizing behavior but can't figure out how to add the master branch as a dependency to codesandbox. does it even work? or would i have to build first? in case it's possible could someone perhaps help me add the master branch to this [codesandbox](https://codesandbox.io/s/nn32k94qzj).
thank you very much!

Issue #922, opened 2018-07-27
Show a message rather than an header-only table on Agenda view with no events.


mulholio:
**Improvement:** Show a message rather than an header-only table on Agenda view with no events.

I thought it would be a good idea to show a message along the lines of 'you have no events' when the agenda view loads with no events. The current view looks like something hasn't loaded properly/generally isn't that nice.

#### Current behavior
![screen shot 2018-07-27 at 13 35 16](https://user-images.githubusercontent.com/20966690/43320864-eefd065e-91a1-11e8-9546-7bfc0c6e7cfe.png)

#### Desired behavior
Either remove the headers and show a message or put 'none' or something similar in a row below the table headers.

PR #923, opened 2018-07-27
Agenda no events msg

mulholio:
#### Solution to https://github.com/intljusticemission/react-big-calendar/issues/922

When in the agenda view, you often come across empty tables of events. This looks unfinished and is confusing to the user as it may appear that data has not loaded, rather than there is simply no data to display.

The solution is to display a message, informing the user that there are no events and the component is functioning correctly.

I think there may be a little bit of work to do in terms of styling/aligning the message. That said, I imagine most people are going to be adding their own project styles and therefore I feel it's okay to leave the styling basic and unopinionated.

### DIFF
--- a/src/Agenda.js
+++ b/src/Agenda.js
@@ -48,26 +48,34 @@ class Agenda extends React.Component {
 
     return (
       <div className="rbc-agenda-view">
-        <table ref="header" className="rbc-agenda-table">
-          <thead>
-            <tr>
-              <th className="rbc-header" ref="dateCol">
-                {messages.date}
-              </th>
-              <th className="rbc-header" ref="timeCol">
-                {messages.time}
-              </th>
-              <th className="rbc-header">{messages.event}</th>
-            </tr>
-          </thead>
-        </table>
-        <div className="rbc-agenda-content" ref="content">
-          <table className="rbc-agenda-table">
-            <tbody ref="tbody">
-              {range.map((day, idx) => this.renderDay(day, events, idx))}
-            </tbody>
-          </table>
-        </div>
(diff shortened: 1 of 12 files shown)


src/Agenda.js
```diff
@@ -48,26 +48,34 @@ class Agenda extends React.Component {
...
+              <table className="rbc-agenda-table">
+                <tbody ref="tbody">
+                  {range.map((day, idx) => this.renderDay(day, events, idx))}
+                </tbody>
+              </table>
+            </div>
+          </React.Fragment>
+        ) : (
+          <span className="rbc-agenda-empty">
+            There are no events in this range
```
jquense: this needs to be configurable. It should go in `messages` and added to the top level calendar prop-type as well

jquense: we should also add some default styling

miltonmc: Since we can configure other labels, I think it should be configurable as well. 
@jquense what do you think about that?

jquense: https://github.com/intljusticemission/react-big-calendar/pull/923#discussion_r205815806

mulholio:
Had another look and can't work out/decided the best place to put the configuration option. Would it be possible to give some pointers? Would I be right in thinking this goes in either `messages` or `components` props?

mulholio:
Have attempted to add `noEventsInRange` into the message `Calendar` props but seem to be getting a bug/issue with the month view not displaying properly (see below) when I try to test with an override message.

screen shot 2018-07-30 at 11 32 44

jquense:
@mulholio can you add a storybook story that demonstrates the issue? I can take a look at that

PR #925, opened 2018-07-29
docs: add header component

### LINKED ISSUE #924: components.header undocumented
_question_

Currently TimeGridHeader makes use of a header component that can be passed from Calendar. https://github.com/intljusticemission/react-big-calendar/blob/d2e02c4bfde41b39ea22db344cbd8b948b17fc8b/src/TimeGridHeader.js#L70

The property is however [not documented](http://intljusticemission.github.io/react-big-calendar/examples/index.html#prop-components). Is this indeed only an implementation detail or meant for actual usage? Currently this seems to be the only way to render a custom header for days in the week view.

eps1lon:
Closes #924 

I added some additional information about where this component is used. Since I'm not familiar with this package this might be incomplete.

### DIFF
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -649,6 +649,10 @@ class Calendar extends React.Component {
       eventContainerWrapper: elementType,
       dayWrapper: elementType,
       dateCellWrapper: elementType,
+      /**
+       * component used as a header for each column in the TimeGridHeader
+       */
+      header: elementType,
       timeGutterHeader: elementType,
 
       toolbar: elementType,


src/Calendar.js
```diff
@@ -649,6 +649,10 @@ class Calendar extends React.Component {
       eventContainerWrapper: elementType,
       dayWrapper: elementType,
       dateCellWrapper: elementType,
+      /**
+       * component used as a header for each column in the TimeGridHeader
+       */
+      header: elementType,
```
jquense: actually this is already documented below in the specific views segment of this proptype

eps1lon: I'm not sure I follow. What's holding this PR back?

jquense: The PR isn't needed the header component is already documented here

jquense:
Thanks for the PR, the docs have this covered already it looks like! https://intljusticemission.github.io/react-big-calendar/examples/index.html#prop-components

eps1lon:
It is currently possible however to pass `components={{ header: () => 'GlobalHeader' }}`

PR #928, opened 2018-07-30
Fix toolbar

mmmbacon:
Looked into a few submissions about the Toolbar not having an appropriate onView handler associated with a generated toolbar component. 

After some digging I noticed the 'onView' prop for the generated toolbar was named 'onViewChange'.

Corrected the prop name and seems to have cleared up the issue.

This is one of my first PR's so please advise if I am doing this totally wrong... Thanks

### DIFF
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -859,7 +859,7 @@ class Calendar extends React.Component {
             view={view}
             views={viewNames}
             label={label}
-            onViewChange={this.handleViewChange}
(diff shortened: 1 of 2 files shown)


klinskyc:
@jquense Thanks for this project! This was a breaking change for users updating from old react-big-calendar versions who were using custom toolbars, and was not called out in the changelog. 

I figured I'd add a note here in case anyone else ran into this

Issue #937, opened 2018-08-05
toolbar should be more customizable without the need to completely write a new component


jstacoder:
## Customizable Toolbar
* fullcalendar uses a declarative configuration api for specifying places to put elements in the toolbar
* this is very useful as it allows changing which buttons are where without neededing to completley replace the toolbar component. 

it would be very nice to have a similar declarative api here, ie: `toolbar` prop on `Calendar` could be an object that describes the shape of the toolbar ie: 
```javascript
toolbar={{
    left: 'today previous,next',
    center: 'title',
    right: 'month,week,day',
}}
```

i will submit a pull request to add this feature

PR #938, opened 2018-08-05
make toolbar more customizable via toolbar config object

jstacoder:
this is to add a declarative configuration api for the toolbar component. 
for more context on this pull-request please see #937


jquense:
Hey there thanks for the PR and sorry for the long delay. 

I prefer that we not add another sort of configuration language to the component. If we wanted to make this easier to extend, we should do it by wrapping up the existing components and exporting them so folks can rearrange and configure using the React component model instead.

something like:
```jsx
<Toolbar>
  <Toolbar.Fieldset>
    <Toolbar.NextButton />
  </Toolbar.Fieldset>
  <Toolbar.Fieldset>
    <Toolbar.ViewButton name='month' />
  </Toolbar.Fieldset>
</Toolbar>
```

PR #944, opened 2018-08-10
Support multiline events

### LINKED ISSUE #939: Support taller events in the month view
*Bug*

I'd like to render a two-line event in the month view of the calendar. I don't know how to make a codepen with RBC, but I've mocked up an example of this with the example code for custom rendering, by adding `<p>Second line...</p>` below [this line](https://github.com/intljusticemission/react-big-calendar/blob/master/examples/demos/rendering.js#L9).

Here is the result:

image

Take note that the 12th has 6 events, but it only shows two (and the second one is cut off). What's really happening in the DOM there is it's rendering all three events, then the +3 more", but it's getting cut off by the `overflow: hidden` for the week.

I would expect that in this case, it renders however many events can fit, then the "+X more", like it normally does for one-line events.

To add a little more info here, I've dug pretty deep into this code. The root of this issue is in [DateContentRow.getRowLimit](https://github.com/intljusticemission/react-big-calendar/blob/master/src/DateContentRow.js#L92-L98). The purpose of that function is to calculate how many rows of events we can render without overflowing. The way it does that is by rendering a dummy event and using dom-helpers's `getHeight` function to get the actual height of an event as it's rendered in the DOM. The problem with this approach is that it's assumed that the dummy event will render the same as the real events, and specifically for my use-case, it's assumed that it will be the height of one line.

I was hoping to find a solution that I could make a PR for, but I'm not really sure what the solution is here. Ideally, we could render the actual events to determine the height of them, but as it's written right now, there's sort of a catch-22 around that because we need to know how many rows we can render, in order to render the events.

If anyone has a good idea for a solution, I'd be happy to code it and submit a PR.

dkniffin:
Fixes #939

### DIFF
--- a/examples/demos/multilineEvents.js
+++ b/examples/demos/multilineEvents.js
@@ -1,0 +1,38 @@
+import React from 'react'
+import BigCalendar from 'react-big-calendar'
+import events from '../events'
+import dates from '../../src/utils/dates'
+
+let allViews = Object.keys(BigCalendar.Views).map(k => BigCalendar.Views[k])
+
+function Event({ event }) {
+  return (
+    <span>
+      <p>
+        {event.title}
+        {event.desc && ':  ' + event.desc}
+      </p>
+      {/* {Math.random() > 0.5 && <p>Second Line</p>} */}
+      <p>Second Line</p>
+    </span>
+  )
+}
+
+let MultilineEvents = ({ localizer }) => (
+  <BigCalendar
+    events={events}
+    views={allViews}
+    step={60}
+    showMultiDayTimes
+    max={dates.add(dates.endOf(new Date(2015, 17, 1), 'day'), -1, 'hours')}
+    defaultDate={new Date(2015, 3, 1)}
+    localizer={localizer}
+    components={{
+      month: {
+        event: Event,
+      },
+    }}
+  />
+)
+
+export default MultilineEvents
--- a/examples/App.js
+++ b/examples/App.js
@@ -25,6 +25,7 @@ import Resource from './demos/resource'
 import DndResource from './demos/dndresource'
 import Timeslots from './demos/timeslots'
 import Dnd from './demos/dnd'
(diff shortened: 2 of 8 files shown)


examples/demos/multilineEvents.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react'
```
jquense: Please move this to a storybook story not a documentation example

dkniffin:
Yeah, I guess I should have said this in the description: There's still an assumption being made here that all events must be the same size. I don't know how to accommodate events of varying heights. But now it will at least render an actual event from the events given to determine the size, rather than relying on a dummy one.

As for being much slower, is there a specific section you think would cause a performance hit?

I think I know what you're saying about making the calendar expand to fit the events. That's actually not quite what I'm looking for. I like the "+X more" and I'm fine letting RBC calculate that for me. What I had a problem with was the height being assumed based on a dummy with the `rbc-row` css class applied to it. I couldn't find an accurate way to change the size of that dummy without messing up a bunch of other styles. I thought about passing in a prop to RBC that was like "eventHeight" and using that in `getRowMetrics`, but I figured it'd be better to calculate it by rendering an actual event.

jquense:
@dkniffin Sorry such a long wait, i think we can merge this, if you can address the conflicts. thanks!

dkniffin:
@jquense I've actually been working on and off on these changes for a bit, and I think there's still a few issues. I'll ping here again once I know for sure everything is working right

dkniffin:
I am planning on revisiting this soon. I have a project that is actively using it, but I need to find the time to take another look. It should be soon.

antoinerousseau:
That'd be great to avoir that `white-space: nowrap;` on event titles and have them flow freely, but it means no more `rbc-row` to contain them if they have variable heights.
Not sure why these rows exist anyway.
I might fork this project and use more flexbox, unless you have an idea?

PR #945, opened 2018-08-10
Exposed onSelectStart, onSelectEnd to views

SudoPlz:
We're trying to find a way to add a **line** on the view (**and some text** indicating the time they've clicked on, while the **mouse is down**, and it will look like this:

but we have no way of knowing when the selection started or ended.

Hopefully this code will help in such cases.


SudoPlz:
Bah, that change alone is not enough, I'm closing and I'll re-open again when it's ready.

PR #950, opened 2018-08-14
Feature/selection time indicator line

SudoPlz:
Allows to specify a component (amongst the other components) that will only get rendered, after the user clicks, but before they start dragging.

Example:

```
components={{
    beforeSelectIndicatorComponent: (selectionProps) => <BigCalendarPreSelectIndicator {...selectionProps} />,
}}
```

Now I can do this after clicking:

See the dashed line and the time tool tip before dragging?

### DIFF
--- a/src/DayColumn.js
+++ b/src/DayColumn.js
@@ -89,16 +89,9 @@ class DayColumn extends React.Component {
         components &&
         components.beforeSelectIndicatorComponent != null
       ) {
+        const SelectIndicator = components.beforeSelectIndicatorComponent
         // if the user is about to start selecting but hasn't started dragging yet
-        selectionElement = React.cloneElement(
-          components.beforeSelectIndicatorComponent,
-          {
-            top,
-            height,
-            startDate,
-            endDate,
-          }
-        )
+        selectionElement = <SelectIndicator {...this.state} />
       } else {
         // the user has already started dragging/selecting a time slot
         selectionElement = (
--- a/src/DayColumn.js
+++ b/src/DayColumn.js
@@ -77,6 +77,41 @@ class DayColumn extends React.Component {
     this.slotMetrics = this.slotMetrics.update(nextProps)
   }
 
+  renderSelection() {
+    const { selecting } = this.state
+    if (selecting) {
+      let { top, height, startDate, endDate, noMovementYet } = this.state
+      const { localizer, components } = this.props
+      let selectDates = { start: startDate, end: endDate }
+      let selectionElement
+      if (
+        noMovementYet &&
+        components &&
+        components.beforeSelectIndicatorComponent != null
+      ) {
+        // if the user is about to start selecting but hasn't started dragging yet
+        selectionElement = React.cloneElement(
+          components.beforeSelectIndicatorComponent,
+          {
+            top,
+            height,
+            startDate,
+            endDate,
+          }
+        )
+      } else {
+        // the user has already started dragging/selecting a time slot
+        selectionElement = (
+          <div className="rbc-slot-selection" style={{ top, height }}>
+            <span>{localizer.format(selectDates, 'selectRangeFormat')}</span>
+          </div>
+        )
+      }
+      return selectionElement
+    }
+    return null
+  }
+
   render() {
     const {
       max,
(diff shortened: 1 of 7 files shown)


src/DayColumn.js
```diff
@@ -77,6 +77,44 @@ class DayColumn extends React.Component {
...
+      let selectDates = { start: startDate, end: endDate }
+
+      let selectionElement
+      if (
+        noMovementYet &&
+        components &&
+        components.beforeSelectIndicatorComponent
+      ) {
+        // if the user is about to start selecting but hasn't started dragging yet
+        selectionElement = components.beforeSelectIndicatorComponent({
```
jquense: this should be a real component not a render prop to be consistent with the other apis

src/DayColumn.js
```diff
@@ -275,9 +308,17 @@ class DayColumn extends React.Component {
     selector.on('selectStart', maybeSelect)
 
     selector.on('beforeSelect', box => {
+      const didNotClickOnAnEvent = !isEvent(findDOMNode(this), box)
+      if (
+        didNotClickOnAnEvent &&
+        this.props.components &&
+        this.props.components.beforeSelectIndicatorComponent
+      ) {
+        maybeSelect({ ...box, noMovementYet: true })
```
jquense: there shouldn't be side effects in beforeSelect if possible

src/addons/dragAndDrop/withDragAndDrop.js
```diff
@@ -136,10 +136,10 @@ export default function withDragAndDrop(Calendar) {
       )
 
       props.components = {
-        ...components,
         eventWrapper: EventWrapper,
         eventContainerWrapper: EventContainerWrapper,
         weekWrapper: WeekWrapper,
+        ...components,
```
jquense: this is handled on master now already

src/DayColumn.js
SudoPlz: Roger that, I'll make the change.

src/DayColumn.js
SudoPlz: Not sure I understand, can you explain a bit better what you mean by that?

src/addons/dragAndDrop/withDragAndDrop.js
SudoPlz: got it.

src/DayColumn.js
jquense: the idea of this callback is only to determine whether the selection should be handled, it should do any work, like storing coordinates. the `mousedown`, `click`, `selectStart`, etc handlers are for that

SudoPlz: Got it, so I can do something like:

```
selector.on('mousedown', box => {
      const didNotClickOnAnEvent = !isEvent(findDOMNode(this), box)
      if (
        didNotClickOnAnEvent &&
        this.props.components &&
        this.props.components.beforeSelectIndicatorComponent != null
      ) {
        maybeSelect({ ...box, noMovementYet: true })
      }
    })
```
?
I'll try that.

SudoPlz: Hmm, nope that doesn't work. I need that code to run on mouse down, but seems like `selector` never receives that event.

src/DayColumn.js
```diff
@@ -77,6 +77,34 @@ class DayColumn extends React.Component {
...
+      let { top, height, startDate, endDate, noMovementYet } = this.state
+      const { localizer, components } = this.props
+      let selectDates = { start: startDate, end: endDate }
+      let selectionElement
+      if (
+        noMovementYet &&
+        components &&
+        components.beforeSelectIndicatorComponent != null
+      ) {
+        const SelectIndicator = components.beforeSelectIndicatorComponent
```
jquense: sorry so delayed here :(

Can we just skip the component and render something like your screenshot as the default? Then folks can customize it via css instead

SudoPlz: Sure we can do that, I'll revise my PR when I find some free time and post back.

SudoPlz: https://github.com/intljusticemission/react-big-calendar/pull/1018 is ready

SudoPlz:
Closing this in favour of https://github.com/intljusticemission/react-big-calendar/pull/1018

PR #952, opened 2018-08-16
This fixes event clicks/double clicks

SudoPlz:
Currently, event clicks can't be clicked on the master branch, because the onClick event is passed on a react component (**NoopWrapper**) and not a Dom element

https://github.com/intljusticemission/react-big-calendar/blob/master/src/DayColumn.js#L198

This PR propagates the onClick and onDoubleClick to the containing div element, so that clicks on events work again.


jquense:
thanks!

PR #955, opened 2018-08-17
No longer duplicating events while dragging/resizing

SudoPlz:
Here's a set of fixes to the dnd add-on.
The most important change is this:

Expected behaviour:

Actual behaviour:

This PR fixes that by creating a subset of events (that doesn't contain the updatee - the event that is being dragged).

That happens for both move and resize actions.


jquense:
where is the relevant change here? A quick glance at this, i think i prefer the current strategy which is to just hide the dragging event visually with css, instead of trying to not render it

SudoPlz:
hmm, but the events are not hiding at all. Am I missing something? Which css class is responsible for hiding the duplicate event?

SudoPlz:
Do you want me to change the PR and make it so that the event get's hidden with CSS and not removed from the dom?

jquense:
Hmm it should already be doing it!

SudoPlz:
Interesting, that's not the case on my end, I'll check it out.

SudoPlz:
https://github.com/intljusticemission/react-big-calendar/pull/959

PR #956, opened 2018-08-18
fix: add @babbel/runtime devDependency

carrmelo:
Hey! I updated react-big-calendar today and the examples wouldn't run.

It seems like the devDependency of @babel/runtime was missing but the latest version wouldn't work either

After some research, I executed
npm install --save-dev @babel/runtime@7.0.0-beta.46

and the script npm run examples works again.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,32 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@4c/layout": {
+      "version": "0.1.1",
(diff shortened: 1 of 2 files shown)


carrmelo:
ok. I followed documentation of the latest @babel/runtime https://www.npmjs.com/package/@babel/runtime that's why I added as devDependency.

anyway, the dependency is missing and the examples script won't run when cloned if it's not installed.... it took me a while to figure it out so I thought about reporting it.

PR #959, opened 2018-08-20
No longer duplicating events while dragging/resizing (with css)

SudoPlz:
Here's a set of fixes to the dnd add-on.
The most important change is this:

Expected behaviour:

Actual behaviour:

This PR fixes that by creating a subset of events (that doesn't contain the updatee - the event that is being dragged).

That happens for both move and resize actions.

### DIFF
--- a/src/addons/dragAndDrop/EventWrapper.js
+++ b/src/addons/dragAndDrop/EventWrapper.js
@@ -3,6 +3,7 @@ import React from 'react'
 import cn from 'classnames'
 import { accessor } from '../../utils/propTypes'
 import { accessor as get } from '../../utils/accessors'
+import { isSameEvent } from '../../utils/helpers'
 
 class EventWrapper extends React.Component {
   static contextTypes = {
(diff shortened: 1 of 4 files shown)


SudoPlz:
@jquense I'm now using the `rbc-addons-dnd-drag-dragged-event` class name for the event that is being dragged, and it works great.

Check it out.

src/addons/dragAndDrop/styles.less
```diff
@@ -11,11 +11,6 @@
     right: 0;
   }
 
-  .rbc-addons-dnd-dragging {
```
SudoPlz: This is not used anywhere in the code.

src/addons/dragAndDrop/EventWrapper.js
```diff
@@ -119,25 +125,39 @@ class EventWrapper extends React.Component {
...
             {children.props.children}
             {EndAnchor}
           </div>
         ),
-      })
+      }
+
+      if (
+        draggable.dragAndDropAction.interacting && // if an event is being dragged right now
+        isSameEvent(draggable.dragAndDropAction.event, event) // and it's the current event
```
jquense: no need for the fancy helper, `===` should be fine here since we aren't editing this event

SudoPlz: I thought so too, but I tested and it only works until the user starts dragging, then the events are different.

jquense: Hmm that's a bug somewhere then, let me take a look

SudoPlz: ok cool

jquense:
thanks! merged in #985

PR #987, opened 2018-09-10
Adding tabbable events

schoettler:
Switching EventWrapper content into `<button>` for accessibility purposes (navigation by tab).

tabbable-events

### DIFF
--- a/src/less/event.less
+++ b/src/less/event.less
@@ -1,6 +1,6 @@
 @import './variables.less';
 
-.rbc-event {
+button.rbc-event {
   padding: @event-padding;
   background-color: @event-bg;
   border-radius: @event-border-radius;
--- a/src/less/event.less
+++ b/src/less/event.less
@@ -3,9 +3,12 @@
 .rbc-event {
   padding: @event-padding;
   background-color: @event-bg;
+  border: 0;
   border-radius: @event-border-radius;
   color: @event-color;
   cursor: pointer;
+  width: 100%;
+  text-align: left;
 
   .rbc-slot-selecting & {
     cursor: inherit;
@@ -14,6 +17,10 @@
 
   &.rbc-selected {
     background-color: darken(@event-bg, 10%);
+  }
+
+  &:focus {
+    border: 2px solid red;
   }
 }
 
@@ -44,4 +51,3 @@
   border-bottom-left-radius: 0;
   border-bottom-right-radius: 0;
 }
-
--- a/src/less/event.less
+++ b/src/less/event.less
@@ -3,7 +3,6 @@
 .rbc-event {
   padding: @event-padding;
   background-color: @event-bg;
-  border: 0;
   border-radius: @event-border-radius;
   color: @event-color;
   cursor: pointer;
@@ -20,7 +19,7 @@
   }
 
   &:focus {
-    border: 2px solid red;
+    outline: 5px auto @event-outline;
   }
 }
 
--- a/src/less/event.less
+++ b/src/less/event.less
@@ -1,6 +1,9 @@
 @import './variables.less';
 
-button.rbc-event {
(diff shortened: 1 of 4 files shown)


src/less/event.less
```diff
@@ -15,6 +16,10 @@
   &.rbc-selected {
     background-color: darken(@event-bg, 10%);
   }
+
+  &:focus {
+    border: 2px solid red;
```
jquense: can we rely on the default focus outline instead? that will also avoid the shifting of buttons as you tab and their border grows

src/less/event.less
```diff
@@ -3,6 +3,7 @@
 .rbc-event {
   padding: @event-padding;
   background-color: @event-bg;
+  border: 0;
```
jquense: i think we probably need a more aggressive button style reset if we are gonna use the button tag. e.g. 

```
button {
  padding: 0;
  margin: 0;
  border: none;
  box-shadow: none;
  background: none;
}
```

src/less/event.less
schoettler: Hmmm I think only Chrome has a default focus outline.

Looking at the agent stylesheet it seems to be:
`outline-color: rgb(59, 153, 252)`

schoettler: Perhaps we can override that behaviour for the events, since `outline` doesn't really respect the `border-radius`:
image

With:
```
 &:focus {
    border: 2px solid rgb(59, 153, 252); // (or hex equivalent, saved as variable)
}
```

And defaulting to:
```
.rbc-event {
  border: 2px solid @event-color;
}

Comment (jquense) [reply src/less/event.less]: you don't see an focus outline when tabbing through buttons on FF et all? They should all have something. 

The main reason i want to stick what the browser default here is it's easier for folks to reset and control themselves, as opposed to having to undo our styles and add something else. It doesn't need to be pretty, just easily changeable

File: src/less/reset.less [hunk 3]
```diff
@@ -4,6 +4,14 @@
   margin: 0;
 }
 
+button {
```
jquense: this will reset all buttons, we need to scope this to just events

schoettler:
Does it look good now? Will probably include readable content (aria labels) for screen readers to complete the accessibility parts on the events in the future o/

src/less/event.less
```diff
@@ -1,11 +1,13 @@
 @import './variables.less';
 
-.rbc-event {
+button.rbc-event {
   padding: @event-padding;
   background-color: @event-bg;
   border-radius: @event-border-radius;
   color: @event-color;
   cursor: pointer;
+  width: 100%;
```
jquense: instead lets do `display:block;` maybe?

src/less/reset.less
```diff
@@ -19,3 +19,11 @@ button.rbc-input::-moz-focus-inner {
   border: 0;
   padding: 0;
 }
+
+button.rbc-event {
```
jquense: its fine to move this up to the event.less file and combine the selector

schoettler:
Does it look good now? Will probably include readable content (aria labels) for screen readers to complete the accessibility parts on the events in the future o/

src/less/event.less
schoettler: `display: block` doesn't seem to apply the full width, how about both?

src/less/reset.less
schoettler: Under the same `.rbc-event` rule of `event.less`?

src/less/event.less
schoettler: Hmm `display: block` seems to ruin the `+5 more` behaviour when events can't fit vertically

schoettler:
Should we merge this? @jquense

jquense:
thanks @schoettler !

PR #1012, opened 2018-09-26
Use meaningful title for docs page

eps1lon:
This is not about what title to use but that a more meaningful title should be used. Makes navigation easier between tabs.


eps1lon:
#997

PR #1019, opened 2018-10-01
No longer hiding the header when we only show one day

SudoPlz:
I don't see any reason why that should be hidden with javascript.
That causes an issue in our project, and I think we should let people decide wether they want to hide that or not (with CSS preferably)

### DIFF
--- a/src/TimeGridHeader.js
+++ b/src/TimeGridHeader.js
@@ -175,12 +175,9 @@ class TimeGridHeader extends React.Component {
                 </div>
               </div>
             )}
-            {/* For rendering only one day no need to show the headers */}
-            {range.length > 1 && (
-              <div className="rbc-row rbc-time-header-cell">
-                {this.renderHeaderCells(range)}
-              </div>
-            )}
+            <div className="rbc-row rbc-time-header-cell">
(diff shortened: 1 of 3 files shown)


jquense:
I don't think this is a great default. You essentially have two headers now for the Day view the toolbar label and now the single header. It makes sense from implementation perspective but a lot less so from a visual usage one.

SudoPlz:
Maybe we should hide it with css then otherwise this behavior can not be overwritten

jquense:
Can you talk through your use case a bit so I have a sense why that would be desirable? It can be tough to discussion additions with a clear sense of what we're gain :)

SudoPlz:
@jquense Here's how we want our implementation to look like:

https://cl.ly/878d880870cc/%255B3c637355dcb42019c86bcfb82b68550a%255D_shouldBe.gif

Here's how it looks like now:

https://cl.ly/b414c5ea7cdc/Screen%20Recording%202018-10-02%20at%2001.46%20PM.gif


I made some changes so that it now hides itself with CSS instead of javascript.
This should make it so that the behaviour remains the same, but it allows that the visibility to be changed via css.

jquense:
thanks!

PR #1021, opened 2018-10-02
fix: event resizing

jdaberkow:
Resizing an event towards another day currently leads to a somewhat buggy in-between visualization state.
If we do not perform the pointInColumn check at this point, everything works as expected.


jquense:
Thanks for the PR, I don't think this is the right fix for the bug here the pointInColumn check is needed generally otherwise all the columns will report a selection at once

Issue #1017, opened 2018-10-01
Current time indicator disappears at midnight


rojobuffalo:
#### Do you want to request a _feature_ or report a _bug_?
bug

#### What's the current behavior?
The current time indicator disappears at midnight.

The condition in `positionTimeIndicator()` that checks `if (timeGutter && current >= min && current <= max)` is `false` and so `timeIndicator.style.display = 'none'` gets set. This happens if you load a page with RBC before midnight and then let it sit open until midnight and later. The default `max` is set to "today at local time 23:59:59". But the max doesn't get updated when the day ends. so `current <= max` is `false`.

Confirmed in Chrome and Firefox.

#### What's the expected behavior?
1) is this the expected behavior? this is relying on the default value for `max`.
2) should we add something that updates/refreshes `max` when `current` is midnight?

rojobuffalo:
I goofed and thought this feature was already part of this repo. I'm using this fork https://github.com/sudoplz/react-big-calendar/ . So maybe put this on hold or ignore it until the time indicator in PR #950 gets added.

SudoPlz:
I'm on it.

SudoPlz:
https://github.com/intljusticemission/react-big-calendar/pull/1018 is ready to be merged

SudoPlz:
Fix ^

jquense:
Thanks for the issue @rojobuffalo we've decided not to fix this right now, you can see the justification in  The time indicator interval is now handled by TimeGrid #1023

PR #1023, opened 2018-10-02
The time indicator interval is now handled by TimeGrid

SudoPlz:
- Fixes: https://github.com/intljusticemission/react-big-calendar/issues/1017

Unfortunately this pr also contains the changes from https://github.com/intljusticemission/react-big-calendar/pull/1018 but I take it that 1018 will be merged anyway.

### DIFF
--- a/src/DayColumn.js
+++ b/src/DayColumn.js
@@ -12,6 +12,7 @@ import { notify } from './utils/helpers'
 import * as DayEventLayout from './utils/DayEventLayout'
 import TimeSlotGroup from './TimeSlotGroup'
 import TimeGridEvent from './TimeGridEvent'
+import SelectIndicator from './SelectIndicator'
 
 class DayColumn extends React.Component {
   static propTypes = {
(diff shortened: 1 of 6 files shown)


jquense:
I think i'm pretty comfortable not fixing #1018, it's a real edge case, and `now` is used a lot of places besides this one which would still be sort of broken even with this change.

SudoPlz:
Is there any reason why you don't want to at least fix this one? I don't see any possible downsides.

jquense:
The downsides are minor, but big enough that i moved from having the indicator in the TimeGrid to the DayColumn. We want to avoiding keeping an interval state value in the grid to reduce unnecessary cascading updates in components that don't need it. In this case the whole grid updates to move the line in one column. IT doesn't seem like a big thing, but especially with Resources full rerenders can be costly. It's a minor thing but that's weighed with i think a minor bug

SudoPlz:
Got it, closing for now then.

SudoPlz:
@jquense I'm re-opening this because I'd like to fix the issue.

This time, I created a standalone component called `TimeIndicatorLine` that only updates itself every minute.
That way we avoid unnecessary cascading updates, and update the indicator line.

Also, because the current active day background of the current day remains the same even even if a day changes, I added a timer inside TimeGrid that ONLY updates the TimeGrid if the day changes (that means 1 re-render per 24 hours which I think is fine).

Can you review that please?
It should be pretty straight forward and without side-effects when it comes to performance.

PR #1025, opened 2018-10-03
added customFormats argument to moment and globalize localizer

jiangzhu-genie:
This PR added a customFormats argument to moment and globalize localizer function

Use case: in my calendar I want to format time string in gutter in different format but keep other format as default, I'll have code like

```
const localizer = BigCalendar.momentLocalizer(moment, {timeGutterFormat: 'LTS'});
```

### DIFF
--- a/src/localizers/globalize.js
+++ b/src/localizers/globalize.js
@@ -46,7 +46,13 @@ export let formats = {
   agendaTimeRangeFormat: timeRangeFormat,
 }
 
-export default function(globalize) {
+/**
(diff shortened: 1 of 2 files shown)


jquense:
is there a big advantage to this approach over passing a `formats` prop to the Calendar itself? Generally there aren't a lot of calendar instances in most apps

jiangzhu-genie:
👍 Good point, I'm new to the library and missed the formats prop of the Calendar. Will close the PR

Issue #1026, opened 2018-10-03
Feature Request - allow group columns by  Day first then Resource


jiangzhu-genie:
#### Do you want to request a _feature_ or report a _bug_?

I want to request a new feature to allow grouping columns by  date first then resources. 

#### What's the current behavior?

With TimeGrid for week view with multiple resources, current logic will show header as Resource and sub-header as Day.

#### What's the expected behavior?

I want it to be table to control the Calendar to apply the opposite logic, i.e. header as Day and sub-header as Resource

I have a resource schedule page built with FullCalendar and built to group by day first then resource. I'll need to provide same functionality if switch to RBC. I noticed the initial resource schedule implementation was group by day first then resource and it was changed later on.

I'm happy to work on the implementation for this and could use some suggestions.

PR #1027, opened 2018-10-03
WIP: allow change day/resoruce grouping logic

jiangzhu-genie:
This PR is for the feature request https://github.com/intljusticemission/react-big-calendar/issues/1026

initial implementation for communication purpose

all comments, suggestions are welcome

storybook - resource group by day first

### DIFF
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -14,6 +14,6 @@ module.exports = function(config) {
         rules.less({ browsers }),
       ],
     },
-    plugins: [...config.plugins, plugins.extractText()],
+    // plugins: [...config.plugins, plugins.extractText()],
   }
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,4 @@
+{
+  "eslint.autoFixOnSave": true,
+  "eslint.packageManager": "yarn"
+}
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -220,6 +220,14 @@ class Calendar extends React.Component {
     resourceTitleAccessor: accessor,
 
     /**
+     * Determines the group header of week view or work_week view with resources
+     *
+     * date - display date as group header and show resources as sub-headers
+     * resource - display resource as group header and show dates as sub-headers
+     */
+    resourceWeekViewHeader: PropTypes.oneOf(['day', 'resource']),
+
+    /**
      * Determines the current date/time which is highlighted in the views.
      *
      * The value affects which day is shaded and which time is shown as
@@ -712,6 +720,7 @@ class Calendar extends React.Component {
 
     resourceIdAccessor: 'id',
     resourceTitleAccessor: 'title',
+    resourceWeekViewHeader: 'resource',
 
     longPressThreshold: 250,
     getNow: () => new Date(),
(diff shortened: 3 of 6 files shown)


.storybook/webpack.config.js
```diff
@@ -14,6 +14,6 @@ module.exports = function(config) {
         rules.less({ browsers }),
       ],
     },
-    plugins: [...config.plugins, plugins.extractText()],
+    // plugins: [...config.plugins, plugins.extractText()],
```
jiangzhu-genie: Storybook cannot start on my local, it complains plugins.extractText is not a function, commented out temporarily

.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
```
jiangzhu-genie: will remove when out of WIP

src/TimeGridHeader.js
```diff
@@ -128,21 +119,58 @@ class TimeGridHeader extends React.Component {
...
+          {this.renderHeaderCell({ date, index, today })}
+        </div>
+        <div className="rbc-row rbc-row-resource">
+          {resources.map(([id, resource], idx) => (
+            <div key={`resource_${id || idx}`} className="rbc-header">
+              {accessors.resourceTitle(resource)}
+            </div>
+          ))}
+        </div>
+        {/* TODO - render date content row */}
```
jiangzhu-genie: I need some advice on how to refactor existing code to allow render DateContentRow to support day/resource grouping

src/Calendar.js
```diff
@@ -219,6 +219,14 @@ class Calendar extends React.Component {
      */
     resourceTitleAccessor: accessor,
 
+    /**
+     * Determines the group header of week view or work_week view with resources
+     *
+     * date - display date as group header and show resources as sub-headers
+     * resource - display resource as group header and show dates as sub-headers
+     */
+    resourceWeekViewHeader: PropTypes.oneOf(['day', 'resource']),
```
jiangzhu-genie: Probably not the best named prop 😄
If have better name, please advice

stories/Resources.js
```diff
@@ -4,6 +4,14 @@ import { storiesOf } from '@storybook/react'
 import { DragableCalendar } from './helpers'
 import resources from './helpers/resourceEvents'
 
-storiesOf('Resources').add('demo', () => (
-  <DragableCalendar events={resources.events} resources={resources.list} />
-))
+storiesOf('Resources')
```
jiangzhu-genie: The repository seems doesn't have component tests e.g. enzyme + snapshot.

What should I do to test my changes on components?

jquense:
I think you stumbled on the main reason why the current approach was taken. There is no easy way to refactoring DateContentRow to handle something like discrete row like the new grouping. 

Honestly, if we think about resizing or drag and drop as well the ux is unclear for the proposed grouping as well. I don't think it's probably worth the change in terms of added complexity :/

jiangzhu-genie:
@jquense Thanks

I agree with your opinion from a developer point of view. In our use cases, there's no events across multiple days therefore for most of our users, it make sense to group by day first then resource.

I'll see what I can do with the DateContentRow in my own folk.

TeaBough:
I'm really interested by this feature, did you managed to do something on your own @jiangzhu-genie ?

jiangzhu-genie:
@TeaBough  not really, we did some investigation with our customers and most of them are happy with  Resource then Date grouping so this issue has been de-prioritised for us.

PR #1034, opened 2018-10-08
avoid unnecessary setState to make dnd faster

jiangzhu-genie:
currently when start dragging an event, the handleInteractionStart get triggered multiple times but only first setState is required. This MR add a if check to avoid unnecessary setState calls.

The speed change is noticeable for DnD in resource schedule with week view.


jiangzhu-genie:
Just noticed it's a duplicate PR to https://github.com/intljusticemission/react-big-calendar/pull/1020

PR #1046, opened 2018-10-11
Fixed calculation of slots number for date when DST ends.

### LINKED ISSUE #1044: Redundant calendar slots for day when DST ends
Hi guys, first of all I wanted to thank you for your hard work with this amazing calendar.

report a bug

I tried the [demo page](http://intljusticemission.github.io/react-big-calendar/examples/index.html#intro) and you can find a screenshot below how I see the calendar there:

screen shot 2018-10-10 at 18 15 42

My current local timezone is **America/Los_Angeles (UTC-7:00)**. The problem is that I can see additional two hours in the calendar, so I have two `12 AM` and `1 AM` slots at the top and at the bottom. Also slot lines for Sunday are not aligned with other days. Note that this problem is reproducible only for week and day view and only for the week (or day) when Daylight Saving Time ends. Here DST ends on the 4th of November, so to reproduce the bug you need to navigate to the view with that date.

Calendar should be rendered with 24 hours max within a day.

#### Possible solution

I think the problem is in [this line of code](https://github.com/intljusticemission/react-big-calendar/blob/master/src/utils/TimeSlots.js#L4). I think instead of

    Math.abs(start.getTimezoneOffset() - end.getTimezoneOffset())

we should have

    start.getTimezoneOffset() - end.getTimezoneOffset()

I don't think we need to get absolute value here because as far as I understand [here](https://github.com/intljusticemission/react-big-calendar/blob/master/src/utils/TimeSlots.js#L15) we use it for calculating the number of slots and we need to adjust difference accordingly: add 1 hour when DST starts and subtract 1 hour when DST ends.

What do you think? I'll be happy to create a PR if you agree with the solution.

anna-kartynnik:
Should fix #1044.

### DIFF
--- a/src/utils/TimeSlots.js
+++ b/src/utils/TimeSlots.js
@@ -1,7 +1,7 @@
 import dates from './dates'
 
 const getDstOffset = (start, end) =>
-  Math.abs(start.getTimezoneOffset() - end.getTimezoneOffset())
+  start.getTimezoneOffset() - end.getTimezoneOffset()
 
 const getKey = (min, max, step, slots) =>
   `${+dates.startOf(min, 'minutes')}` +
@@ -105,7 +105,7 @@ export function getSlotMetrics({ min: start, max: end, step, timeslots }) {
       if (dates.lt(date, start, 'minutes')) return slots[0]
 
       const diffMins = dates.diff(start, date, 'minutes')
-      return slots[(diffMins - diffMins % step) / step + offset]
+      return slots[(diffMins - (diffMins % step)) / step + offset]
     },
 
     startsBeforeDay(date) {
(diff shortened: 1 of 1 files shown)


quyphan:
Before:

screen shot 2018-11-02 at 10 17 02 am

After:

screen shot 2018-11-02 at 12 47 57 pm

This fix looks good!  Tested it on the example.

mikeocool:
Hey would be great to get a release with this merged in! 

Alternatively if anyone has any suggestions on how to install the current master as dependency that would be great, installing directly from git doesn't seem to work, as `lib/` isn't committed and there's no prepare script defined.

Thanks!

anna-kartynnik:
@mikeocool I used `https://github.com/anna-kartynnik/react-big-calendar/tarball/fix/slots-number-when-dst-ends-build` in package.json for now.

PR #1047, opened 2018-10-16
Add zIndex to day column events

bgelineau:

A bug

calendar - wrong

Events on the end of a row overlapping another row are hidden by the row behind them. See the small bit of event on the end of the pointer.

Event 2 should be visible.

calendar - better

#### PR implementation notes

To solve this problem, Google Calendar set a zIndex corresponding to the number of the column the events belong to, starting from 4 and going up. In red the zIndex corresponding to the events below it.

gcal

I took the same approach, but as we don't compute columns across rows, I set a zIndex corresponding to the offset, 5 + xOffset. Having the leftmost event with a non-zero z-index will show users if they have a display bug sooner. Just setting the number in the row as a zIndex wouldn't work with the following example :

calendar - 5

I modified the zIndex given in the CSS file provided by the project but it may break the display of people using their own.

As a side note, the following example still don't work, but the way rows are computed would need to be rewritten and I don't have time right now : 

calendar - 4

### DIFF
--- a/src/less/time-column.less
+++ b/src/less/time-column.less
@@ -83,7 +83,7 @@
     left: 0;
     background-color: white;
     border-right: 1px solid @cell-border;
-    z-index: 10;
+    z-index: 120;
     margin-right: -1px;
   }
 
--- a/src/TimeGridEvent.js
+++ b/src/TimeGridEvent.js
@@ -25,7 +25,7 @@ function TimeGridEvent(props) {
 
   let userProps = getters.eventProp(event, start, end, selected)
 
-  let { height, top, width, xOffset } = style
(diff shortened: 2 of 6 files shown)


src/less/time-column.less
```diff
@@ -83,7 +83,7 @@
     left: 0;
     background-color: white;
     border-right: 1px solid @cell-border;
-    z-index: 10;
+    z-index: 120;
```
jquense: these new z-index feel a bit arbitrary to me, it seems like calendars with many overlapping events or very wide calendars would hit this limit pretty easily? Or maybe i'm remembering how offset works wrong :P

PR #1054, opened 2018-10-19
Fix misplacement of current time indicator

jdaberkow:
After switching the weekly calendar towards the current week, the current time indicator is initially misplaced. By checking whether the date of the component changed in the componentDidUpdate cycle and then calling triggerTimeIndicatorUpdate, this issue can get addressed.

### DIFF
--- a/src/DayColumn.js
+++ b/src/DayColumn.js
@@ -84,6 +84,15 @@ class DayColumn extends React.Component {
     this.slotMetrics = this.slotMetrics.update(nextProps)
   }
 
+  componentDidUpdate(prevProps) {
+    if (prevProps.date.getTime() !== this.props.date.getTime()) {
+      if (this.props.isNow) {
+        this.positionTimeIndicator()
+        this.triggerTimeIndicatorUpdate()
+      }
+    }
+  }
+
   triggerTimeIndicatorUpdate() {
     // Update the position of the time indicator every minute
     this._timeIndicatorTimeout = window.setTimeout(() => {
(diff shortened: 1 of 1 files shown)


src/DayColumn.js
```diff
@@ -188,7 +197,7 @@ class DayColumn extends React.Component {
       events,
       accessors,
       slotMetrics,
-      minimumStartDifference: Math.ceil(step * timeslots / 2),
+      minimumStartDifference: Math.ceil((step * timeslots) / 2),
```
rgazelot: I don't think that this modification is part of the PR

src/DayColumn.js
```diff
@@ -84,6 +84,15 @@ class DayColumn extends React.Component {
     this.slotMetrics = this.slotMetrics.update(nextProps)
   }
 
+  componentDidUpdate(prevProps) {
+    if (prevProps.date.getTime() !== this.props.date.getTime()) {
+      if (this.props.isNow) {
+        this.positionTimeIndicator()
+        this.triggerTimeIndicatorUpdate()
```
rgazelot: This is not useful to trigger the setTimeout again, the initial one is still running.

src/DayColumn.js
jquense: this is just prettier doing it's thing

src/DayColumn.js
jquense: I think it makes sense to trigger it here, the time has explicitly changed and you may not want to wait up to a minute for that to show

src/DayColumn.js
```diff
@@ -84,6 +84,15 @@ class DayColumn extends React.Component {
     this.slotMetrics = this.slotMetrics.update(nextProps)
   }
 
+  componentDidUpdate(prevProps) {
+    if (prevProps.date.getTime() !== this.props.date.getTime()) {
```
jquense: The resolution of this comparison should be minutes: `!dates.eq(prevDate, nextDate, 'minutes')

Georift: I believe that method comes from the package `date-arithmetic` which `utils\dates.js` extends.

Issue #1051, opened 2018-10-18
day and week views break when setting max time to a full hour


eschreiner:
#### Do you want to request a _feature_ or report a _bug_?

a bug :-(

- version 0.19.0 works fine
- versions 0.20.x show this behaviour

#### What's the current behavior?

When limiting the time range of a calendar by setting max to a full hour, e.g. `max={new Date(2018,1,1,20,0,0)}`, the day and week views do no longer align the events properly:

• the time gutter renders an additional hour slot
• the events are no longer aligned with the gutter

![image](https://user-images.githubusercontent.com/184471/47133401-edc13100-d2a8-11e8-9eee-f8b673ce9e6b.png)

#### What's the expected behavior?

That the events are rendered properly :-)

#### Steps to reproduce

In the examples file `demos/selectable.js`, insert line 42 as

```
max={new Date(2018,1,1,20,0,0)}
```

The day and week views' time gutters render an additional hour, events are not aligned.

```
max={new Date(2018,1,1,19,59,59)}
```

works (visually).

eschreiner:
views break also when setting `max` to a half hour, e.g. 17:30

PR #1057, opened 2018-10-22
Fix elements position on TimeGrid if max prop is set

### LINKED ISSUE #999: Bug 🐞: min + max props breaks layout.

Bug

Passing `min` and `max` props breaks the position of events.

https://codesandbox.io/s/14jq64y553

screen shot 2018-09-23 at 14 02 57

Events to be in the correct position.

dmitrykrylov:
Fixes #999, #1051

Calendar shows integer number of timeslot groups, therefore even if `max` prop is set to 23:00, it will still render area with 00:00 as highest border.

### DIFF
--- a/src/utils/TimeSlots.js
+++ b/src/utils/TimeSlots.js
@@ -105,7 +105,7 @@ export function getSlotMetrics({ min: start, max: end, step, timeslots }) {
       if (dates.lt(date, start, 'minutes')) return slots[0]
 
       const diffMins = dates.diff(start, date, 'minutes')
-      return slots[(diffMins - diffMins % step) / step + offset]
+      return slots[(diffMins - (diffMins % step)) / step + offset]
     },
 
     startsBeforeDay(date) {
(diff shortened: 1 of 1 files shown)


jquense:
Hey thanks for the PR! I'm not sure I understand how numslots * timeslots ends up differing from totalmin, assuming that min and max are compatible with the step and slots

dmitrykrylov:
@jquense, I'm not sure how to explain better, I don't event know if my pr is the really correct way to fix it.

In examples/demo/basic js you have
`max={dates.add(dates.endOf(new Date(2015, 17, 1), 'day'), -1, 'hours')}`
How it looks in browser:
screen shot 2018-10-24 at 00 03 12

We have `min = 00:00` and max = `23:00` in the example above
- your total value is just `max - min`, in our case is 23 hours
- in DayColumn.js you map over the timeslot groups and render an integer number of TimeSlotGroup. If Group range equals to 2 hours, and that's why it actually renders a range of 24 hours

Seems that before the revamping DnD it worked in the other way.

dmitrykrylov:
@jquense hi, are there plans to merge this PR or propose any other fix for misplaced elements?

PR #1058, opened 2018-10-23
Fix today's time indicator in week view

rgazelot:
When you navigate into weeks and come back to `today`, the time indicator position was reset to the very top of the column. To keep it accurate, we re-compute its position after the component did update.


rgazelot:
Accidentally push this PR. Ref to https://github.com/intljusticemission/react-big-calendar/pull/1054 for the fix.

PR #1065, opened 2018-10-28
fix: Examples not fully shown until first state change

Georift:
Previously sending someone to the example link like: https://intljusticemission.github.io/react-big-calendar/examples/index.html#cultures would not completely load the interface. (The culture drop down wasn't present until after menu selection)

I've moved the hash detection to componentDidMount() and used a set state to ensure we trigger a re-render.

To test select a non-default example and reload the page. Opening the menu and selecting the same value again should not cause any re-renders.

### DIFF
--- a/examples/App.js
+++ b/examples/App.js
@@ -45,20 +45,26 @@ const EXAMPLES = {
   dnd: 'Addon: Drag and drop',
 }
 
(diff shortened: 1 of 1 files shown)


jquense:
thanks!

PR #1069, opened 2018-10-29
Chore/update version

joelmeiller:
**Setup**
- add rollup to build lib
- adjust to optune style guides


arecvlohe:
@joelmeiller I don't think this is a request to update the origin but rather to create your own forked version per this commit: d6b785b0c8d4c5a55295c96c0ae766729b4a141e

Issue #1072, opened 2018-10-30
DnD: dragging is disabled if resizing is not allowed


xtursky:
#### Do you want to request a _feature_ or report a _bug_?

bug

#### What's the current behavior?

I cannot enable drag&drop without resizable functionality.
If event is draggable or not is incorrectly decided only by evaluating whether resizing is allowed or not.
So if I set draggableAccesor to true and resizableAccessor to false it won't work.

https://github.com/intljusticemission/react-big-calendar/blob/master/src/addons/dragAndDrop/EventWrapper.js#L118

#### What's the expected behavior?

To check if either isDraggable or isResizable and later add corresponding props based on these settings.

Change check to 
```
if (isResizable || isDraggable) {
```

And add anchors only if 
```
      if (isResizable) {
        // replace original event child with anchor-embellished child
        newProps.children = (
          <div className="rbc-addons-dnd-resizable">
            {StartAnchor}
            {children.props.children}
            {EndAnchor}
          </div>
        );
      }
```

xtursky:
sry, I misclicked

PR #1073, opened 2018-10-30
Fix dragging is disabled if resizing is not allowed

xtursky:
### Summary
Drag and Drop was disabled if resizing is not allowed. This PR should take into account both isDraggable and isResizable when evaluating.

### Issue
#1072


jquense:
thanks!

PR #1085, opened 2018-11-02
Adding background support

vmurillo:


vmurillo:
This pr is just to get feedback, please dont merge as it is now

jquense:
Hi there, thanks for the PR and bringing this up. We've talked a bit about this in the past (though the specific issue escapes me). I don't like the idea of conflating events with "background events", I think that's what FullCalendar does, and it feels like an API wart not something you'd build from scratch today. My guess is that since there is some technical overlap it made sense to write them in terms of each other technically, but as an api it feels weird 

In other words, it seems like "background events" aren't really events, they are regions drawn behind events, and I think it makes sense that they not be in the same array as `event` and generally not share the same public components. That would keep everyone from having to check what type events are when customizing as well

jquense:
closing for staleness, happy to pick this back up if you have time to work on it

PR #1100, opened 2018-11-09
onRangeChange: new argument

newsiberian:
Hi,

Added additional argument `view` for cases when `onRangeChange` called right after `handleViewChange` will be called. This guarantees to us to receive fresh `view` with fresh dates.


jquense:
thanks!

PR #1101, opened 2018-11-10
Pass resourceId to slotPropGetter

WebspressoCo:
This tiny change enables users of this library to add custom slot classes depending on the slot resource & time.

I think that this is the most flexible solution for visually supporting things like business hours, breaks, times designated to specific types of events and so on.


jquense:
thanks!

snuzi:
Hey, does anyone know why the slotPropGetter in a calendar with drag and drop does not provide the resourceId?  Thnx

PR #1110, opened 2018-11-14
Filter and range should include current day in agenda

tiagofucilini:
#850 didn't completely fixed the problem since events are being filtered in the render method and passed to renderDay. The agenda was also including the same day in two different ranges when you navigated forward or back because it was including start and end in the range (it was actually showing lenght + 1 days).

### DIFF
--- a/src/Agenda.js
+++ b/src/Agenda.js
@@ -221,7 +221,7 @@ Agenda.navigate = (date, action, { length = Agenda.defaultProps.length }) => {
 }
 
 Agenda.title = (start, { length = Agenda.defaultProps.length, localizer }) => {
-  let end = dates.add(start, length, 'day')
+  let end = dates.add(start, length - 1, 'day')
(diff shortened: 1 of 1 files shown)


src/Agenda.js
```diff
@@ -38,11 +38,13 @@ class Agenda extends React.Component {
...
     let { length, date, events, accessors, localizer } = this.props
     let { messages } = localizer
-    let end = dates.add(date, length, 'day')
+    let end = dates.add(date, length - 1, 'day')
 
     let range = dates.range(date, end, 'day')
 
-    events = events.filter(event => inRange(event, date, end, accessors))
+    events = events.filter(e =>
+      inRange(e, dates.startOf(date, 'day'), dates.endOf(end, 'day'), accessors)
```
jquense: the `startOf` and `endOf` calls should be outside the filter iterator so they aren't called many times

tiagofucilini:
Any news?

PR #1117, opened 2018-11-19
Add option to overwrite default TimeGridHeader component

janb87:
I would like to overwrite the TimeGridHeader component. Reason is to allow it to be sticky using something like react-sticky.

Eg:

```
import React from 'react';
import TimeGridHeader from 'react-big-calendar/lib/TimeGridHeader';
import { Sticky } from 'react-sticky';

export default function StickyTimeGridHeader(props) {
    return (
        <Sticky topOffset={105}>
            <TimeGridHeader {...props} />
        </Sticky>
    );
}
```

### DIFF
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -636,6 +636,7 @@ class Calendar extends React.Component {
      *   eventContainerWrapper: MyEventContainerWrapper,
      *   dayWrapper: MyDayWrapper,
      *   dateCellWrapper: MyDateCellWrapper,
+     *   timeGridHeader: MyTimeGridHeader,
      *   timeSlotWrapper: MyTimeSlotWrapper,
      *   timeGutterHeader: MyTimeGutterWrapper,
      *   toolbar: MyToolbar,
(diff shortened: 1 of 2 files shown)


jquense:
isn't the header already sticky?

janb87:
It is not sticky. It has a scroll (overflow: auto) inside the calendar itself. But in my case I want to have just one scrollbar on the entire page.

janb87:
@jquense If you want I can create a simple example so you can see how this is different from the current overflow behavior.

For browsers which support "position: sticky" this hook is not really needed. I mainly need this hook now to get it working in IE11.

jquense:
I think it's fine to add tho perhaps it's better to go with a proper polyfill for position sticky vs something like react-sticky. If we do think it's worth the wrapper, we should probably also do it for the time gutter for folks that want to polyfill that (used with resources)

jquense:
@janb87 thanks for looking into this more. My original thought with the polyfill was something that wouldn't require the wrapper at all, since the wrapper API ends up allow a lot more potentially buggy interactions as people do other things. 

Given your example is it just as easy to use the rbc classNames and stickyfill handle the work from ouhtside the calendar? what do you think?

janb87:
@jquense That makes sense indeed. It is only required for older browsers like IE11 so extending the api for this is not a really good use case.

I have worked around the issue now by using an alias ('./TimeGridHeader') in my webpack config. I'll stick to that for now although it is not so clean :)

PR #1119, opened 2018-11-19
Fix drag cancelation

bs85:
- Handle escape while dragging
- Fix dropping an event out of bounds leaving the calendar in an intermediate state (no listener on `reset`)

### DIFF
--- a/src/Selection.js
+++ b/src/Selection.js
@@ -283,7 +283,7 @@ class Selection {
       return this.emit('reset')
     }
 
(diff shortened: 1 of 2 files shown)


MinhNguyen41092:
The third bug still happens in month view and when I drop event onto time gutter.

xtursky:
Hi guys, as mentioned 🔝 this problem still appears for month view. I prepared PR #1322 based on your's fix and tested it in my project. Could you please check? Thanks

PR #1123, opened 2018-11-21
improve date manipulation performance

bs85:
After debugging performance issues with a very heavily loaded calendar (~300 events per day across ~15 resources, 3 months of events passed in the `events` variable), I realized a couple functions were responsible for over 90% of the processing time.

This version is an order of magnitude faster for my use-case, going from ~900ms to < 70ms to render a heavily loaded week-view.

The main culprit I found was `date-arithmetic`, especially the `startOf` function, that while written very elegantly has absolutely horrible performance as it mutates the date many times.

While debugging it, I also realized `eventLevels.inRange` implementation was very costly, and could be sped up 10-20x by changing the input params a bit (passing the real range we want to filter and just using `<` and `>`).

I also added a autoDetectAllDayEvents prop to the calendar, that is true by default. Changing it to false will stop analysing every event time to see if both start and end are 00:00:00.

This works theoretically and I've tested it with my dataset, but there might be corner cases I'm not thinking of (zero time events combined with something else, daylight savivng, etc?).

Thanks in advance for having an in-depth look!

### DIFF
--- a/src/TimeGrid.js
+++ b/src/TimeGrid.js
@@ -29,6 +29,7 @@ export default class TimeGrid extends Component {
 
     scrollToTime: PropTypes.instanceOf(Date),
     showMultiDayTimes: PropTypes.bool,
+    autoDetectAllDayEvents: PropTypes.bool,
 
     rtl: PropTypes.bool,
     width: PropTypes.number,
@@ -94,7 +95,7 @@ export default class TimeGrid extends Component {
     raf.cancel(this.rafHandle)
     this.rafHandle = raf(this.checkOverflow)
   }
-  
+
   componentWillUnmount() {
     window.removeEventListener('resize', this.handleResize)
 
@@ -145,7 +146,7 @@ export default class TimeGrid extends Component {
     let { min, max, components, accessors, localizer } = this.props
 
     const resources = this.memoizedResources(this.props.resources, accessors)
-    const groupedEvents = resources.groupEvents(events) 
+    const groupedEvents = resources.groupEvents(events)
 
     return resources.map(([id, resource], i) =>
       range.map((date, jj) => {
@@ -191,6 +192,7 @@ export default class TimeGrid extends Component {
       min,
       max,
       showMultiDayTimes,
+      autoDetectAllDayEvents,
       longPressThreshold,
     } = this.props
 
@@ -204,15 +206,18 @@ export default class TimeGrid extends Component {
     let allDayEvents = [],
       rangeEvents = []
 
+    const dayAfterEnd = dates.dayAfter(end)
+
     events.forEach(event => {
-      if (inRange(event, start, end, accessors)) {
+      if (inRange(event, start, dayAfterEnd, accessors)) {
         let eStart = accessors.start(event),
           eEnd = accessors.end(event)
-
         if (
           accessors.allDay(event) ||
-          (dates.isJustDate(eStart) && dates.isJustDate(eEnd)) ||
-          (!showMultiDayTimes && !dates.eq(eStart, eEnd, 'day'))
+          (autoDetectAllDayEvents &&
+            dates.isJustDate(eStart) &&
+            dates.isJustDate(eEnd)) ||
+          (!showMultiDayTimes && !dates.eqDate(eStart, eEnd))
         ) {
           allDayEvents.push(event)
         } else {
@@ -316,5 +321,7 @@ export default class TimeGrid extends Component {
     }
   }
 
-  memoizedResources = memoize((resources, accessors) => Resources(resources, accessors))
+  memoizedResources = memoize((resources, accessors) =>
+    Resources(resources, accessors)
+  )
 }
--- a/src/utils/dates.js
+++ b/src/utils/dates.js
@@ -83,6 +83,14 @@ let dates = {
     )
   },
 
+  eqDate(date, otherDate) {
+    return (
+      date.getDate() === otherDate.getDate() &&
+      date.getMonth() === otherDate.getMonth() &&
+      date.getFullYear() === otherDate.getFullYear()
+    )
+  },
+
   isJustDate(date) {
     return (
       dates.hours(date) === 0 &&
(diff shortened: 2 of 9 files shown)


src/utils/eventLevels.js
```diff
@@ -55,17 +55,14 @@ export function eventLevels(rowSegments, limit = Infinity) {
...
-export function inRange(e, start, end, accessors) {
-  let eStart = dates.startOf(accessors.start(e), 'day')
-  let eEnd = accessors.end(e)
+export function inRange(e, lowerBound, upperBound, accessors) {
+  const start = accessors.start(e)
+  const end = accessors.end(e)
 
-  let startsBeforeEnd = dates.lte(eStart, end, 'day')
-  // when the event is zero duration we need to handle a bit differently
-  let endsAfterStart = !dates.eq(eStart, eEnd, 'minutes')
```
jquense: this line was preventing a ton of edge case bugs, i'm hesitant to just remove it without understanding how the new logic is equivalent.

jquense: also probably fine to do a straight getTime() comparison vs minutes

src/utils/dates.js
```diff
@@ -145,12 +153,18 @@ let dates = {
     return dates.startOf(new Date(), 'day')
   },
 
-  yesterday() {
-    return dates.add(dates.startOf(new Date(), 'day'), -1, 'day')
+  tomorrow(date) {
```
jquense: how about `dayAfter` for a name. tomorrow to me always means the day after today :P

src/utils/dates.js
```diff
@@ -145,12 +153,18 @@ let dates = {
...
+    const tomorrow = new Date(date.valueOf())
+    tomorrow.setDate(tomorrow.getDate() + 1)
+
+    return tomorrow
   },
 
-  tomorrow() {
-    return dates.add(dates.startOf(new Date(), 'day'), 1, 'day')
+  yesterday(date) {
```
jquense: `dayBefore`...are there any existing usages of this that would have broke?

src/TimeGrid.js
```diff
@@ -203,15 +205,19 @@ export default class TimeGrid extends Component {
     let allDayEvents = [],
       rangeEvents = []
 
+    const upperBound = new Date(end.valueOf())
```
jquense: `dates.tomorrow`?

src/utils/dates.js
```diff
@@ -145,12 +153,18 @@ let dates = {
     return dates.startOf(new Date(), 'day')
   },
 
-  yesterday() {
-    return dates.add(dates.startOf(new Date(), 'day'), -1, 'day')
+  tomorrow(date) {
+    const tomorrow = new Date(date.valueOf())
```
jquense: can we use `getTime()` or `+date` instead of `valueOf`

src/utils/eventLevels.js
```diff
@@ -55,17 +55,14 @@ export function eventLevels(rowSegments, limit = Infinity) {
   return { levels, extra }
 }
 
-export function inRange(e, start, end, accessors) {
-  let eStart = dates.startOf(accessors.start(e), 'day')
-  let eEnd = accessors.end(e)
+export function inRange(e, lowerBound, upperBound, accessors) {
```
jquense: can you add a comment noting the performance concerns of this method as an explanation for why it does use `date-arithmetic`, so someone doesn't accidentally remove the code later :P

jquense:
this is awesome thanks 👍

bs85:
Thanks a lot for the comments, will cleanup style and add comments, and deal with the failing tests ASAP

jquense:
any chance to clean this up @bs85 ?

bs85:
Not yet sorry. Do you want to close it and re-open when I can get to it? Might not be until a couple more weeks.

jquense:
I'm happy to leave it, I just didn't want to close if your planning on getting to it. no rush of course

src/utils/eventLevels.js
bs85: Changed my implementation slightly. upperBound is now exclusive, and in case the event start is equal to upperBound, we check if the event has a zero duration, to ensure that the function always returns true if the range and the event have the exact same bounds.

Could you say more about those edge cases? This function is only called in 3 locations (in Month.js, Agenda.js and TimeGrid.js). Other parts of the code call the `inRange` method of `utils/date`.

src/utils/dates.js
bs85: Done

src/TimeGrid.js
bs85: Done

src/utils/eventLevels.js
bs85: Commented the inRange method

src/utils/dates.js
bs85: using `getTime` everywhere now

src/utils/dates.js
bs85: It wasn't used anywhere, the part that used it was probably changed

bs85:
Found some time to clean it up!

As the `inRange` method's behavior changed, I also rewrote the tests based on the assumptions of the new version.

I think `eventLevels.inRange` could reasonably be considered a private method, but in practice some projects might be importing it and using it, especially in custom components.

I overwrote it in this version, but if you think it is too dangerous, I can also create a new `fastInRange` method, and deprecate `inRange` for now but keep it.

test/utils/eventLevels.test.js
```diff
@@ -261,107 +261,71 @@ describe('inRange', () => {
         expect(inRange(event, rangeStart, rangeEnd, accessors)).toBe(result)
       })
     }
-    const weekOfThe5th = [d(5), d(11)]
-    const weekOfThe12th = [d(12), d(18)]
-    ;[
+
+    const sample = [
```
bs85: prettier was prepending a semi-colon here, and eslint was complaining. Assigning to `sample` as a workaround

src/utils/eventLevels.js
jquense: > Could you say more about those edge cases? 

There are a few tests that cover them, but it mostly comes down to events that span weeks or sit on the leading/trailing border of a week. If you don't account for zero duration events specifically they can get dropped from the week or show up at the start and end. If tests are passing tho, that should indicate the case is covered (should)

PR #1134, opened 2018-12-02
changes to allow calendar to work with React 15.6.2

iPhatty:
Hey all, came across an issue when upgrading to the latest calendar (v0.20.1). It's using `React.createRef()`, but my app is currently on React 15.6.2 which does not have the API. The latest calendar had features which I really wanted, but I currently do not have the time to upgrade the app, and so I went through and just made some small changes so that it is no longer using the createRef.

### DIFF
--- a/src/Agenda.js
+++ b/src/Agenda.js
@@ -49,7 +49,7 @@ class Agenda extends React.Component {
     return (
       <div className="rbc-agenda-view">
         {events.length !== 0 ? (
-          <React.Fragment>
+          <div>
             <table ref="header" className="rbc-agenda-table">
               <thead>
                 <tr>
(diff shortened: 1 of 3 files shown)


src/Agenda.js
```diff
@@ -70,7 +70,7 @@ class Agenda extends React.Component {
                 </tbody>
               </table>
             </div>
-          </React.Fragment>
+          </div>
```
jquense: this isn't a free change, it adds markup that wasn't there before, which i'm not super comfortable with since that likely will break ppl's apps and cause unclear regressions

src/addons/dragAndDrop/EventContainerWrapper.js
```diff
@@ -224,7 +224,7 @@ class EventContainerWrapper extends React.Component {
               continuesLater={startsAfterDay}
             />
           )}
-        </React.Fragment>
+        </div>
```
jquense: especiially here

jquense:
Thanks for the PR, I think the right approach at this point is to only support v16+ up. It's need out for a long time and while i understand it's hard to stay up to date in some apps, it's hard to maintain both versions well and the maintainer's time is currently very limited

PR #1140, opened 2018-12-04
Added fallback to getNow in `handleViewChange`

valeryq:
When view changed and props `date === undefined` we have unexpected behavior for `handleRangeChange` and then for `onRangeChange` methods. It returns sometimes `undefined` and sometimes `Invalid Date`, so I suggest fix for this to provide default date if `date` prop isn't present.

@jquense

### DIFF
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -952,7 +952,7 @@ class Calendar extends React.Component {
     }
 
     let views = this.getViews()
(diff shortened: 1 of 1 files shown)


jquense:
thanks!

valeryq:
@jquense when you plan to release with this fix?

PR #1148, opened 2018-12-11
feat: implement enabledDays prop

bs85:


src/Day.js
```diff
@@ -22,13 +22,45 @@ Day.range = date => {
   return [dates.startOf(date, 'day')]
 }
 
-Day.navigate = (date, action) => {
+Day.navigate = (date, action, { enabledDays }) => {
+  // enabledDays being non-empty but containing invalid values would lead to
+  // an infinite loop, throw instead if the loop goes on for too long
```
jquense: why not validate `enabledDays` instead of the loop bailout?

src/Calendar.js
```diff
@@ -86,6 +86,19 @@ class Calendar extends React.Component {
...
+     *
+     * The values should indicate days of the week according to Date.prototype.getDay(),
+     * e.g. to show Tue - Fri, the array should be [2, 3, 4, 5].
+     *
+     * When navigating with next and previous, disabled days are skipped
+     *
+     * @type number[]
+     * @default null
+     */
+    enabledDays: PropTypes.arrayOf(PropTypes.number),
```
jquense: this essentially has to be a contiguous array, since month view can't handle holes in a week. why not make this a range e.g. start/end like we do already in timeGrid

src/Week.js
```diff
@@ -32,12 +32,16 @@ Week.navigate = (date, action) => {
...
 }
 
-Week.range = (date, { localizer }) => {
+Week.range = (date, { localizer, enabledDays }) => {
   let firstOfWeek = localizer.startOfWeek()
   let start = dates.startOf(date, 'week', firstOfWeek)
   let end = dates.endOf(date, 'week', firstOfWeek)
 
-  return dates.range(start, end)
+  return enabledDays
```
jquense: not sure what to do with the WorkWeek view give n this now...

PR #1149, opened 2018-12-11
feat: add an alternate layout algorithm, allow using a custom one

bs85:
Provides a new algorithm that doesn't stack events, instead packing them similarly to fullcalendar, but with constant width.

Also enables passing a custom algorithm through the `positionedEventsGetter` prop.

This is a WiP, I'm opening a pull request so you guys can have a look. There are no tests for the new layout, I'll get to that when I have some time.

### DIFF
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -17,6 +17,7 @@ import { mergeWithDefaults } from './localizer'
 import message from './utils/messages'
 import moveDate from './utils/move'
 import VIEWS from './Views'
+import LAYOUTS from './Layouts'
 import Toolbar from './Toolbar'
 import NoopWrapper from './NoopWrapper'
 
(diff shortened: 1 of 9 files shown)


jquense:
Nice! I'll try and give it a closer look a bit later. Generally tho seems good

src/Calendar.js
```diff
@@ -17,6 +17,7 @@ import { mergeWithDefaults } from './localizer'
 import message from './utils/messages'
 import moveDate from './utils/move'
 import VIEWS from './Views'
+import LAYOUTS from './Layouts'
```
jquense: I think as a point of structure we should try and put the new layout in an add-on, to better ensure the pluggable API is clean, and also so we don't have to include code for both in every app

PR #1161, opened 2018-12-20
feat: Add background events 

rrjoson:
**Summary of Changes**
- Add a background events feature similar to https://fullcalendar.io/docs/background-events

**Notes**
- This is based on https://github.com/intljusticemission/react-big-calendar/pull/1085#issuecomment-436610463

**Screenshots**
- screen shot 2018-12-20 at 7 39 30 pm

### DIFF
--- a/examples/App.js
+++ b/examples/App.js
@@ -16,6 +16,7 @@ import './prism.less'
 import Card from './Card'
 import ExampleControlSlot from './ExampleControlSlot'
 import Basic from './demos/basic'
+import BackgroundEvents from './demos/backgroundEvents'
 import Selectable from './demos/selectable'
 import Cultures from './demos/cultures'
 import Popup from './demos/popup'
@@ -35,6 +36,7 @@ let demoRoot =
 
 const EXAMPLES = {
   basic: 'Basic Calendar',
+  backgroundEvents: 'Background Events',
   selectable: 'Create events',
   cultures: 'Localization',
   popup: 'Show more via a popup',
@@ -69,6 +71,7 @@ class Example extends React.Component {
     let selected = this.state.selected
     let Current = {
       basic: Basic,
+      backgroundEvents: BackgroundEvents,
       selectable: Selectable,
       cultures: Cultures,
       popup: Popup,
--- a/examples/backgroundEvents.js
+++ b/examples/backgroundEvents.js
@@ -1,0 +1,8 @@
+export default [
+  {
+    id: 0,
+    title: 'Background Event 1',
+    start: new Date(2015, 3, 13, 5, 0, 0),
+    end: new Date(2015, 3, 13, 16, 0, 0),
+  },
+]
--- a/examples/backgroundEvents.js
+++ b/examples/backgroundEvents.js
@@ -1,7 +1,6 @@
 export default [
   {
     id: 0,
-    title: 'Background Event 1',
     start: new Date(2015, 3, 13, 5, 0, 0),
     end: new Date(2015, 3, 13, 16, 0, 0),
   },
--- a/examples/demos/backgroundEvents.js
+++ b/examples/demos/backgroundEvents.js
@@ -1,0 +1,22 @@
+import React from 'react'
+import BigCalendar from 'react-big-calendar'
+import events from '../events'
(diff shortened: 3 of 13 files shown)


jquense:
Nice! I think this is a good start, in terms of API and internal separation from the "real" events. I think it may be helpful before continuing on an implementation to do some work speccing out a few things:

- What use-cases should feature enable?
- In what specific ways are bg events different from plain events (this helps us determine the api, and solve questions like "should drag and drop work on them")

at that point i think we can DRY up places where there is a lot of overlap and separate more the points in which they differ.

how's that sound?

rrjoson:
Sounds good! Thanks for the feedback! 

I'll spec out here first for feedback the use cases that the bg events feature enables and then specify how the bg events are different from plain events.

After that, I'll proceed with: 
- DRY up the code 
- Separate the points in which they differ
- Continue with other changes to the API like "Drag and drop" feature on the bg events, etc.

rrjoson:
Hi @jquense. Let me know your thoughts about this when you're free. Thanks!

**In which use-cases should we enable this feature?**

- As a user, I want to only be able to schedule an event within certain time slots
- As a user, I want to emphasize or highlight certain time slots on the calendar view 
- As a user, I want to emphasize certain days in the calendar month view
- As a user, I want to show which slots are available to book an event with
- As a user, I want to add labels to certain days or timeslots 

**In what specific ways are bg events different from plain events?**

- Background events are read only "events" as these are only used to highlight or add labels to time slots.
- Background events can have some kind of constraint that allows us to only create or drop events inside these background events (See: https://github.com/fullcalendar/fullcalendar/blob/master/demos/background-events.html#L33)

jquense:
looks good @rrjoson I do wonder if we need the same layout mechanism as normal events for these as well. Do you know of any prior art aside from Full calendar we can look at for UX/UI on that front?

rrjoson:
@jquense sorry, I don't know any. Actually I was thinking of just copying the background event UI from Full calendar. What do you think?

image

jquense:
looks great 👍

rohit-sourcefuse:
I am also waiting for this pull request. Any update on this?

rrjoson:
Hi @rohit-sourcefuse 

Sorry wasn't able to continue on this. Will try to finish it within the following week.

rrjoson:
Hi @jquense 

When you're free, can I have a quick feedback on my progress here?

Thanks!

crosscompile:
@jquense anything I can do to help you out here?  I'm creating a fork to use these changes, but would love to help merge into the official repo.

nthgol:
we've also forked this PR and merged in master.
Our main use case is showing free/busy, read-only events pulled from a 3rd party calendar API overlaying the user's ability to select.
Here's an unstyled example. 
I support reopening this and adding this functionality.
image

sgraham3311:
This is a good PR and should be merged. Can we get an update on this @jquense

jquense:
Oops sorry this got lost in time. GH is bad at this and I'm not a good maintainer. If anyone wants to pick this and get it updated I'll merge it

sgraham3311:
@jquense 

After doing some work/research I was able to achieve the desired result with the latest version of react-big-calendar. I essentially just provided a custom `dayLayoutAlgorithm` giving the background events their own styled events.
```javascript
import nooverlap from 'react-big-calendar/lib/utils/layout-algorithms/no-overlap';

export default ({ events, minimumStartDifference, slotMetrics, accessors }) => [
	...nooverlap({
		events: events.filter((e) => e.type === 'background'),
		minimumStartDifference,
		slotMetrics,
		accessors
	}),
	...nooverlap({
		events: events.filter((e) => e.type !== 'background'),
		minimumStartDifference,
		slotMetrics,
		accessors
	})
];
```
By using the `eventPropGetter` I am able to apply styles to the background events and most importantly set `pointer-events: none` so they don't receive click events.

You can end up with something like this
Screen Shot 2020-05-15 at 1 00 53 PM

KhaledMohamedP:
How does this implementation handles overlapping background events?

jmauzy:
I've picked this up in PR #1727

Issue #1007, opened 2018-09-25
Event resizing


McDoyen:
#### Do you want to request a _feature_ or report a _bug_?
_bug_

#### What's the current behavior?
Once an event has been resized form one slot event to cover more than two slots, its not possible to resize it back to one slot.

![3](https://user-images.githubusercontent.com/19818344/46015651-7d752480-c0db-11e8-985c-c38a01349a4d.gif)

react big calendar v 0.20.1

jquense:
what are the date values of the event starting, after resize, and as you try and set it back. I believe this is a sort of unfortunate quirk of date math (the worst)

McDoyen:
@jquense After resizing back to 12th, some how the end date is 14th 

![3](https://user-images.githubusercontent.com/19818344/46061966-fecac680-c170-11e8-8b3e-15e2cce13bf9.jpeg)

jquense:
Can you post a codesandbox with the one event? I really need to see the exact dates, it'll be easier if I can run something :thumbsup:

McDoyen:
Was unable able to create the codesandbox, but I have just realized the same bug exists on the demo project http://intljusticemission.github.io/react-big-calendar/examples/index.html#dnd

liben96:
Any updates on this bug?

morrisgwilliam:
has this issue been resolved? I am having the same bug. Cannot shrink tthe event to one day unless the start date (left drag) has been modified. then the event is allowed to reduce to one day.

PR #1169, opened 2019-01-08
fix allow  bar to resize till start place

Julious1994:
Hi,
I also got below issue.
https://github.com/intljusticemission/react-big-calendar/issues/1007
I tried to solve it.
you can accept pull request if it is right for this code.

Thanks
Amit Patel

### DIFF
--- a/src/addons/dragAndDrop/WeekWrapper.js
+++ b/src/addons/dragAndDrop/WeekWrapper.js
@@ -136,7 +136,7 @@ class WeekWrapper extends React.Component {
         return
       }
 
-      end = dates.max(end, dates.add(start, 1, 'day'))
+      end = dates.max(end, dates.add(start, 0, 'day'))
     } else if (direction === 'LEFT') {
       // inbetween Row
       if (cursorInRow) {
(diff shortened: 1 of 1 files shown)


src/addons/dragAndDrop/WeekWrapper.js
```diff
@@ -155,7 +155,7 @@ class WeekWrapper extends React.Component {
         return
       }
 
-      start = dates.min(dates.add(end, -1, 'day'), start)
+      start = dates.min(dates.add(end, 0, 'day'), start)
```
jquense: i think this is the same as `end`? does adding nothing have an effect here?

Julious1994: when we decrease a bar that time it will make possible to decrease till single day bar (same as start day).

Julious1994: yes it is same as end. When we want to set bar for one day that time start and end will be same i think.
So review it and accept if it is ok.

jquense: if its the same, lets remove the `dates.add()` and just use `end` same for the line above with `start`

Julious1994: yes we can. If we can create new branch or something like for beta/alpha version so main version would not affect. May be we can get an issue after this change.

morrisgwilliam: has this issue been resolved? I am having the same bug. Cannot shring the event to one day unless the start date (left drag) has been modified. then the event is allowed to reduce to one day.

PR #1177, opened 2019-01-14
Bump react peer dependency to 16.3

dpikt:
[TimeGrid](https://github.com/intljusticemission/react-big-calendar/blob/master/src/TimeGrid.js#L68) relies on `React.createRef`, which was added in React v16.3.


jquense:
thanks!

Issue #822, opened 2018-04-26
Agenda headers misaligned in Chrome on initial agenda load


bmueller-sykes:
#### Do you want to request a _feature_ or report a _bug_?
Bug
#### What's the current behavior?
Go here in Chrome (on macOS):
http://intljusticemission.github.io/react-big-calendar/examples/index.html

...by default, the calendar appears on month view (which is fine). Click "agenda", and the agenda table head is misaligned with the body. As soon as the browser is resized, the head snaps into the correct location.
<img width="767" alt="screen shot 2018-04-26 at 12 49 05 pm" src="https://user-images.githubusercontent.com/12236379/39322659-399b604c-4950-11e8-8221-43d8aa4a1d07.png">

Safari and FF on macOS appear to work fine. MS Edge also works fine. (The library appears incompatible with IE11?)

#### What's the expected behavior?
Headers should line up with agenda body at all times.

dgana:
Yes im having the same problem so what is the quick fix for this

dgana:
`
viewChange = view => {
    if (view === "agenda")
      setTimeout(() => {
        this.setState(
          {
            width: window.innerWidth - 1
          },
          () =>
            this.setState({
              width: window.innerWidth + 1
            })
        );
      }, 1);
  };

  render() {
    return (
            <BigCalendar
              onView={this.viewChange}
            />
    );
  }
`

My quick fix is use set width and set it to style in render for 1ms and turn it back

Soundvessel:
This also occurs when you get a view with the empty agenda text then navigate back into an agenda with events.

laradevitt:
This is still an issue. I'm seeing the behaviour as described by @bmueller-sykes and @Soundvessel. My workaround was to wrap component in a conditional to slightly delay rendering.

PR #1181, opened 2019-01-16
Fixed issue with gutter width initialization

janb87:
Reproduce the original issue (in Chrome):

1. Directly go to http://intljusticemission.github.io/react-big-calendar/examples/index.html#selectable
2. Notice that you cannot see the day headers because the width of the gutter is set way too big

By wrapping it in the requestAnimationFrame it ensures that it is set after the initial paint is finished (which does the css etc).


janb87:
Relates to #822

jquense:
thanks!

tukhfatov:
Hi gents,
Thanks @janb87. 
@jquense when are you going to release this fix?

PR #1183, opened 2019-01-18
Customize render of day number

Yadro:
Provide a function that returns an object of className or style props to be applied to the number of day
to wrapper of DateHeaderComponent.

2019-01-18-11-31-react app

### DIFF
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -497,6 +497,15 @@ class Calendar extends React.Component {
      */
     dayPropGetter: PropTypes.func,
 
+    /**
+     * Optionally provide a function that returns an object of className or style props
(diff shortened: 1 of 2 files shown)


jquense:
I think `DateHeaderComponent` covers this case already? It's probably good to not have two ways of doing something if we can help it.

Yadro:
> I think `DateHeaderComponent` covers this case already? It's probably good to not have two ways of doing something if we can help it.

Yep, you are right. I missed that we can override `DateHeaderComponent`. Thank you!
ps [#prop-components](http://intljusticemission.github.io/react-big-calendar/examples/index.html#prop-components) not actual :(

Issue #1174, opened 2019-01-10
[Feature Request] Custom Resource Header


healthycola:
I'm looking to add custom header for resources. I found that it's pretty much hard coded in [here](https://github.com/intljusticemission/react-big-calendar/blob/5857d8fb44d614047d43838be8c52bbbe2dbff55/src/TimeGridHeader.js#L172). I'm not sure if anyone is already working on this, if not, would it be okay if I created a PR to enable custom headers for resources?

Thanks!

lksilva:
This feature has already added in last version but don't exists in @types.

"@types/react-big-calendar": "^0.20.10",
"react-big-calendar": "^0.20.4",

```
export interface Components<TEvent extends Event = Event> {
    event?: React.ComponentType<EventProps<TEvent>>;
    eventWrapper?: React.ComponentType<EventWrapperProps<TEvent>>;
    eventContainerWrapper?: React.SFC | React.Component | React.ComponentClass | JSX.Element;
    dayWrapper?: React.SFC | React.Component | React.ComponentClass | JSX.Element;
    dateCellWrapper?: React.SFC | React.Component | React.ComponentClass | JSX.Element;
    timeSlotWrapper?: React.SFC | React.Component | React.ComponentClass | JSX.Element;
    timeGutterHeader?: React.SFC | React.Component | React.ComponentClass | JSX.Element;
    timeGutterWrapper?: React.SFC | React.Component | React.ComponentClass | JSX.Element;
    toolbar?: React.ComponentType<ToolbarProps>;
    agenda?: {
        date?: React.SFC | React.Component | React.ComponentClass | JSX.Element;
        time?: React.SFC | React.Component | React.ComponentClass | JSX.Element;
        event?: React.SFC | React.Component | React.ComponentClass | JSX.Element;
    };
    day?: {
        header?: React.SFC | React.Component | React.ComponentClass | JSX.Element;
        event?: React.SFC | React.Component | React.ComponentClass | JSX.Element;
    };
    week?: {
        header?: React.SFC | React.Component | React.ComponentClass | JSX.Element;
        event?: React.SFC | React.Component | React.ComponentClass | JSX.Element;
    };
    month?: {
        header?: React.SFC | React.Component | React.ComponentClass | JSX.Element;
        dateHeader?: React.SFC | React.Component | React.ComponentClass | JSX.Element;
        event?: React.SFC | React.Component | React.ComponentClass | JSX.Element;
    };
    /**
[...]

PR #1187, opened 2019-01-21
Add ability to set custom resource headers

healthycola:
Add an ability to set a custom resource headers. For #1174

### DIFF
--- a/src/ResourceHeader.js
+++ b/src/ResourceHeader.js
@@ -2,7 +2,7 @@ import PropTypes from 'prop-types'
 import React from 'react'
 
 const ResourceHeader = ({ label }) => {
-  return <span>{label}</span>
+  return <React.Fragment>{label}</React.Fragment>
 }
 
 ResourceHeader.propTypes = {
--- a/src/ResourceHeader.js
+++ b/src/ResourceHeader.js
@@ -1,0 +1,20 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+
+const ResourceHeader = ({ label, index }) => {
+  return (
+    <div className="rbc-row rbc-row-resource">
+      <div key={`resource_${index}`} className="rbc-header">
+        {label}
+      </div>
+    </div>
+  )
+}
+
+ResourceHeader.propTypes = {
+  label: PropTypes.node,
+  index: PropTypes.number,
+  resouce: PropTypes.object,
+}
+
+export default ResourceHeader
--- a/src/ResourceHeader.js
+++ b/src/ResourceHeader.js
@@ -14,7 +14,7 @@ const ResourceHeader = ({ label, index }) => {
 ResourceHeader.propTypes = {
   label: PropTypes.node,
   index: PropTypes.number,
-  resouce: PropTypes.object,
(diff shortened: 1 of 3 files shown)


src/ResourceHeader.js
```diff
@@ -0,0 +1,20 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+
+const ResourceHeader = ({ label, index }) => {
+  return (
+    <div className="rbc-row rbc-row-resource">
+      <div key={`resource_${index}`} className="rbc-header">
+        {label}
```
jquense: the Header component should be just for the `{label}` part otherwise specifying a component would break the styles

src/TimeGridHeader.js
```diff
@@ -142,11 +146,11 @@ class TimeGridHeader extends React.Component {
...
             {resource && (
-              <div className="rbc-row rbc-row-resource">
-                <div key={`resource_${idx}`} className="rbc-header">
-                  {accessors.resourceTitle(resource)}
-                </div>
-              </div>
+              <ResourceHeaderComponent
+                index={idx}
+                label={accessors.resourceTitle(resource)}
+                resource={resource}
```
jquense: needs a `key`

src/ResourceHeader.js
```diff
@@ -0,0 +1,14 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+
+const ResourceHeader = ({ label }) => {
+  return <span>{label}</span>
```
jquense: should be ok to do `<>{label}</>` and not add an extra span

src/TimeGridHeader.js
```diff
@@ -144,7 +148,12 @@ class TimeGridHeader extends React.Component {
             {resource && (
               <div className="rbc-row rbc-row-resource">
                 <div key={`resource_${idx}`} className="rbc-header">
```
jquense: err looks liek there was an existing bug here, can you move the `key` to the correct dive (the one right above this one)

src/less/time-grid.less
```diff
@@ -94,6 +94,11 @@
     border-left-width: 0;
     border-right: 1px solid @cell-border;
   }
+
+  >.rbc-row.rbc-row-resource {
+      border-bottom: 1px solid @cell-border;
+      flex-shrink: 0;
```
jquense: why is a new class needed? This change shouldn't break be changing any existing markup right?

healthycola: I added this in because otherwise, the new resource headers aren't given flexibility in terms of how they are rendered.

Using a custom header, this is what I see.
With the change:
screen shot 2019-01-22 at 8 48 20 am

Without the chance:
screen shot 2019-01-22 at 8 48 44 am

In fact, I think the I think the `> .rbc-row.rbc-row-resource {` that's under the `.rbc-time-header {` is not doing anything (as far as i could see anyway). But I left it there because I didn't want to break anything.

src/ResourceHeader.js
healthycola: sure.

jquense:
thanks!

PR #1195, opened 2019-01-29
Remove global window from code called during server-side rendering

sedenardi:
When I render RBC on the server with the `week` view, it produces an error with the following stack trace:

```
ReferenceError: window is not defined
    at Object.groupEvents (/[project path]/node_modules/react-big-calendar/lib/utils/Resources.js:18:30)
    at TimeGrid.renderEvents (/[project path]/node_modules/react-big-calendar/lib/TimeGrid.js:166:35)
    at TimeGrid.render (/[project path]/node_modules/react-big-calendar/lib/TimeGrid.js:263:14)
    at processChild (/[project path]/node_modules/react-dom/cjs/react-dom-server.node.development.js:2863:18)
```
This is due to using `window.Map()` instead of simply `Map()` in `utils/Resources.js`. While all node versions going back to v0.12 support the ES6 `Map`, node has no `window`, hence the break. 

All tests pass with the change, and to satisfy the eslint `global-require` rule I've added the `es6` environment.

Please let me know if you've got any questions/comments/concerns. Thanks for all your hard work on this excellent library.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,7 +2,8 @@ parser: babel-eslint
 extends:
   - jason/react
   - prettier
-
+env:
+  es6: true
 rules:
   global-require: off
   no-unused-vars: [2, {
--- a/src/utils/Resources.js
+++ b/src/utils/Resources.js
@@ -10,7 +10,7 @@ export default function Resources(resources, accessors) {
     },
 
     groupEvents(events) {
-      const eventsByResource = new window.Map()
+      const eventsByResource = new Map()
 
       if (!resources) {
         // Return all events if resources are not provided


sedenardi:
Additionally, you'll probably notice I didn't remove _all_ `window` declarations. This is because all other calls happen in React lifecycle methods that are never called during server-side rendering. While they can be safely removed, I wanted the surface area of this change as small as possible.

sedenardi:
Thank you @jquense! Do you know if it's possible to get a notification when this is merged and pushed to npm? No rush, just don't want to miss it.

adam-26:
@jquense I am also experiencing this, would really appreciate it if you could publish a new version with this fix, thanks :)

sedenardi:
Hi @jquense, any sense on when you're planning on publishing a new version? 

Really appreciate your time. Thanks again.

PR #1197, opened 2019-01-30
Provide esm support for bundlers

TrySound:
In this diff I added esm bundle which will allow user bundlers
to produce much smaller code. Currently regular babel output is
wrapped with module iife.

This project has a lot of files so with single esm file users will
bundle it slightly faster.

As a bonus we see the size of the project without dependencies.


jquense:
thanks!

TrySound:
@jquense Would be good to see this released. What is the plan?

PR #1202, opened 2019-01-31
Add resourceHeader to Proptypes

healthycola:


healthycola:
@jquense would you mind reviewing. also any idea when the next release might be?

Issue #1203, opened 2019-02-01
Appointments scheduled for 11:59pm appear outside calendar boundary


NathanBP:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

Appointments scheduled to start at 11:59pm and end in the next day will appear outside the boundary of the calendar.

Code Sandbox:
[![Edit n3nn2492np](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/n3nn2492np)

Screenshot:
![image](https://user-images.githubusercontent.com/37284011/52101204-51882380-2626-11e9-8e4b-0cf89bbf1b7e.png)

#### What's the expected behavior?

Appointments that start at 11:59pm should be contained within the calendar boundary as shown in the screenshot below.

![image](https://user-images.githubusercontent.com/37284011/52101247-71b7e280-2626-11e9-9593-de9ded639b3d.png)

NathanBP:
Pull request is still outstanding. Comment to keep issue open

PR #1204, opened 2019-02-01
fix: bug where appointments can appear outside the calendar

NathanBP:
- Fixes bug #1203. Appointments set for 11:59pm now stay inside the calendar boundary
- Added the dndresource demo to the demo drop down so that it can be viewed. 
- Updated resources in the dndresource demo to show bug has been fixed.


src/utils/TimeSlots.js
```diff
@@ -130,7 +130,10 @@ export function getSlotMetrics({ min: start, max: end, step, timeslots }) {
 
       const rangeStartMin = positionFromDate(rangeStart)
       const rangeEndMin = positionFromDate(rangeEnd)
-      const top = (rangeStartMin / (step * numSlots)) * 100
+      const top =
+        rangeStartMin === rangeEndMin
+          ? (rangeStartMin / (step * numSlots)) * 98
```
jquense: what's going on here with `98`, is there a deterministic reason to use 98%? This feels like a number that would only work with the default slot numbers.

NathanBP: Yes you are correct after further testing with smaller step values it does make the events not sit in the correct space. I have modified the code for this section to factor in the step value so as to allow this to now work correct for any step value.

test/utils/TimeSlots.test.js
```diff
@@ -22,3 +22,81 @@ describe('getSlotMetrics', () => {
...
+    expect(range.top + range.height).toBeLessThan(100)
+    expect(range.height).toBeGreaterThan(0)
+  })
+
+  test('getRange: multi day appointment stays within calendar', () => {
+    let range = slotMetrics.getRange(
+      new Date(2018, 0, 29, 0, 0, 0),
+      new Date(2018, 0, 30, 4, 0, 0)
+    )
+    expect(range.top + range.height).toBeLessThan(100)
```
jquense: Should this be equal to 100?

NathanBP: I believe so, as there is code in the getRange that selects the max value between the start of the day and the start of the appointment. Along with the start of the day and the end of the appointment. My understanding is that this is what splits a multi day appointment over separate days in the week view when the `showMultiDayTimes` prop is set. 

I might be wrong though as this is just my understanding of how the calendar displays appointments

NathanBP:
> Looks great, is it worth adding a test covering a start date before the time slot range as well?

I thought this scenario would have been covered by having a test that is from 00:00 to 00:15 and the multi day test? I can add another test if you'd like, if possible could you clarify further by giving an example date/time?

NathanBP:
Just a quick comment to keep this pull request open

jquense:
thanks!

PR #1225, opened 2019-02-19
sass version added

zilahir:
I have added `scss` version of the styles. 

The reason I did this, because I prefer `sass` over `less`, therefore the project I intended to use this component, is purely based on `sass`, and using `node-sass`.

So with this `pr` users will be able to do `scss` import, such as: 

`import './path/to/calendar/style.css`


cutterbl:
I feel bad, as I didn't see this and filed my own PR #1235...

Issue #1230, opened 2019-02-21
Feature Request: Fire onRangeChange On Component Render


cutterbl:
### Feature Request

There are many who have a large database of events for display, and it isn't always convenient to load all of them to the Scheduler at once. Using the `onRangeChange` property, it is possible to make ajax requests for this data based upon the `range` data it gets as an argument. This is especially helpful in the `month` view, as it also gives you the full range of displayed dates.

However, `onRangeChange` is only called, currently, during `handleNavigate` or `handleViewChange`. It would be highly beneficial if this was also called on mount.

cutterbl:
PR Created

PR #1231, opened 2019-02-21
Broadcast initial view range on component mount

cutterbl:
This addresses the FR made in Issues #1230. This will broadcast the initial view date range when the component mounts, if the developer has configured `onRangeChange`.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,25 +1,25 @@
 {
   "./dist/react-big-calendar.js": {
-    "bundled": 470853,
-    "minified": 153064,
-    "gzipped": 42298
+    "bundled": 455385,
+    "minified": 146083,
+    "gzipped": 41768
   },
   "./dist/react-big-calendar.min.js": {
-    "bundled": 413805,
-    "minified": 136004,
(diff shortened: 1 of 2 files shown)


jquense:
Thanks for the PR, The approach here has a few downsides and is generally not idiomatic of how controllable props work. This requires that you wait for the calendar to mount before starting to fetch events which isn't great, Instead it would be better to calculate the visible range apart from the calendar and use that. In this instance you really don't even need the specific range, it should be fine (and probably preferable) to overfetch events, e.g. for the the current month as well as months before and after, to reduce loading.

cutterbl:
Does this inclusion create issues or adverse side effects that would make it untenable? While I understand your opinion in this matter, isn't that really a decision to be made by the developer using Big Calendar? Adding a note to the `onRangeChange' documentation, stating that this is called when the component mounts, should be enough for any developer to plan how they treat this accordingly. 

Some architects prefer to only fetch the smallest amount of data required for a view to display (especially in mobile environments). This means that the developer doesn't have to run their own date calculations that are already being done by the component itself.

PR #1235, opened 2019-02-22
Provide SASS implementation of styles

cutterbl:
I created SASS versions of the styles. I was going to clean it up and make it more "SASS"y, but decided to keep exact same structure for parity with the LESS, making it easier to find and make changes to both when necessary.

I provided npm scripts for running a SASS builds (for testing), as well as copying the SASS files into the `lib` directory as part of the `build` process. I also figured out how to change the implementation in Storybook, for testing. I left those bits commented out after running my testing, so they can be easily interchanged when necessary.

I also provided a **Custom Styling** section to the README file, giving a rough outline on how a developer can include either the LESS or the SASS in their own application for overriding styles to match their own application styling, as well as a caveat warning about how their changes could adversely affect Big Calendar if not tested.


cutterbl:
@jquense I can handle the doc changes, and updates to the build steps (plus storybook). That said, don't you think there's value in giving the developer the option? I support multiple applications, personally, some using SASS and some using LESS. If you want to drop LESS all together, that's cool, just throwing it out there...

jquense:
I do like providing both, but it's too much work to maintain, and to much to ask contributors to provide styles twice, I think it's generally better to do one, and convert between them automatically or just provide one

cutterbl:
@jquense OK, I get it. I'll update the markdown, Storybook webpack, examples webpack, and the build steps. I'll also look through the Examples site to see if there are any ref's there to consider.

cutterbl:
@jquense I created a new PR (#1313 ) containing all of the changes from this PR, plus the removal of LESS, and additional updates. That PR is passing all checks, so I'm closing this PR, while referencing this PR in it's comments.

PR #1236, opened 2019-02-23
Replace classnames with clsx

TrySound:
A few points

- more performant
- smaller without iife
- esm support without hacky conversion to commonjs with module wrapper

Refs https://unpkg.com/classnames@2.2.6/index.js, https://twitter.com/MaterialUI/status/1092928715804299264


jquense:
If this is a problem with classnames wouldn't it make WAY more sense to PR to classnames instead of trying to boil the ocean in the react ecosystem to switch packages?

TrySound:
It would be good. The problem is that classnames package is stuck. So I just started migrating all packages to clsx. material-ui, material-ui-pickers and react-virtualzied are already migrated.
https://github.com/JedWatson/classnames/pull/150#issuecomment-407935829

TrySound:
Reactivate.

TrySound:
classnames package still is not active with this topic. `clsx` becomes more and more popular. @jquense what do you think?

TrySound:
Keep it please

TrySound:
@jquense If you ok to go with clsx I can help to migrate all your packages. Just give me a list of them.

PR #1239, opened 2019-02-25
Fix misplacement of current time indicator

selenir:
Hello,
It looked like PR for this issue (#1054) was abandoned and the bug still exists so I've decided to continue jdaberkow's work.
Also I've added check for `getNow` prop, because when you want to implement timezones you may have different `getNow` function depending on which timezone is selected. When user changes the timezone indicator should move to correct position for that timezone.

### DIFF
--- a/src/DayColumn.js
+++ b/src/DayColumn.js
@@ -44,12 +44,18 @@ class DayColumn extends React.Component {
   }
 
   componentDidUpdate(prevProps, prevState) {
-    if (prevProps.isNow !== this.props.isNow) {
+    const getNowChanged = !dates.eq(
+      prevProps.getNow(),
+      this.props.getNow(),
+      'minutes'
+    )
+
+    if (prevProps.isNow !== this.props.isNow || getNowChanged) {
       this.clearTimeIndicatorInterval()
 
       if (this.props.isNow) {
         const tail =
-          this.props.getNow === this.props.getNow &&
+          !getNowChanged &&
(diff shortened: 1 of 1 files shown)


src/DayColumn.js
```diff
@@ -48,9 +48,12 @@ class DayColumn extends React.Component {
       this.clearTimeIndicatorInterval()
 
       if (this.props.isNow) {
-        this.setTimeIndicatorPositionUpdateInterval(
+        const tail =
+          this.props.getNow === this.props.getNow &&
```
jquense: this shouldn't compare the function but the result of the function. e.g the actual `now` value. Also you are comparing getNow with it self here...

I don't actually think this is needed, it should already update if now changes

selenir: It won't work in a scenario when `isNow` doesn't change, but `getNow` does, because few lines earlier there's a check if `isNow` has changed:
https://github.com/intljusticemission/react-big-calendar/blob/1c101aae13f4a16426685208ab7403ba05c4c378/src/DayColumn.js#L47

> Also you are comparing getNow with it self here...

My bad, meant to compare it with previous value :wink: 
Check out the second commit, I've restructured the code a bit.

selenir: @jquense any updates on this one?

jquense:
thanks!

selenir:
@jquense No problem, when can we expect new version to be released?

PR #1241, opened 2019-02-28
Adding missing props on slot select in Month view

aravindsrivats:
Pass the 'bounds' and 'box' props on slot select in Month view to the handler


jquense:
thanks!

PR #1245, opened 2019-03-05
Fix the Example page's title, and improve docs

MatthewHerbst:
This PR does the following:

- Standardizes some documentation language
- Properly sets the Example page's document title
  - Because we use [`html-webpack-plugin`](https://www.npmjs.com/package/html-webpack-plugin) (via [`webpack-atoms`](https://www.npmjs.com/package/webpack-atoms)) the `index.html` file that was here was not being used
  - This properly sets the page's title using the `title` attribute to the `html-webpack-plugin`


MatthewHerbst:
Oh, I just noticed that this repo has a `yarn.lock` file. Would you prefer me remove the `package-lock.json` and change the docs to run `yarn` instead of `npm install`?

jquense:
that would be great yeah 👍

MatthewHerbst:
@jquense done!

MatthewHerbst:
@jquense this is ready for final review I believe when you get a chance, thank you!

jquense:
thanks!

PR #1246, opened 2019-03-06
Support point-in-time events in the Agenda view

MatthewHerbst:
Some events may have the same start/end date/time, and are meant to represent point-in-time events, such as deadlines. These are not all day events, but they also should not be represented as `${startTime} - ${endTime}` since the values will be the same.

This will cause an event that is not marked as all day, and where the start and end are equal, to show as:

```js
// old
`${startTime} — ${endTime}` // 5:04 pm — 5:04 pm

// new
`${startTime}` // 5:04 pm
```

### DIFF
--- a/examples/events.js
+++ b/examples/events.js
@@ -1,3 +1,5 @@
+const now = new Date()
+
 export default [
   {
     id: 0,
@@ -102,5 +104,11 @@ export default [
     title: 'Today',
     start: new Date(new Date().setHours(new Date().getHours() - 3)),
     end: new Date(new Date().setHours(new Date().getHours() + 3)),
+  },
+  {
+    id: 15,
+    title: 'Point in Time Event',
+    start: now,
+    end: now,
   },
 ]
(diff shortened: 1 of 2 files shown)


src/localizers/moment.js
```diff
@@ -36,6 +39,7 @@ export let formats = {
   agendaHeaderFormat: dateRangeFormat,
 
   agendaDateFormat: 'ddd MMM DD',
+  agendaSingleTimeFormat: singleTimeFormat,
```
jquense: Can we use the format below `agendaTimeFormat` instead of adding a new one? It should be what we want here 👍

MatthewHerbst: Good catch, done!

MatthewHerbst:
@jquense are there any additional changes you want for this one?

src/localizers/oldGlobalize.js
```diff
@@ -36,6 +36,7 @@ export let formats = {
   agendaHeaderFormat: dateRangeFormat,
 
   agendaDateFormat: 'ddd MMM dd',
+  agendaSingleTimeFormat: 't',
```
jquense: sorry, same for this one, I think `agendaTimeFormat` should be good here without the need for a new format

MatthewHerbst:
Changes made! Thanks for the review

jquense:
Thanks!

PR #1247, opened 2019-03-06
Use bound functions for callbacks

MatthewHerbst:
Using anonymous functions in props will cause the child to always re-render since the function is redefined every render. By using a bound function the reference to the function is always the same, avoiding useless re-render.


jquense:
Thanks for the PR, and taking the time to write it up. Bound methods only avoid a render if the child component implements shouldComponentUpdate which the host (DOM) elements don't so it's not actually saving any work. In this case it's probably better to avoid the etc code and indirection all things being equal :)

PR #1249, opened 2019-03-06
Add `onSelectEvent` & `onDoubleClickEvent` support to Agenda

MatthewHerbst:
This adds support for the `onSelectEvent` and the `onDoubleClickEvent` props to the Agenda view. Specifically, if an event in the "Event" column is clicked/doubleClicked, it will trigger the props if they were given.

### DIFF
--- a/src/Agenda.js
+++ b/src/Agenda.js
@@ -63,11 +63,13 @@ class Agenda extends React.Component {
 
   renderDay = (day, events, dayKey) => {
     let {
-      selected,
-      getters,
       accessors,
-      localizer,
       components: { event: Event, date: AgendaDate },
+      getters,
+      localizer,
+      onDoubleClickEvent,
+      onSelectEvent,
+      selected,
     } = this.props
 
     events = events.filter(e =>
@@ -110,7 +112,13 @@ class Agenda extends React.Component {
           <td className="rbc-agenda-time-cell">
             {this.timeRangeLabel(day, event)}
           </td>
-          <td className="rbc-agenda-event-cell">
+          <td
+            className="rbc-agenda-event-cell"
+            onClick={e => onSelectEvent && onSelectEvent(event, e)}
+            onDoubleClick={e =>
+              onDoubleClickEvent && onDoubleClickEvent(event, e)
+            }
+          >
             {Event ? <Event event={event} title={title} /> : title}
           </td>
         </tr>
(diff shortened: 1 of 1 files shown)


jquense:
There is some previous discussion around why these aren't on agenda in another PR/issue. The short answer tho is that adding them would broadly conflict with events on custom agenda components, a situation that's a bit more likely than in the other views. I'm sorry I don't have the previous issue but it should be not to tough to search for, for more context and possible next steps

MatthewHerbst:
@jquense what if we only did this when the user did not pass a custom component?

peterzanetti:
Why was this not handled? I am trying to find ways to make Agenda clickable. It's the only view that works well with large lists of events.

peterzanetti:
This should be reopened and reconsidered. It makes sense for Agenda to be clickable like every other view.

peterzanetti:
> There is some previous discussion around why these aren't on agenda in another PR/issue. The short answer tho is that adding them would broadly conflict with events on custom agenda components, a situation that's a bit more likely than in the other views. I'm sorry I don't have the previous issue but it should be not to tough to search for, for more context and possible next steps

This is a such an incredibly poor reason. People select a calendar library so that they don't have to build calendar functionality. You're basically saying that "since this feature wasn't there already, some people have built their way around it, and adding the feature now would conflict with those custom, unofficial, unsupported solutions." This completely flies against the logic of an open source library. It is a good idea, others have contributed it for you (multiple times). It should be included.

pietrofxq:
I don't see a problem with this PR since there are checks. The consumer of the component will be responsible for dealing with conflicts if there are any. This won't affect current usage since these are new props

pietrofxq (approved):

MatthewHerbst:
Been a while since I looked at this one! Will try to clean up the merge conflicts this evening. Thanks for the review @pietrofxq

MatthewHerbst:
@pietrofxq I've resolved the merged conflicts. With a final review this should be good to be merged, thanks!

pietrofxq:
@MatthewHerbst Yes, you can use `git commit --amend`

MatthewHerbst:
Best I can tell I did things right, but it claims there are merge conflicts due to code that isn't even on `master`, I'm beyond confused.

peterzanetti:
Thank you for this. I just tried it in my app and it is working. I think you might have forgot cursor: pointer style for the agenda events, but it works.

PR #1255, opened 2019-03-10
added icons to toolbar

JessicaBorja:
Added fontawesome icons to Previous, Next and Agenda buttons


jquense:
hi there, thanks for the PR. Not sure if you meant to send this to your fork, but I don't think this is a good change for the default component. It adds build complexity and design opinions that are harder to change. If you want add icons you can still do via a custom toolbar or custom messages thanks!

PR #1256, opened 2019-03-11
Allow for custom titles in the toolbar

MatthewHerbst:
Currently, even when building a custom toolbar, the `label` prop passed to it is not controllable. Further, for someone who wants to modify just the `label`, building a custom toolbar component is overly complex.

This allows for the user to pass a new prop, `getLabel`, which allows for customizing the label passed to the toolbar. The prop is a function which is given access to the current date, the current view, and the view's length (when needed).

This also adds the `Custom Header` example to the examples list. This page existed but was not listed. This fixes that.

### DIFF
--- a/src/Agenda.js
+++ b/src/Agenda.js
@@ -5,7 +5,7 @@ import getWidth from 'dom-helpers/query/width'
 import scrollbarSize from 'dom-helpers/util/scrollbarSize'
 
 import dates from './utils/dates'
-import { navigate } from './utils/constants'
+import { navigate, views } from './utils/constants'
 import { inRange } from './utils/eventLevels'
 import { isSelected } from './utils/selection'
 
@@ -218,8 +218,16 @@ Agenda.navigate = (date, action, { length = Agenda.defaultProps.length }) => {
   }
 }
 
-Agenda.title = (start, { length = Agenda.defaultProps.length, localizer }) => {
+Agenda.title = (
+  start,
+  { getLabel, length = Agenda.defaultProps.length, localizer }
+) => {
+  if (getLabel) {
+    return getLabel(views.AGENDA, start, length)
+  }
+
   let end = dates.add(start, length, 'day')
+
   return localizer.format({ start, end }, 'agendaHeaderFormat')
 }
 
--- a/examples/App.js
+++ b/examples/App.js
@@ -20,6 +20,7 @@ import Selectable from './demos/selectable'
 import Cultures from './demos/cultures'
 import Popup from './demos/popup'
 import Rendering from './demos/rendering'
+import CustomHeader from './demos/customHeader'
 import CustomView from './demos/customView'
 import Resource from './demos/resource'
 import DndResource from './demos/dndresource'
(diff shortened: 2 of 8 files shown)


src/Agenda.js
```diff
@@ -218,8 +218,16 @@ Agenda.navigate = (date, action, { length = Agenda.defaultProps.length }) => {
...
+  start,
+  { getLabel, length = Agenda.defaultProps.length, localizer }
+) => {
+  if (getLabel) {
+    return getLabel(views.AGENDA, start, length)
+  }
+
   let end = dates.add(start, length, 'day')
+
   return localizer.format({ start, end }, 'agendaHeaderFormat')
```
jquense: what do think about instead of adding `getLabel` suggesting folks use a custom format?

MatthewHerbst: What I'm trying to do is allow for custom text in the label. I tried solving this with `formats.monthHeaderFormat` but the custom text messes with the formatting. Specifically, I'm trying to add "(Loading...)" to the title when my events are loading. Since I'm using moment, it thinks the "L" is part of the format string, which it's not.

jquense: I think you can you a function that returns a string or react node instead of just a string and it should work? Worth a try before adding another API!

MatthewHerbst: `monthHeaderFormat` is merged with the other formats of the localizer in `mergeWithDefaults` when the context is first created:

```js
localizer: mergeWithDefaults(localizer, culture, formats, msgs),
```

It is then fed directly into the localizer, e.g. `localizer.format(date, 'monthHeaderFormat')`. For Moment.js, the `format` function is:

```js
format(value, format, culture) {
  return locale(moment(value), culture).format(format)
},
```

The Moment.js `format` function only accepts a string, and then we get to my original problem: that string needs to be very specific or you will get unexpected results.

So there's no spot that I cans see where passing a function would work unfortunately.

Besides honestly, I wouldn't expect a formatting prop to be able to completely change what React node is rendered.

I agree API expansion should be looked carefully upon, but I think the current approach is the only way to do this based on the current architecture. Please let me know if you see a better way, I'd happily try it.

jquense: all localizers accept functions, it happens before the snippet in the one you noted (https://github.com/intljusticemission/react-big-calendar/blob/master/src/localizer.js#L9)

which is how the various default messages that are functions actually get run!

so 

```js
formats: {
  monthHeaderFormat: (date) => `render what you want`
}
```

>  I wouldn't expect a formatting prop to be able to completely change what React node is rendered.

I agree it's a little weird but that's mostly down to the naming of the prop. completely changing what react node is rendered is exactly what these do already. I definitely agree that this feels a little awkward of a place for this, but the alternative isn't super good either, we end up with a bunch of `get*` apis that shadow format calls and then effectively do the same thing, but with a clearer name. 

The other problem here is the nesting of API's, it's not enough to just provide the date to getLabel, you'd need to also provide the localizer so a consumer could format the date, etc. It starts to get complex and messy, when maybe the better approach would be to rename `formats` to `labels` (not saying we should do that)

MatthewHerbst:
Thanks for the help finding a solution to this!

PR #1276, opened 2019-04-02
Fix: prevent un/mounting of date components

idanwe:
I wanted to make a transition to date elements when they were selected, releasing that a couldn't get the `transition` property working. 

After some digging to my code and react-big-calendar code, I found that after selecting a date `DateContentRow` un&mount  their children components.

This happens due to the call of [this](https://github.com/intljusticemission/react-big-calendar/blob/master/src/DateContentRow.js#L121) `renderDummy` happens after setting `needLimitMeasure` to true in `Month.jsx` after the date is changed.

----------------------------

Prevent re-mount of DateContentRow children when date change in the same month
 * Prevent setting `needLimitMeasure` to true when the month of the date is not changed,
 in order to prevent unmounting and mounting of all the weeks' components.

 * Make it possible to add transition to the background of the event when
 implementing selection of date within the calendar.

### DIFF
--- a/src/Month.js
+++ b/src/Month.js
@@ -36,7 +36,7 @@ class MonthView extends React.Component {
 
   componentWillReceiveProps({ date }) {
     this.setState({
-      needLimitMeasure: !dates.eq(date, this.props.date),
+      needLimitMeasure: !dates.eq(date, this.props.date, 'month'),
     })
   }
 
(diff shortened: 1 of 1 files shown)


idanwe:
@jquense Is it possible to merge the PR? do you have comments / things I can do to make it easier to merge it?

idanwe:
Thanks!

idanwe:
@jquense do you know when a new release containing that fix will be released?

PR #1279, opened 2019-04-08
added a flag to determine if events should overlap

### LINKED ISSUE #1060: Add option to make events not overlapped when they are in the same timeslot?
Hi everyone.

A feature
Currently when events are in the same timeslot they are overlapped each other.

So my client wants the layout to look somewhat like this.
events

As I searched there is no option to change layout to make events not overlapped like in FullCalendar. Do you guys have any plan to add this feature in the future? I think it looks a lot better like that.

sgraham3311:
fixes #1060 

I think this solves this issue of allowing an optional flag for overlapping meetings. This fix only provides a global flag for all events in the calendar. I am not sure there is a need to have it on a resource level also.

### DIFF
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -631,6 +631,11 @@ class Calendar extends React.Component {
        * An optional event time range for events that continue from another day
        */
       eventTimeRangeEndFormat: dateFormat,
+
+      /**
+       * Determines if meetings should overlap
+       */
+      overlapMeetings: PropTypes.bool
     }),
 
     /**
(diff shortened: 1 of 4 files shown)


euvs:
I don't think the flag should be called `overlapMeetings`. "Meeting" is too domain specific term. The calendar is used for many other purposes than for displaying meetings (e.g. room availability, appointment time etc.). Something like `overlapEvents` would be more consistent with the terminology used in RBC...

sgraham3311:
Good catch. I will update the flag.

jquense:
Thanks for the PR, this isn't quite the way we'd want to implement this. Ideally there should be a way to configure how events lay out generally so we don't have to add one-off flags for specific configurations

Issue #1090, opened 2018-11-06
Drop external events onto calendar?


MinhNguyen41092:
Hi everyone. I've developed an app that requires dragging and dropping events from an external list onto the calendar. I've made it work in the old version using React-DnD. But in the new version it was ditched for another technology that I still can't find out what it is. So I'd like to ask is it possible to drop an outside event onto the calendar by using the new DnD approach? If yes could you give me a hint? Thank you very much.

jquense:
> ditched for another technology that I still can't find out what it is.

no other library we do the drag and drop "in house", all the code for it is in the dnd addon. At the moment there is no first class API for dropping events from an external element, but i'd be open to adding one if anyone has a good idea  for it

zagoret:
> Hi everyone. I've developed an app that requires dragging and dropping events from an external list onto the calendar. I've made it work in the old version using React-DnD. But in the new version it was ditched for another technology that I still can't find out what it is. So I'd like to ask is it possible to drop an outside event onto the calendar by using the new DnD approach? If yes could you give me a hint? Thank you very much.

For which version (old one) did you implemented?

MinhNguyen41092:
@zagoret v0.19.1

zagoret:
@MinhNguyen41092  Thanks for your response.
I implemented also for version 0.19.0 and it works fine.
Any update about the last version, still doesn't work external dnd?

MinhNguyen41092:
@zagoret we found a way to work around it. We wrap the calender with react-dnd drag source and try to calculate the drop position externally. It's not pretty but atleast it works

jquense:
Btw happy to add an API for this. It should be as simple as handling onDrop events and passing back the, already caclulated pointer to slot position

RoslerWS:
I've tried to accomplish this with a method similar to @MinhNguyen41092. except by passing in a custom dateCellWrapper component as a component prop to the calendar that is wrapped with react-dnd DropTarget. Unfortunately, it doesn't work for date rows containing events, because the date cell wrapper wraps the 'rbc-day-bg' cells in the 'rbc-row-bg' row, whereas the 'rbc-row-content' row is what contains the events..... hence the drop target isn't detected because there is something on top of it. If I bring the background cell to the forefront, then the built-in react-big-calendar drag and drop no longer works for these cells, because it is underneath....

kamry-bowman:
[...]

PR #1290, opened 2019-04-19
Provide onDropFromOutside prop for Dnd Cal

kamry-bowman:
This PR is meant to resolve issue #1090.

## Basic callback for outside drops
The change exposes the `onDropFromOutside` prop on the withDragAndDrop HOC, which takes a callback that fires when an outside draggable item is dropped onto the calendar. The callback receives as a parameter an object with start and end properties that are times based on the drop position and slot size. 

a4be055597d294f257d59a6fa2982f27

It is worth noting that it is entirely up to the user to handle actual event creation based on the callback. All that this API does is allow `draggable` DOM elements to trigger a callback that receives start and end times for the slot an item was dropped on, and a boolean as to whether it's an all-day event. If the user wants to know which event was dropped, they will have to handle that themselves outside of React-Big-Calendar. An example added to the example App demonstrates how this can be done.

## Optional selective dropping
By default, if `onDropFromOutside` prop is passed, all draggable events are droppable on calendar. If the user wishes to discriminate as to whether draggable events are droppable on the calendar, they can pass an additional `onDragOver` callback function. The `onDragOver` callback takes a DragEvent as its sole parameter. If it calls the DragEvent's `preventDefault` method, then the draggable item in question is droppable. If it does not call `preventDefault` during the function call, it will not be droppable.

e374b60b55809f471b2f275d7f166278

An example was also added to the examples App, this one labelled `Addon: Drag and Drop (from outside calendar). The GIFs show this example in action. 

I also added the following comments into the withDragAndDrop HOC by way of documentation.

```
 * Additionally, this HOC adds the callback props `onDropFromOutside` and `onDragOver`. 
 * By default, the calendar will not respond to outside draggable items being dropped
 * onto it. However, if `onDropFromOutside` callback is passed, then when draggable 
 * DOM elements are dropped on the calendar, the callback will fire, receiving an 
 * object with start and end times, and an allDay boolean.
 *
 * If `onDropFromOutside` is passed, but `onDragOver` is not, any draggable event will be
 * droppable  onto the calendar by default. On the other hand, if an `onDragOver` callback
 * *is* passed, then it can discriminate as to whether a draggable item is droppable on the
 * calendar. To designate a draggable item as droppable, call `event.preventDefault`
 * inside `onDragOver`. If `event.preventDefault` is not called in the `onDragOver`
 * callback, then the draggable item will not be droppable on the calendar.
```

Hopefully this gives users the flexibility they need, without getting react-big-calendar overly involved with managing outside drag and drop scenarios. Any feedback/discussion/harangues are welcome!

### DIFF
--- a/src/Selection.js
+++ b/src/Selection.js
@@ -54,6 +54,7 @@ class Selection {
     this._handleMoveEvent = this._handleMoveEvent.bind(this)
     this._handleTerminatingEvent = this._handleTerminatingEvent.bind(this)
     this._keyListener = this._keyListener.bind(this)
+    this._dropFromOutsideListener = this._dropFromOutsideListener.bind(this)
 
     // Fixes an iOS 10 bug where scrolling could not be prevented on the window.
     // https://github.com/metafizzy/flickity/issues/457#issuecomment-254501356
@@ -64,6 +65,10 @@ class Selection {
     )
     this._onKeyDownListener = addEventListener('keydown', this._keyListener)
     this._onKeyUpListener = addEventListener('keyup', this._keyListener)
+    this._onDropFromOutsideListener = addEventListener(
+      'drop',
+      this._dropFromOutsideListener
+    )
     this._addInitialEventListener()
   }
 
@@ -185,6 +190,19 @@ class Selection {
         touchStartListener.remove()
       },
     }
+  }
+
+  _dropFromOutsideListener(e) {
+    const { pageX, pageY, clientX, clientY } = getEventCoordinates(e)
+
+    this.emit('dropFromOutside', {
+      x: pageX,
+      y: pageY,
+      clientX: clientX,
+      clientY: clientY,
+    })
+
+    e.preventDefault()
   }
 
   _handleInitialEvent(e) {
--- a/examples/App.js
+++ b/examples/App.js
@@ -25,6 +25,7 @@ import Resource from './demos/resource'
 import DndResource from './demos/dndresource'
 import Timeslots from './demos/timeslots'
 import Dnd from './demos/dnd'
+import DndOutsideSource from './demos/dndOutsideSource'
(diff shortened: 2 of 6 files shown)


Phuong97:
Does it work in version 0.20.4 ?

kamry-bowman:
I'm not sure if I understand. It's a PR into the master branch which is already a few commits ahead of version 0.20.4. So it's not in that version, it would be in a later release assuming it gets merged. But its functionality is compatible with the 0.20.4 codebase from what I can tell; it doesn't rely on anything added in a subsequent commit for its functionality.

src/Selection.js
```diff
@@ -64,6 +65,10 @@ class Selection {
     )
     this._onKeyDownListener = addEventListener('keydown', this._keyListener)
     this._onKeyUpListener = addEventListener('keyup', this._keyListener)
+    this._onDropFromOutsideListener = addEventListener(
```
jquense: dumb question, how do you know if this is "from the outside", is it b/c for "internal" drops we don't actually use the drop/drag events?

jquense:
this is awesome, thanks for taking the time to do this!

src/Selection.js
kamry-bowman: Yes, that's exactly right. I didn't really think about it that way but that's the reason. I think if we did move to drop events for the internals we would have to refactor this approach.

dhyegocalota:
Thank you for your awesome work @kamry-bowman 

I just wanted to know... is there any way also make the events be draggable to outside of the calendar?

yasserzubair:
Is there any documentation or example related to this?

Nikualla1:
> Is there any documentation or example related to this?

https://github.com/jquense/react-big-calendar/blob/master/examples/demos/dndOutsideSource.js

PR #1292, opened 2019-04-22
fix: change tooltipAccessor key to 'tooltip' from 'title'

SergeyBelofost:
Events tooltip was from title of them. Changed key of tooltipAccessor to 'tooltip' for correct work.

### DIFF
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -735,7 +735,7 @@ class Calendar extends React.Component {
     drilldownView: views.DAY,
 
     titleAccessor: 'title',
-    tooltipAccessor: 'title',
+    tooltipAccessor: 'tooltip',
     allDayAccessor: 'allDay',
     startAccessor: 'start',
     endAccessor: 'end',


jquense:
thanks for the PR. This is actually intentional. The default is to use the title of the event for the tooltip. If you want use `tooltip` in your app, then you can specify `tooltipAccessor='tooltip'` on your calendar

SergeyBelofost:
i have "Property 'tooltipAccessor' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<BigCalendar..."

SergeyBelofost:
> thanks for the PR. This is actually intentional. The default is to use the title of the event for the tooltip. If you want use `tooltip` in your app, then you can specify `tooltipAccessor='tooltip'` on your calendar

**Property 'tooltipAccessor' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<BigCalendar**

jquense:
That is a problem with the typescript types, which we don't maintain. I'd send DT a PR if it's missing there.

Angelk90:
Hi @jquense , it would be possible to implement the possibility of being able to replace the corrwnte tooltip of the event with one that is passed as props to the calendar.

For example I would like to use this: https://material-ui.com/components/tooltips/

PR #1297, opened 2019-04-29
Changes to Min and Max cause bugs in Current Time Indicator

aminland:
Ensure that changes in min and max time is reflected when computing the Now indicator

### DIFF
--- a/src/DayColumn.js
+++ b/src/DayColumn.js
@@ -44,18 +44,17 @@ class DayColumn extends React.Component {
   }
 
   componentDidUpdate(prevProps, prevState) {
-    const getNowChanged = !dates.eq(
-      prevProps.getNow(),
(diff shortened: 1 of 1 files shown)


arecvlohe:
@aminland Since I am not familiar with this issue can you provide steps to reproduce the error.

sonnyp:
related: https://github.com/intljusticemission/react-big-calendar/pull/1311

aminland:
Sorry, didnt see this.

Steps to reproduce are, have a calendar, and set the min and max time to some dynamic value. 
In our usage for each day, the min and max time are different (We dynamically compute what time range to display based on when there are events.)

The other PR #1311 would address this also in the `min` case, so feel free to close this out if you like that solution.

PR #1301, opened 2019-05-06
chore(deps): update overlay

cutterbl:
Update the react-overlay dependency and usage within the project. Necessary update to avoid conflicts with those using latest version of react-overlay (for instance, those creating Modal managers in their projects, consuming react-bootstrap).


arecvlohe:
@jquense Thoughts?

cutterbl:
I fixed the merge conflict. Since the Popper get's it's `ref` from Overlay, there's no need for the `createRef` directly in the Popper.

cutterbl:
Sorry for the back and forth. The only OS projects I regularly send updates to are my own, so reacquainting with the workflow here. Think all the bits are there now. Tests are passing, examples are rebuilt and running properly, and no console messages.

src/Month.js
```diff
@@ -12,7 +12,7 @@ import getPosition from 'dom-helpers/query/position'
 import raf from 'dom-helpers/util/requestAnimationFrame'
 
 import Popup from './Popup'
-import Overlay from 'react-overlays/lib/Overlay'
+import { Overlay } from 'react-overlays'
```
jquense: lets keep the cherry picked import i don't want to rely on treeshaking working correctly if we don't need to

jquense: `import Overlay from 'react-overlays/Overlay'`

cutterbl: Done

jquense:
thanks!

PR #1308, opened 2019-05-07
chore: upgrade storybook

cutterbl:
Upgrade Storybook dependencies and configuration/addon registration.


.storybook/webpack.config.js
```diff
@@ -1,19 +1,13 @@
...
-      ],
-    },
-    plugins: [...config.plugins],
-  }
+module.exports = function({ config }) {
+  config.module.rules.push(rules.js())
+  config.module.rules.push(rules.fonts())
+  config.module.rules.push(rules.images())
+  config.module.rules.push(rules.css())
+  config.module.rules.push(rules.less({ browsers }))
```
jquense: this will now use the default rules, which i don't think we want...

cutterbl: I've pushed an update to completely override the 'rules'. They did have a rule (from the `@storybook/react` package) already for the js compile, as well as some bits for handling media, but this new config more closely mirrors what you were doing before.

jquense:
thanks!

PR #1311, opened 2019-05-08
fix: update time indicator position if min prop changes

sonnyp:
When changing the `min` prop, the current time indicator is not re-positioned and therefore shows at the wrong place.

demonstration: https://codesandbox.io/s/y241x9njq1

### DIFF
--- a/src/DayColumn.js
+++ b/src/DayColumn.js
@@ -61,6 +61,11 @@ class DayColumn extends React.Component {
 
         this.setTimeIndicatorPositionUpdateInterval(tail)
       }
+    } else if (
+      this.props.isNow &&
+      !dates.eq(prevProps.min, this.props.min, 'minutes')
+    ) {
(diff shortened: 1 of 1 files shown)


src/DayColumn.js
```diff
@@ -61,6 +61,8 @@ class DayColumn extends React.Component {
 
         this.setTimeIndicatorPositionUpdateInterval(tail)
       }
+    } else if (this.props.isNow && prevProps.min !== this.props.min) {
```
jquense:
```suggestion
    } else if (this.props.isNow && !dates.eq(prevProps.min, this.props.min, 'minutes')) {
```

sonnyp:
@jquense done

jquense:
thanks!

sonnyp:
do you plan a release soon?

aminland:
Changes to the max time will still have the same issue in this PR

PR #1313, opened 2019-05-08
feat: switch to Sass for styles

cutterbl:
Updates to component styling. Includes all code from PR #1235 .

- Build Process (remove LESS and include autoprefixer)
- Examples
- Storybook
- README
- Examples Into


cutterbl:
@jquense You will definitely want a "Breaking Change" entry in the Changelog for this, for those using the LESS...

cutterbl:
@jquense I'm reviewing the package.json, and curious about dependency usage. Seeing `react` and `react-dom` listed in both the `devDependencies` and the `peerDependencies`, with different versions. Also wondering if some of the `dependencies` shouldn't move into either `devDependencies` or `peerDependencies` (classnames, prop-types, etc).

jquense:
react is in both b/c the peerDep is signal that we need react but npm/yarn won't install it automatically (up to you to install the one version you want). It's also in devDeps, b/c to run tests and CI, etc we actually need to install react. There is probably som dep cleanup to be done but generally only things that must be installed  only once belong in peer deps, so classnames, proptypes are better as direct dependencies

package.json
```diff
@@ -116,12 +117,16 @@
...
+    "autoprefixer": "^9.5.1",
+    "bootstrap-sass": "^3.4.1",
     "classnames": "^2.2.6",
     "date-arithmetic": "^3.0.0",
     "dom-helpers": "^3.4.0",
     "invariant": "^2.2.4",
     "lodash": "^4.17.11",
     "memoize-one": "^4.0.3",
+    "postcss": "^7.0.16",
+    "postcss-cli": "^6.1.2",
```
jquense: these all belong in devDeps

cutterbl: On it (still getting used to yarn again)

cutterbl: Done

jquense:
thanks!

Coriou:
Thanks for the awesome work ! Since it's merged into master, can we expect a NPM publish soon ? Thanks :)

xtursky:
@cutterbl thanks for great work.
If I may ask, why was bootstrap-sass added to dependencies instead of devDependencies? It's used only in examples and both bootstrap and react-bootstrap are also defined there.

jquense:
maybe a bug, happy to take a PR

PR #1315, opened 2019-05-09
Open Examples Direct

cutterbl:
This change, to the Examples webpack config, handles several changes

- Builds all assets to `./examples/static` (folder targeted by the `clean:examples` script)
- Automatically opens the default browser to the target content (`./examples/static/index.html`) when running `examples` script.
- Remove `server.js` (ref'd `../webpack/examples.config.es6`, which doesn't exist, and the file wasn't used)

[Side Note: I realize the webpack-dev-server doesn't use the `./examples/static` folder, in the project, at all, but this does help the cleanup and placement of those files generated by `examples:build` script, which I'm sure are used for the website)]


examples/webpack.config.js
```diff
@@ -26,8 +32,8 @@ module.exports = {
   module: {
     rules: [
       rules.js({}),
-      rules.images(),
-      rules.fonts(),
+      rules.images({ outputPath: 'static/' }),
+      rules.fonts({ outputPath: 'static/' }),
```
jquense: shouldn't be needed

examples/webpack.config.js
```diff
@@ -38,6 +44,8 @@ module.exports = {
   },
   plugins: [
     plugins.html({ title: 'React Big Calendar' }),
-    plugins.extractCss(),
+    plugins.extractCss({
+      filename: 'static/[name].[hash].css',
```
jquense: also shouldn't be needed, things will be output to the the `output.path` automatically

cutterbl:
Latest change allows `build:examples` to build to new root `/docs` folder. Then you can switch the GitHub pages `source` to the `master branch /docs folder` option, serving the GitHub pages project site direct from the `/docs` folder.
Screen Shot 2019-05-09 at 11 27 21 AM

cutterbl:
I included a GitHub pages 404.html, with auto redirect. This will take users from the current link directly to the new link that would be created by this process. I had to temporarily remove the `clean:examples` bits from the `examples` scripts, so that those tasks would not blow this file out (must stay in root of docs, and I can't build example assets to different outputPaths, as `webpack-atoms` doesn't support custom options for these rules)

PR #1321, opened 2019-05-15
redeclared all sass variables using !default

DevelopmentByDavid:
This is a pretty minor pr, but nice to have.


jquense:
Thanks!

PR #1322, opened 2019-05-15
Fix drag cancelation for month view

xtursky:
This is same fix as already done for day and week views done in #1119 done for month view.

* Fix dropping an event out of month view bounds leaving the calendar in an intermediate state (no listener on reset)


jquense:
thanks!

PR #1323, opened 2019-05-16
docs(dnd): remove deprecated comment about `react-dnd`

onlyann:
This is really minor. 
The comment had me confused for a while until I figured out react-dnd is no longer used on this project.


jquense:
thanks!

PR #1324, opened 2019-05-16
add customizeClass for rbc-row-segment

pl2476:
image
image


jquense:
Hi there, I think you can already accomplish what you want here with `eventPropGetter` which lets you return a custom class based on the event

pl2476:
ok, I get it, thank you,

pl2476:
Can I add style to onDropFromOutside ? So I can get the custom style when dragging external data.

pl2476:
> Can I add style to onDropFromOutside ? So I can get the custom style when dragging external data.

@jquense

PR #1330, opened 2019-05-19
feat(event): specific background color for each event

mohammadmoozeh:
Just add 'bgColor' field in object.

### DIFF
--- a/examples/events.js
+++ b/examples/events.js
@@ -7,6 +7,7 @@ export default [
     allDay: true,
     start: new Date(2015, 3, 0),
     end: new Date(2015, 3, 1),
+    bgColor: '#4ef12d'
   },
   {
     id: 1,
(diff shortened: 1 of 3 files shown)


jquense:
You can use eventPropGetter for this already happily 👍

mohammadmoozeh:
Would you give me an example?

mohammadmoozeh:
@jquense I tried eventPropGetter but it changes all events color.
Could you help me?
Thanks

jquense:
It's a function that's called for _each_ event change what you return for only the events you want.

```
event => ({ className: event.isBusy ? 'red' : null })
```

PaitoAnderson:
Just in case anyone needs a more complete example implementation...

```jsx
<Calendar
    events={this.state.events}
    eventPropGetter={(event, start, end, isSelected) => {
        if (event.color) {
            return { style: { backgroundColor: event.color }}
        }
        return {}
    }}
/>
```

Issue #1329, opened 2019-05-17
Set rage of days on week view


lksilva:
Is there any way to set rage of days in week view mode?

I'm using code below:

```js
        <DragAndDropCalendar
          selectable={true}
          localizer={localizer}
          events={filtered}
          onEventDrop={this.moveEvent}
          resizable={true}
          onEventResize={this.resizeEvent}
          defaultView={BigCalendar.Views.WEEK}
          date={date}
          min={min}
          max={max}
          step={interval}
          timeslots={60 / interval}
          onSelectSlot={this.newEvent} />
```

My problem is that when I fetch data from backend I receive payload with date today and seven days ahead, however in week mode always start on sunday but I need set start week day, is there any way?

In the case below I want day of week began on friday 

> <img src="https://i.imgur.com/fHsN4EY.png" />

jquense:
start of the week is determined by the localizer. you can use the localizer api to customize. if you want to control specifically which days show up you need to provide a Custom view, take a look at how WorkWeek is implemented

lksilva:
What do you think about I make a PR that enable receive a prop into Calendar to determine start day of week? For example:

```js
      <DragAndDropCalendar
          selectable={true}
          localizer={localizer}
          events={filtered}
          onEventDrop={this.moveEvent}
          resizable={true}
          onEventResize={this.resizeEvent}
          defaultView={BigCalendar.Views.WEEK}
          startWeekDay={3} // In this case 3 is Wednesday
```

Because I could not do a example as you suggested

PR #1333, opened 2019-05-20
feat: implement firstOfWeek to determines which day of the week to start Grid in week view

lksilva:
I need set start day in week view, then I make this PR with code solution, a better explanation about this problem in https://github.com/intljusticemission/react-big-calendar/issues/1329


lksilva:
@jquense any news?

PR #1334, opened 2019-05-21
 add style to onDropFromOutside

pl2476:


pl2476:
I want to get custom style onDropFromOutside,  please confirm this requirement or there are other ways to achieve, thank you very much. 
@jquense

jquense:
why do you need the style? or what data is in there that is specifically helpful to you? It seems less than ideal to just pass it through

pl2476:
> why do you need the style? or what data is in there that is specifically helpful to you? It seems less than ideal to just pass it through

Example:  There is some external data that needs to display different styles depending on the state when dragging in. I think this is a common usage scenario. Of course, I want to use it.

Issue #1234, opened 2019-02-22
IE11 findDOMNode exception in click selection with multiple routes


martin-strobel:
#### Do you want to request a _feature_ or report a _bug_?
Bug

#### What's the current behavior?

Environment: IE11 with
- react-big-calendar 0.20.3
- moment 0.24.0
- react 16.8.2
- react-router 4.3.1
- react-app-polyfill 0.2.1
- babel-polyfill 7.2.5
installed.

Test-URL: https://martin-strobel.github.io/rbc-ie11-bug/dashboard
Test-Repo: https://github.com/martin-strobel/rbc-ie11-bug/

How to reproduce bug:

1. Open DEV-Console in IE11
2. Click on a date
3. Once weekview is visible: Try to select a timeslot or click somewhere else
   This also occurs after switching back to dashboard route.
4. React shows error "Unable to find node on an unmounted component."

![image](https://user-images.githubusercontent.com/46812665/53236155-eb764580-3693-11e9-9af7-e7775f7e8a49.png)

This seems to happen because of findDOMNode calls, but I was unable to fix this.
On an unrelated note: Sometimes timeslots are not rendered correctly, this doesn't seem to influence the reported bug.

This trace happened on the calendar (WeekView) page:
![image](https://user-images.githubusercontent.com/46812665/53236293-3a23df80-3694-11e9-9279-09788547f4da.png)

#### What's the expected behavior?

There shouldn't be any (uncaught) errors when switching between Routes.

DenysZP:
It seems that IE incorrectly or not fully unsubscribes from listeners.
I suggest adding `isDetached` flag to the Selection class.
``` js
class Selection {
  constructor(node, { global = false, longPressThreshold = 250 } = {}) {
    this.isDisabled = false
    ...
  }

  teardown() {
    this.isDisabled = true
    ...
  }

  _handleInitialEvent(e) {
    if (this.isDisabled) {
      return
    }
    ...
  }
}
```
If this approach makes sense, I can create a PR.

dkniffin:
I'm also experiencing this issue, and this change also fixed it for me. @jquense can you confirm that this is a good approach to fixing it, then @DenysZP can make a PR?

dkniffin:
Additionally, adding another `if (this.isDetached) { return }` inside `_handleMoveEvent` addresses this other error that's coming up: `Unable to get property 'x' of undefined or null reference`

vienpndn:
[...]

PR #1338, opened 2019-05-23
fix: ie fix for event bindings on unmounted components

dkniffin:
Addresses #1234


dkniffin:
@jquense Is there anything else I need to do to get this merged?

jquense:
thanks!

Issue #1105, opened 2018-11-12
DnD: dragging won't start on first attempt (when no events are visible when initialized?)


xtursky:
#### Do you want to request a _feature_ or report a _bug_?

bug

#### What's the current behavior?

This took me quite some time to simulate, so bear with my explanation.

Problem is, that DnD at some cases seems to be not fully initialized and then is not working on each odd attempt. 
Problem can be easily seen here: https://codesandbox.io/s/640nmy650r
- If I set defaultDate to be at the same week as events, DnD is working.
- If I set defaultDate to start at some other week without events, then after navigation via Previous/Next controls to events, DnD won't start on each odd attempt. Every second attempt works.

In depth in first attempt I saw that drag is not started due to **beforeSelect** check, where state accessible via `dragAndDropAction` is not yet filled. Maybe issue with event listeners?

I had this problem starting in WEEK and DAY view (maybe it's also issue with MONTH).
**!Beware**: hot reload causes correct re-initialization, so you need to refresh whole page before each try.

Problem is not connected to OS or browser.

#### What's the expected behavior?

DnD works no matter what `defaultDate` and `view` is used.

bminer:
I seem to be able to confirm this bug.  Here's how:
- Use `onSelectSlot` to create a new event and update the `events` state
- Ensure that `events` does not contain events that will render on initial load

If the `events` array has one or more items in it that will be rendered, everything works fine; otherwise, the move/resize events only work every other time, and other strange behavior occurs also.

Consider the following code:

```js
import React, { Component } from "react";
import Calendar from "react-big-calendar";
import moment from "moment";
import withDragAndDrop from "react-big-calendar/lib/addons/dragAndDrop";

import "react-big-calendar/lib/addons/dragAndDrop/styles.css";
import "react-big-calendar/lib/css/react-big-calendar.css";

const localizer = Calendar.momentLocalizer(moment);

const DnDCalendar = withDragAndDrop(Calendar);

class App extends Component {
  state = {
    events: [
      {
        start: new Date("1970-01-01"),
        end: new Date("1970-01-01 2:00 PM"),
        title: "Some title"
      }
    ]
  };

  onSelectSlot = ({action, start, end }) => {
    this.setState(state => {
      const events = [...state.events, {
        start, end,
        title: "New Event"
      }]
      return { events };
    });
  }

  onEventResize = ({ event, start, end, allDay }) => {
    this.setState(state => {
      const events = [...state.events]
[...]

Issue #1285, opened 2019-04-15
Fix DnD for multiday events to not cut end date


xtursky:
#### Do you want to request a _feature_ or report a _bug_?

Bug / Question

#### What's the current behavior?

Currently it's not possible to DnD multiday event (in case of showMultiDayTimes=true). When dragging starts, end time is limited to end of current day by this line of code https://github.com/intljusticemission/react-big-calendar/blob/475674c6167391bd6212c1247fa29c1aea4c1b00/src/utils/TimeSlots.js#L129

@jquense question, is this just some safety done before DnD was introduced, or does it have it's meaning?

If this line is ignored, DnD seems to work for both one day and multiday events. End time is correctly calculated. Only small issue is, that only first day is visualized while dragging, but that I can live without.

#### What's the expected behavior?
To be able to DnD multiday event (in case of showMultiDayTimes=true) correctly. So the start and end time are not limited to the current day.

arecvlohe:
@xtursky This seems like it would be an easy fix if your solution above works. Mind submitting a PR?

xtursky:
@arecvlohe thanks for info. No problem to create a PR.
As I asked above, not sure if this is needed when DnD is not used.
So I wasn't sure if to change this generally or just for DnD addon. I'll prepare PR, test it and we will see.

felgeekpe:
@xtursky hi, any updates on the fix? Thanks!

xtursky:
@felgeekpe yeap, there is already PR waiting for review, just didn't reference it correctly. Thanks for asking.

PR #1342, opened 2019-05-29
fix(addons): do not cut end while dragging multiday event

xtursky:
Tries to fix issue #1105.
Added options to ignore timeslot start or end so day is not cut to the end of current day during DnD move.
Used only in DnD handleMove - rest remains as before.
Only end date is ignored - tested, worked ok. 
When start date was ignored it causes some issues slot not being rendered in correct slot sometimes.

Wasn't sure about names, so feel free to comment :)


xtursky:
Sorry guys, accidentally put here wrong issue reference.
This one fixes this small problem in #1285

jquense:
thanks!

PR #1348, opened 2019-06-09
feat: provide named exports api

TrySound:
```js
import {
  Calendar,
  DateLocalizer,
  momentLocalizer,
  globalizeLocalizer,
  move,
  Views,
  Navigate,
  components
} from 'react-big-calendar';
```


TrySound:
Rebased

jquense:
the docs and examples need to be updated as well actually

PR #1349, opened 2019-06-09
Remove prop-types-extra

TrySound:
I found `all(object, customHandler)` can be replaced with
`objectOf(customHandler)` so we may drop prop-types-extra package.

There is still usage in react-overlays. Gonna send a PR too.


jquense:
Thanks!

PR #1350, opened 2019-06-09
feat: use lodash-es for esm bundle

TrySound:
Similar way is used by formik.

https://unpkg.com/formik@1.5.7/dist/formik.esm.js
https://unpkg.com/formik@1.5.7/dist/formik.cjs.development.js

As you can see there is nice size reduction.


jquense:
Size goes up with this change?

TrySound:
Its just -es suffix per every usage. Bundled version is more descriptive.
https://github.com/intljusticemission/react-big-calendar/pull/1350/files#diff-29fd455fe4a1c8d19c5e5b1420ce14faR9

TrySound:
@jquense rebased

PR #1353, opened 2019-06-10
fix: rtl incorrectly named or not propagated

xtursky:
During some previous refactoring rtl was on some places renamed to `isRtl` or forgotten to propagate, causing it to not work at all - css was ignored.

I've reverted back to `rtl` notation as it common word already used in styles and in documentation as well.


jquense:
thanks!

xtursky:
@jquense thanks for quick response, could I ask if there is some release planned

PR #1354, opened 2019-06-10
chore(deps): upgrade date-math

TrySound:
The diff is quite nice.

```diff
-    "minified": 150810,
+    "minified": 148190,
-    "gzipped": 45487
+    "gzipped": 44792
```

### DIFF
--- a/src/utils/dates.js
+++ b/src/utils/dates.js
@@ -1,5 +1,24 @@
 /* eslint no-fallthrough: off */
-import dateMath from 'date-arithmetic'
+import * as dates from 'date-arithmetic'
+
+export {
+  milliseconds,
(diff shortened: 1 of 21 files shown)


src/utils/dates.js
```diff
@@ -1,5 +1,24 @@
...
+  add,
+  eq,
+  gte,
+  gt,
+  lte,
+  lt,
+  inRange,
+  min,
+  max,
+} from 'date-arithmetic'
```
jquense: both imports?

TrySound: Nope. This one is only reexport.

jquense: 👍

PR #1365, opened 2019-06-22
Feature/button components

jstacoder:
i have a need for a custom toolbar, but i would like to keep the default buttons, so i thought i would submit this PR. please let me know if there are any suggestions.


jstacoder:
the benefit is that with this you can add to the buttons, which you currently cannot do, what if someone just wants a button, maybe an add event button, but they want to keep the functionality and look of the existing toolbar and buttons.

mokimo:
I'm with jquense here. IMO the functionality to customize the toolbar is given in a simple way.
If you want to extend the functionality, you're most of the time better of having a custom-toolbar

jstacoder:
sure, if you only want this used by developers, but its almost easy enough to use by anyone, there is a big difference between knowing how to write a simple button component with an onChange callback vs reimplementing the entire toolbar (dont get me wrong, i can and have reimplemented the toolbar)  but i know not everyone can or will do that, but people might be able to add a button or 2. Im just trying to make it useful to more people. but if you guys still think its a bad fit i can close it.

PR #1369, opened 2019-06-26
feat(dnd): add preview of an item inside cell while dragging

asakalou-exadel:
While dragging over a cell it might be useful to see placeholder for an item where it will be inserted as soon we drop it.

**Please see example below**

drag


asakalou-exadel:
@jquense Hello! Thanks for merge!

When do you plan next release?

cjduncana:
Just noticed that this PR causes the events to overlap existing events, instead of temporarily adding it to the list.

PR #1374, opened 2019-07-03
Fix: use fixed date arithmetic lib and move bt-sass to devdependecies

xtursky:
Increased date-arithmetic lib version with fixed bug which causes issue with month range calculation.
I also moved bootstrap-sass to dev-dependencies, as discussed in #1313 as it's only used in examples.


TrySound:
@xtursky Did you run yarn?

xtursky:
Ooh, sry, forgot about that. Will fix it right away.

xtursky:
@TrySound fixed, probably missed that during branch creation

jquense:
thanks!

Issue #1390, opened 2019-07-16
Wrong dashes in date and time spans


oestrogen:
#### Do you want to request a _feature_ or report a _bug_?
Bug?
#### What's the current behavior?

Date and time ranges use em dashes (—) between the start and end, except for week ranges that use hyphens (–).

#### What's the expected behavior?

The typographically correct dash to use is the en dash (–). This according to most style guides, such as the one from APA, American Psycological Association.

Using en dashes in ranges also increases the chance that screen readers reads the dash as "to" instead of skipping over it.

PR #1391, opened 2019-07-16
#1390 Only use en dashes in ranges

oestrogen:
As described in #1390, it's wrong to use em dashes (and suboptimal to use hyphens) in time ranges.

Here I have changed all dashes in all ranges to en dashes for accessibility and typographical correctness.


jquense:
thanks!

nullsage:
So in my app i am not seeing any dash because the character is not recognized by my browser or something squirly is going on. What can i do to make this work properly?
no-dash

PR #1405, opened 2019-07-31
Improve DayEventLayout events arrangement algorithm

### LINKED ISSUE #1397: Having Concurrent Later Overlapping Events Hides Earier Event

_Bug_

A regression/edge case of #404 (which should be reopened)

Image Of Issue

Code Sandbox: https://codesandbox.io/embed/react-big-calendar-example-96cue

This edge case occurs precisely when the following two conditions are satisfied:

1. There is an event A that overlaps with another event B which ends later
2. There are events C and D that begin after A but overlap with B

Affected Browser(s):

- Mozilla Firefox 68.0
- Google Chrome 75.0.3770.142

For event B not to be completely hidden by events C/D. Example (from Google Calendar):
ReactBigCalendarExpectedResult

dmitrykrylov:
Fixes #1397 

Before
Screen Shot 2019-07-31 at 13 56 42

After
Screen Shot 2019-07-31 at 13 56 16

### DIFF
--- a/src/utils/DayEventLayout.js
+++ b/src/utils/DayEventLayout.js
@@ -19,77 +19,9 @@ class Event {
     this.height = height
     this.data = data
   }
-
-  /**
-   * The event's width without any overlap.
-   */
-  get _width() {
-    // The container event's width is determined by the maximum number of
-    // events in any of its rows.
-    if (this.rows) {
-      const columns =
-        this.rows.reduce(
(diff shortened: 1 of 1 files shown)


Canuckaholic:
Would love to see this accepted and merged!

jquense:
@Canuckaholic sorry for missing this for so long but if it's still relevant we can get it merged in if you want

PR #1411, opened 2019-08-07
Added a work month view. Month calandar, excluding weekend days

jvineveld:
With this pull request there is a option for a work_month calendar.
This is the month calendar, but with weekend days hidden


src/WorkMonth.js
```diff
@@ -0,0 +1,83 @@
...
+    days.push(current)
+    current = dates.add(current, 1, unit)
+  }
+
+  days = days.filter(date => isWorkDay(date))
+
+  return days
+}
+
+class WorkMonthView extends MonthView {
```
jquense: can we change the Monthview api such that you can compose them normally instead of inheritance?

jvineveld: Hi Jquense,

Good idea, much cleaner indeed. Just didn't want to modify your cadebase.
I've now created a pull request that moves the logic into the month view component.

When the component is initialized with the 'workdaysOnly' prop, it will hide the weekends

jvineveld: Hi Jquense,

So you see some time to review the new pull request? 
Thanks.

Kind regards,
Jonas

PR #1412, opened 2019-08-08
Changed month view to accept 'workdaysOnly' prop

jvineveld:
To hide the weekends in the month view, just put the 'workdaysOnly' prop on the calendar when initializing. 

<Calendar
    events={...}
    workdaysOnly
  />


src/Views.js
```diff
@@ -1,12 +1,14 @@
 import { views } from './utils/constants'
 import Month from './Month'
+import WorkMonth from './WorkMonth'
```
jvineveld: these should still be removed, sry

PR #1416, opened 2019-08-14
Add resourceId to onSelecting callback

michalak111:
I've added `resourceId` to onSelecting callback since without its impossible to detect which column is selecting when we have resources.


michalak111:
@jquense any chance to merge it? or you need more insights from me why this would be useful?

jquense:
Thanks! Sorry for the delay

PR #1426, opened 2019-08-20
fix: remove time indicator when is null

lksilva:
There are a problem with time `.rbc-current-time-indicator` when timeIndicatorPosition is null and that happens when current hour is bigger or smaller that grid rage.

```js
  positionTimeIndicator() {
    const { min, max, getNow } = this.props
    const current = getNow()

    if (current >= min && current <= max) {
      const { top } = this.slotMetrics.getRange(current, current)
      this.setState({ timeIndicatorPosition: top })
    } else {
      this.clearTimeIndicatorInterval()
    }
  }
```

Then the `.rbc-current-time-indicator` stay on top of daycolumn grid.

>

### DIFF
--- a/src/DayColumn.js
+++ b/src/DayColumn.js
@@ -117,12 +117,18 @@ class DayColumn extends React.Component {
     } = this.props
 
     let { slotMetrics } = this
-    let { selecting, top, height, startDate, endDate } = this.state
+    let {
+      selecting,
+      top,
+      height,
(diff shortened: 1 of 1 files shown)


schester44:
thanks @lksilva. @jquense any chance we could get this merged?

schester44:
Was this resolved elsewhere? I'm still seeing this issue.

PR #1455, opened 2019-09-14
Drop warning

TrySound:
Maintainers of `warning` refuse evolution saying it's a 1:1 fork of
fbjs version. I tend to replace it with just console statements wrapped
in NODE_ENV check. Looks much simpler to me, does not require
additional tools like babel plugin and does not require commonjs module.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -6,7 +6,10 @@ env:
   es6: true
 rules:
   global-require: off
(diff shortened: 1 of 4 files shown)


TrySound:
I submitted the same PR in overlays
https://github.com/react-bootstrap/react-overlays/pull/488

PR #1457, opened 2019-09-16
prevent showing timeIndicator when indicator is outside of Calendar's min/max range

schester44:
I haven't tested other views yet but the timeline indicator is visible for the Day view when the time indicator is outside of the Calendar's set min/max values.

`this.state.timelineIndicatorPosition` will be `null` when min/max is set and the time indicator is before/after those values which causes the indicator to display at the top of the calendar.


schester44:
ah, looks like

Issue #1442, opened 2019-09-03
Incorrect styling (top & height) in day view


davehowson:
#### Do you want to request a _feature_ or report a _bug_?
Bug

#### What's the current behavior?
In the day view of Big Calendar, the events sometimes appear in incorrect positions due to incorrect top and height css values. This does not happen in events with times ending with 00 minutes (eg. 12:00, 17:00), but it happens when the events have odd times like ending on 12:29.

Codesandbox - https://codesandbox.io/s/react-big-calendar-issue-001e4?fontsize=14

![Capture](https://user-images.githubusercontent.com/26568721/64176539-4676c980-ce7b-11e9-86bf-31b44989a206.PNG)

Issue #1396, opened 2019-07-19
Current Time Indicator and short Events are offset by step-size number of minutes


stianeikeland:
#### Do you want to request a _feature_ or report a _bug_?

Bug / Regression caused by #1204

#### What's the current behavior?

https://codesandbox.io/s/react-big-calendar-example-feofz

1) Scroll down to current time.
2) Notice that the current time indicator is off by 15 minutes (one step size)

See related issue with events here: https://codesandbox.io/s/react-big-calendar-example-e1o2n
Events shorter than step-size minutes are offset.

#### What's the expected behavior?

The current time indicator should be at current time.

#### Likely cause

https://github.com/intljusticemission/react-big-calendar/blob/7619e592a6e880860c7fc2e9cbc6473ede386955/src/utils/TimeSlots.js#L135-L138

![image](https://user-images.githubusercontent.com/680040/61541038-20b67200-aa3f-11e9-8f6f-92fc16a151d0.png)

When calculating the time line position we hit the top branch of this ternary if. The timeslot step (ex: 15) is subtracted from the number of minutes since calendar min.

A calendar starting at 00:00, ending at 23:59, with current time 14:00 has rangeStart/End `14*60=840`. We should be at 58,33% of the day, but `(840-15)/(15*96)*100=57,29`. 15 minutes..

stianeikeland:
Seems like #1204 caused other regressions too. Create a 0 minute long event, and it will offset by negative stepsize minutes in the calendar.

dkenul:
@stianeikeland That PR cases the behavior outlined here as well.
https://github.com/intljusticemission/react-big-calendar/issues/1381

longsleep:
I suggest to revert #1204 the part introducing the breakage until it can be fixed without breaking time ranges for current time indicator and short events.

This does the trick for me:
```
diff --git a/src/utils/TimeSlots.js b/src/utils/TimeSlots.js
index 06e399d..ded31f4 100644
--- a/src/utils/TimeSlots.js
+++ b/src/utils/TimeSlots.js
@@ -132,10 +132,7 @@ export function getSlotMetrics({ min: start, max: end, step, timeslots }) {
 
       const rangeStartMin = positionFromDate(rangeStart)
       const rangeEndMin = positionFromDate(rangeEnd)
-      const top =
-        rangeEndMin - rangeStartMin < step
-          ? ((rangeStartMin - step) / (step * numSlots)) * 100
[...]

PR #1462, opened 2019-09-23
Fix for TimeSlots

MincedCutlet:
This small pull request fixes Issue #1442 

This issue occurs due to Pull Request #1204 

In the case where the range of an event is less than the step, this unmodified line will cause the position of the event to be positioned too early.

The bug, as demonstrated [here](https://codesandbox.io/s/react-big-calendar-issue-8mz4i)
When the `step` attribute is set to the minimum time, this fixes the issue, as seen [here](https://codesandbox.io/s/react-big-calendar-issue-i8vf9), but this forces a change in UI, which is not desirable.

While this current fix resolves the issue to correctly push any events past the calendar height back within, it has the unfortunate side effect of utilizing the `step` attribute to determine the height. For an arbitrarily large step, this will push it to the top of the large step block.

I welcome any feedback prior to modifying the tests.

### DIFF
--- a/src/utils/TimeSlots.js
+++ b/src/utils/TimeSlots.js
@@ -131,7 +131,7 @@ export function getSlotMetrics({ min: start, max: end, step, timeslots }) {
       const rangeStartMin = positionFromDate(rangeStart)
       const rangeEndMin = positionFromDate(rangeEnd)
       const top =
-        rangeEndMin - rangeStartMin < step && !dates.eq(end, rangeEnd)
+        rangeEndMin > step * (numSlots - 1) && !dates.eq(end, rangeEnd)
           ? ((rangeStartMin - step) / (step * numSlots)) * 100
           : (rangeStartMin / (step * numSlots)) * 100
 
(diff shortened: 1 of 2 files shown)


stianeikeland:
I can confirm that this PR solves events being offset, as reported in #1396

antondomratchev:
@MincedCutlet Can you please update the failing tests. I think the assertions should be `.toBeGreaterThanOrEqual` to pass but I am not sure if that's the desired assertion.

Npervic:
Is this getting anywhere? I'm having multiple problems displaying events...

stianeikeland:
Any chance of merging this?
It fixes this offset issue:

antondomratchev:
@stianeikeland Can you update the broken tests so the merge can happen. See here for details: https://travis-ci.org/intljusticemission/react-big-calendar/builds/588275219?utm_source=github_status&utm_medium=notification

mattmattvanvoorst:
Any chance of this being merged in the near feature?

jquense:
need to see if tests pass, i can't push to this branch and i can't restart manually.

Issue #1464, opened 2019-09-24
🎉🎉 Rearrange algorithm 🚑🚑


ckcks12:
#### Do you want to request a _feature_ or report a _bug_?

I want to request a feature. It's related to arrangement of events.

#### What's the current behavior?

![image](https://user-images.githubusercontent.com/12825679/65958181-c64b8000-e489-11e9-9246-bd351ae5c9bb.png)

Now, in this calendar, if there are more than two events having same start time, one of them **cover the other events** and I think it could be imporved by writing **new rearrange algorithm**.

#### What's the expected behavior?

![image](https://user-images.githubusercontent.com/12825679/65958425-51c51100-e48a-11e9-9c6c-a92118706b22.png)

It is expected to work as just like Toast UI Calendar.

Furthermore I'd like to to be a contributor for this awesome project. So if you don't mind can I make pull request for this feature? If so, I'm gonna optimize it and also write some explanation too!

Thank you.

=====

See how my algorithm handles plenty of events.

![](https://miro.medium.com/max/1633/1*w3nD80V2C9B6odJmm-Gnnw.png)

dmitrykrylov:
@ckcks12 I already created a pull request with similar updates on algorithm. It seems that maintainers didn't treat my proposal as a good solution and haven't approved it.
https://github.com/intljusticemission/react-big-calendar/pull/1405

Please create your pull request and maybe your changes will be merged. I am really interested in solving arrangement issues.

KhaledMohamedP:
@dmitrykrylov  & @ckcks12,

Does your solutions cover stacked events? ref: #1440

ckcks12:
@KhaledMohamedP  Yeah of course :) haha

![image](https://user-images.githubusercontent.com/12825679/65929803-5f54a980-e43e-11e9-8d04-77b0a65e531a.png)

![image](https://user-images.githubusercontent.com/12825679/65929765-45b36200-e43e-11e9-99f0-a5d4a744a8f0.png)
![image](https://user-images.githubusercontent.com/12825679/65929779-519f2400-e43e-11e9-832f-d64443215fdd.png)


above one is from 01:00 ~ 01:40 and the other one is from 01:40 ~ 02:10.

So, it worths $100 ? haha :laughing: :lol:

ckcks12:
@KhaledMohamedP 
You expect that those events are connected seamlessly? without the 2% padding between them?
[...]

Issue #1440, opened 2019-08-29
Alignment issues with short events!


KhaledMohamedP:
#### What's the current behavior?
If you book a 20 mins slot appointment back to back for one hour the alignment of these appointments doesn't align in top of each other (see screenshot below)

![Screen Shot 2019-08-29 at 2 55 58 PM](https://user-images.githubusercontent.com/5921853/63968247-45096380-ca6d-11e9-83ec-4be3f3c63f21.png)

Styling the library applies:

![Screen Shot 2019-08-29 at 2 56 23 PM](https://user-images.githubusercontent.com/5921853/63968245-433fa000-ca6d-11e9-8111-5703560810d3.png)

> Note this happens for 15 mins slots too* 😢 

#### What's the expected behavior?
Each event should stack in top of each other based on the start and end time. 

_I will add codesandbox tomorrow for anyone to test, for now I just wanted to get the discussion going and to see if anyone else is encountering this issue yet?_

beaugunderson:
hey, i'm Khaled's coworker--Canvas (where we work) will sponsor $100 for a fix to this issue :)

manhnguyenba:
I have the same issue. Is this an intended behavior or a bug?

manhnguyenba:
@KhaledMohamedP @beaugunderson I try to detect the issue and found that. The bug didn't happen with version 0.18.0. You can check the changelog, if your app no need to use latest version you can downgrade to 0.18.0 to fix this. Is that $100 for me? :))

KhaledMohamedP:
@manhnguyenba thanks for looking into this issue 🙏  

Someone needs to investigate what were the changes made between 0.18 and the current version that caused this behavior, and possibly try to opening a PR with a fix. 

We want to ensure this fix gets to everyone else using this library without using an older version of this library or forking it 😉 

Good luck 😃

PR #1473, opened 2019-10-01
Rearrangement Algorithm Implementation

ckcks12:
#1464 

From 
image

To
image

# Algorithm
Sort them by *Earlier -> Up, Longer -> Left.*
And then group them who are being overlapped each other.
And give number by order then stretch them to the same size except the last one. *last one could be greedy like he can eat all the rest place!*

# Performance

### original
image
### with new algorithm
image

### original
image
### with new algorithm
image

## About the Optimization

I leave some comments like **TODO** where further optimization would be effective.
But as you can see above Performance Metrics, even a thousand of events doesn't affect much.
in case you want to know how A THOUSAND of events looks like,
image

The source might be kinda dirty and hard to read because I hardly use the `map()` for loop which would be prettier. Honestly it causes a performance issue. Here you can find why I used `for`. 
https://github.com/dg92/Performance-Analysis-JS

# And...
I put some more like padding between events. like `2px` between them.
I think it's prettier but also worry bout it's hard coded.


src/DayColumn.js
```diff
@@ -13,6 +13,18 @@ import * as DayEventLayout from './utils/DayEventLayout'
 import TimeSlotGroup from './TimeSlotGroup'
 import TimeGridEvent from './TimeGridEvent'
 
+function dfs(node, maxIdx, visited) {
```
ckcks12: I put it outside on purpose bc of hoisting. and also I think it's better than declared inside of class.

KhaledMohamedP:
@ckcks12 Great job 👏 

This PR addressees #1440

beaugunderson:
@ckcks12 this looks great; we're very excited about getting this merged!

jquense:
thanks for doing this, hoping to get some time to review in the next few days!

KhaledMohamedP:
Looks like a lot of people are waiting for this PR to be merge! @jquense, did you get chance to review this PR?

Appreciate your time and help here 🙏

nikhilpingle:
Nice fix. I also need to have a couple of more enhancement in UI. I hope the community welcomes them as well.

src/DayColumn.js
jquense: don't abbreviate the name plz

ckcks12: Okayy.

examples/events.js
```diff
@@ -111,4 +111,52 @@ export default [
...
+    title: 'Cooking Class',
+    start: new Date(2015, 3, 14, 17, 30, 0),
+    end: new Date(2015, 3, 14, 19, 0, 0),
+  },
+  {
+    id: 23,
+    title: 'Go to the gym',
+    start: new Date(2015, 3, 14, 18, 30, 0),
+    end: new Date(2015, 3, 14, 20, 0, 0),
+  },
```
ckcks12: These are for overlapping situation

src/TimeGridEvent.js
```diff
@@ -42,10 +46,10 @@ function TimeGridEvent(props) {
...
         style={{
           ...userProps.style,
-          top: `${top}%`,
-          height: `${height}%`,
-          [rtl ? 'right' : 'left']: `${Math.max(0, xOffset)}%`,
-          width: `${width}%`,
+          top: stringifyPercent(top),
+          [rtl ? 'right' : 'left']: stringifyPercent(xOffset),
+          width: stringifyPercent(width),
+          height: stringifyPercent(height),
```
ckcks12: Sometimes the algorithm needs `calc()` so I just made it as `'string' || 'number'`. number will be followed by `%`.

Maybe you think it is no good that I didn't make it all `string` only. Frankly I tried it but you know that plain css string is quite annoying for testing, right? 

image

I also tried to `parse()` and `floor()` but... yeah... too much. so this is why.

ckcks12:
@jquense Hallo~

I made it as strategy pattern like pluggable design. :laughing: 

`DayEventLayout.js` -> `layout-algorithms/overlap.js` or `layout-algorithms/no-overlap.js`

And from the top of the library, the `Calendar`, I made new attribute `dayLayoutAlgorithm` and its prop type is `overlap` | `no-overlap` | `Function`.

and also I created a demo section for `no-overlap` algorithm.
image



Sooooooo hope you like it :eyes:

jquense:
just need a moment to try it out then gtg

KhaledMohamedP:
@jquense did you find any blockers?

Canuckaholic:
Very excited for this! Thanks for all the work on this everyone involved.

Dragomir-Ivanov:
Having overlap strategy as an option to Calendar would be very helpful. I also need all events visible at the cost of ugly presentation.

KhaledMohamedP:
@jquense Thanks for merging :) 
 
Can we please push this to npm and increment the package version as a `MINOR` update? 

Looking forward to this update 😄

PR #1489, opened 2019-10-11
needLimitMeasure must always true. to make 'x_more' works

icharge:
**AS-IS**: 
First time events loaded, month is same as default. So 'x_more' will not show. Until I try to resize window it come.


icharge:
I decided to make a Custom component by extend from Mount.js instead.

PR #1501, opened 2019-10-18
Make scrollToTime=00:00 working

### LINKED ISSUE #1474: BUG: scrollToTime 00h:00m doesn't work
Report a bug

scrollToTime prop does not work if the value's date has `00h 00m`

demo: https://codesandbox.io/embed/react-big-calendar-issue-scrolltotime-dp9zy
video: https://monosnap.com/file/dP5aEOh73L2K7Z7IfU2XJQtNnxdtJS

It mostly likely happens because of this check https://github.com/intljusticemission/react-big-calendar/blob/master/src/TimeGrid.js#L261 - scrollToTime is not being applied because `this._scrollRatio` is zero (false) (calculated at line https://github.com/intljusticemission/react-big-calendar/blob/master/src/TimeGrid.js#L275)

view scrolls to top when scrollToTime has been set to `00h:00m`

amok:
fixes #1474

### DIFF
--- a/src/TimeGrid.js
+++ b/src/TimeGrid.js
@@ -24,6 +24,7 @@ export default class TimeGrid extends Component {
 
     this.scrollRef = React.createRef()
     this.contentRef = React.createRef()
+    this._scrollRatio = null
   }
 
   componentWillMount() {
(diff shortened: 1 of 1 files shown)


src/TimeGrid.js
```diff
@@ -267,7 +268,7 @@ export default class TimeGrid extends Component {
   }
 
   applyScroll() {
-    if (this._scrollRatio) {
+    if (typeof this._scrollRatio === 'number') {
```
jquense: `this._scrollRatio != null`

amok: @jquense done

jquense:
thanks!

PR #1515, opened 2019-10-29
Refactor deprecated lifecycles

jetpackjarrett:
Since `componentWillMount` and `componentWillReceiveProps` are deprecated, my implementation was getting a lot of warnings. I went  refactored those components to use safe lifecycle methods

### DIFF
--- a/src/Month.js
+++ b/src/Month.js
@@ -35,12 +35,6 @@ class MonthView extends React.Component {
     }
   }
 
-  componentWillReceiveProps({ date }) {
-    this.setState({
-      needLimitMeasure: !dates.eq(date, this.props.date, 'month'),
-    })
-  }
-
   componentDidMount() {
     let running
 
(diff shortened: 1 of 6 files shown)


src/Month.js
```diff
@@ -60,8 +54,14 @@ class MonthView extends React.Component {
     )
   }
 
-  componentDidUpdate() {
-    if (this.state.needLimitMeasure) this.measureRowLimit(this.props)
+  componentDidUpdate(prevProps) {
+    if (!dates.eq(prevProps.date, this.props.date, 'month')) {
```
jquense: this sets should be collapsed and the state removed. this check just determines if the measureRowLimit should run, so instead of setting state, do the work instead

src/TimeGutter.js
```diff
@@ -18,8 +18,8 @@ export default class TimeGutter extends Component {
     })
   }
 
-  componentWillReceiveProps(nextProps) {
-    const { min, max, timeslots, step } = nextProps
+  componentDidUpdate() {
+    const { min, max, timeslots, step } = this.props
     this.slotMetrics = this.slotMetrics.update({ min, max, timeslots, step })
```
jquense: again this too late to update this, because it's consumed in render. it needs to update before render

jacksonrayhamilton:
Thanks for submitting this @jetpackjarrett.  Good to know that once this is merged the package will be quieter in recent versions and better prepared for React 17.

jacksonrayhamilton:
We will probably still want changes like the ones in this PR eventually.  While https://github.com/jquense/react-big-calendar/pull/1578 suppresses warnings about the deprecated methods, my understanding is that using “UNSAFE_” methods will still eventually cause compatibility issues with React.

Issue #1519, opened 2019-11-02
Enable semantic-cli for releasing to npm


KhaledMohamedP:
New Feature
---
Type: `Build setup`

### What's the current behavior?
The author has merged a bunch of small bug fixes and a giant new feature (ref #1473) to master, but users are patiently waiting for the new version to get released.

### Solution
Each PR that gets merged should be released automatically to npm & updating git release tags statment, relying on the git messages committed by the contributor. 

I will recommend using something like https://github.com/semantic-release/semantic-release

but open to alternative solutions!

jquense:
I don't probably have the time to set this up, but if you wanna work on a PR to add (probably via github actions) I'm open to it 👍

PR #1520, opened 2019-11-03
chore(github-actions): add semantic release and test workflow

KhaledMohamedP:
Improve the releasing setup 
---

Addresses issue #1519

Tools used: [Github Actions](https://help.github.com/en/github/automating-your-workflow-with-github-actions/about-github-actions) |  [Semantic Release](https://github.com/semantic-release/semantic-release)

This PR introduces 2 workflows
1. Tests Workflow: Runs in every PR and ensures that tests & linting are passing 
    - This flow is been tested here ✅[more details](https://github.com/KhaledMohamedP/react-big-calendar/commit/60b9478d75bf84e599bb44833ee8d0953a1741c8/checks?check_suite_id=293604078)
2. Release Workflow: Runs when anything gets pushed to master
    - Releasing involve pushing to npm registry and adding a release tag
    - This flow has not been fully tested, due to missing npm/github tokens ❌[more details](https://github.com/KhaledMohamedP/react-big-calendar/commit/a5790e34271d10e301abcbefc7e14d5176c4baab/checks?check_suite_id=293614141) 

Add Secrets
--- 
This action is required before merging:

Please add the following secrets needed for semantic release to enable publishing and git tags. 
1. `GITHUB_TOKEN` 
2. `NPM_TOKEN`

From the repo settings, navigate to `repo -> settings -> secrets` [more info here](https://help.github.com/en/github/automating-your-workflow-with-github-actions/virtual-environments-for-github-actions#creating-and-using-secrets-encrypted-variables)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,13 +18,10 @@ jobs:
         with:
           node-version: '12.x'
 
-      - name: Install yarn
-        run: sudo apt update && sudo apt install yarn
-
       - name: Install Dependencies
         run: yarn install
 
-      - name: Jest Unit Test & Linting
+      - name: Unit Test & Linting
         run: yarn test
 
       - name: Build
@@ -40,5 +37,5 @@ jobs:
             @semantic-release/changelog
 
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN_KEY }}
-          NPM_TOKEN: ${{ secrets.NPM_TOKEN_KEY }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,44 @@
+name: Release
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  release:
+    name: release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: '12.x'
+
+      - name: Install yarn
+        run: sudo apt update && sudo apt install yarn
+
+      - name: Install Dependencies
+        run: yarn install
+
+      - name: Jest Unit Test & Linting
+        run: yarn test
+
+      - name: Build
+        run: yarn build
+
+      - name: Semantic Release
+        uses: cycjimmy/semantic-release-action@master
+        id: semantic
+        with:
+          branch: master
+          extra_plugins: |
+            @semantic-release/git
+            @semantic-release/changelog
+
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN_KEY }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN_KEY }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,6 +1,7 @@
 name: Test
 
-on: [push, pull_request]
+on: [pull_request]
+
 
 jobs:
   test:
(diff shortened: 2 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,44 @@
...
+        uses: cycjimmy/semantic-release-action@master
+        id: semantic
+        with:
+          branch: master
+          extra_plugins: |
+            @semantic-release/git
+            @semantic-release/changelog
+
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN_KEY }}
```
jquense: `secrets .GITHUB_TOKEN` is automatically available in workflows

KhaledMohamedP: Thanks for the review @jquense, are you recomending removing this env or replacing with `secrets.GITHUB_TOKEN`?

.github/workflows/test.yml
```diff
@@ -0,0 +1,28 @@
+name: Test
+
+on: [push, pull_request]
```
KhaledMohamedP: Not sure if this should only trigger on `pull_request` rather than `[push, pull_request]`? also will this replace the current travis ci setup? please advice here

.github/workflows/test.yml
```diff
@@ -0,0 +1,28 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: '12.x'
+
+      - name: Install yarn
```
jquense: is there an action that already has yarn? this likely adds more time than needed to CI

.github/workflows/test.yml
jquense: would probably replace travis, both push and PR is ok i think? Maybe you just need `push`?

.github/workflows/release.yml
jquense: I think replacing with `secrets.GITHUB_TOKEN`

KhaledMohamedP: This is done, I found the doc in github 

https://help.github.com/en/actions/automating-your-workflow-with-github-actions/authenticating-with-the-github_token

FYI: I also updated the npm token key value to `NPM_TOKEN`

.github/workflows/test.yml
KhaledMohamedP: I removed the `push` and tested `pull_request` by itself and it worked on 

- Opening PRs
- Syncing or pushing changes to the PR
- Re-opining the PR

Noticed, when I had the `push, pull_request` it triggers two unnecessary jobs 

Release job will only trigger when anything gets pushed to `master`

.github/workflows/test.yml
KhaledMohamedP: There is this https://github.com/Borales/actions-yarn

I have updated the code to use this action. My only concern is that we won't have the `build/dist` directory available when pushing to npm registry. Not sure if we lose our place in the server, when using different github actions 🤔

jquense: I see, right. these aren't like docker images, where you have some env setup. Maybe it does make sense then to use it like you had it? Might be worth checking out how we do this in react-bootstrap: https://github.com/react-bootstrap/react-bootstrap/blob/master/.github/workflows/nodejs.yml

KhaledMohamedP: Ok thanks for sharing the link, it looks like we never needed to install yarn since `actions/setup-node@v1` does it for us. What waste of time 🤦‍♂️

Appreciate all your comments. The PR looks good now, much cleaner.

Let's put it to test!

KhaledMohamedP: Don't forgot to setup the `NPM_TOKEN` key in the settings

jquense:
lets see how it goes! thanks!

KhaledMohamedP:
Thanks for merging the PR @jquense, it looks like github actions is not supported for this repo 🤔 

Ref:
```
Workflows can't be executed on this repository. Please check your payment method or billing status.
```

After researching I found this useful info:
https://github.community/t5/GitHub-Actions/GitHub-Actions-workflows-can-t-be-executed-on-this-repository/td-p/38153

https://twitter.com/webKnjaZ/status/1194964915506831362

jquense:
yeah :\ I need to move this project to a different org, but i've not had a chance unfortunately

antoinerousseau:
the documentation is now missing:
https://intljusticemission.github.io/react-big-calendar/examples/index.html

PR #1554, opened 2019-12-18
Added dragging ability from the monthly Popup component

christianpatrick:
I've added in the ability to move events from the Popup in the Month view. Once the event has dropped onto the new day of the month, it will remove itself from the Popup and subsequently close that component.

I also added this into the `dnd` example to show the functionality.

### DIFF
--- a/examples/demos/dnd.js
+++ b/examples/demos/dnd.js
@@ -12,16 +12,39 @@ class Dnd extends React.Component {
     super(props)
     this.state = {
       events: events,
+      displayDragItemInCell: true,
     }
 
     this.moveEvent = this.moveEvent.bind(this)
     this.newEvent = this.newEvent.bind(this)
   }
 
-  moveEvent({ event, start, end, isAllDay: droppedOnAllDaySlot }) {
+  handleDragStart = event => {
+    this.setState({ draggedEvent: event })
+  }
+
+  dragFromOutsideItem = () => {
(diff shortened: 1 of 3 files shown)


christianpatrick:
@jquense, is this a feature you'd be alright with having? It fixes https://github.com/jquense/react-big-calendar/issues/1616.

jquense:
sorry @christianpatrick for not responding :| yes i'd be happy to merge this, can you push again to re-kick off the build?

cmmash:
Does this also allow to drag from one month to next or before?

quanghuyTiPiCi:
Why function reset not trigger after drag event from popup to outside, segment always !== null. Event is dragged overide another event.
image
image
Chrome 2023-08-01 11-42-34

cutterbl:
@quanghuyTiPiCi What do you mean by 'override another event'? Are you saying that the layout algorithm is no longer sorting the event correctly? It's hard for me to tell, without times being shown on your event layout.

quanghuyTiPiCi:
what i mean is that after dropping the event, EventRow component is not hidden. You can see the image above

cutterbl:
@quanghuyTiPiCi So, again, the little video isn't really clear. Explain to me exactly what it is that you think is supposed to handle, as opposed to exactly what it is that (you think) is happening.

PR #1560, opened 2019-12-26
Rename componentWillReceiveProps function

paul-cruz:
Rename componentWillReceiveProps to UNSAFE_componentWillReceiveProps avoiding warnings


schester44:
#1515 is open for this

jquense:
#1515 is not mergable at the moment so this is a fine approach, but it's incomplete, there are are more lifecycle methods that need to be renamed

PR #1562, opened 2020-01-02
Fixes `dayLayoutAlgorithm` prop with custom function

Dragomir-Ivanov:
Without this patch, custom functions are ignored.


SunnyTam:
this is a good suggestion.

jquense:
thanks!

PR #1574, opened 2020-01-24
feature: create luxon localizer

dctalbot:
Adds support for [luxon](https://moment.github.io/luxon/)


jquense:
@dctalbot can you push again to this branch to restart CI?

deevus:
What happened with this? I'm looking at using this but I need luxon support.

PR #1576, opened 2020-01-27
React deprecated items replaced

AdetunjiShennaike:
Replaced the componentWillMount and the componentWillReceiveProps to get rid of the warning/error message in the console as well as to help update before the new React v17 removes it's usage completely.

### DIFF
--- a/src/BackgroundCells.js
+++ b/src/BackgroundCells.js
@@ -19,18 +19,27 @@ class BackgroundCells extends React.Component {
 
   componentDidMount() {
     this.props.selectable && this._selectable()
-  }
-
-  componentWillUnmount() {
     this._teardownSelectable()
   }
 
-  componentWillReceiveProps(nextProps) {
+  componentDidUpdate(nextProps) {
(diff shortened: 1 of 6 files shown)


jquense:
Thanks for the attempt, but commenting out code doesn't really solve the problem here, that code is required, and needs to be migrated correctly to newer API's, see the other PR's on this for more context

AdetunjiShennaike:
> Thanks for the attempt, but commenting out code doesn't really solve the problem here, that code is required, and needs to be migrated correctly to newer API's. Moving it all to `componentDidUpdate` is not the right migration path, see the other PR's on this for more context

I made those comments to show what I changed. i actually changed the code to match what it should have been with the deprecated code. For some of the files changing it to ComponentDidUpdate and ComponentDidMount is fine. I do see what your saying about it  not being the right answer for files where those already exist. My implementation for those 3 files are band-aids at best, but the rest of the files should operate just fine with changing to didUpdate or didMount. I will try again and open a new pull request if it isn't solved by the time i figure it out. (I'm using this for a project so sorta need it to work haha)

PR #1578, opened 2020-01-29
Prefix all deprecated React lifecycle methods with UNSAFE

zsiglin:
There hasn't been any movement on #1515 or #1560 lately, so I thought I'd open this simple fix.

This PR prefixes all deprecated React lifecycle methods with the `UNSAFE_` keyword. In the long term these lifecycle methods should be refactored out, but in the interim there is no reason we need to continue to see the console full of warnings. 

For more context:
https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html

### DIFF
--- a/src/BackgroundCells.js
+++ b/src/BackgroundCells.js
@@ -25,7 +25,7 @@ class BackgroundCells extends React.Component {
     this._teardownSelectable()
   }
 
-  componentWillReceiveProps(nextProps) {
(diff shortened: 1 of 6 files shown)


zsiglin:
After further investigation, it looks like this is mostly a duplicate of #1497 which hasn't been approved either. I don't have a preference if we merge this one or the other as long as we can suppress the warnings. 😬

jquense:
thanks!

PR #1583, opened 2020-02-05
bump memoize-one and migrate new isEqual API

aginzberg:
The memoize-one team updated the isEqual API in version 5.0.0: https://github.com/alexreardon/memoize-one/releases/tag/v5.0.0

I'm taking that dependency bump and migrating the usage of isEqual here.


wheeler:
🙏

PR #1586, opened 2020-02-11
Update memoize-one dependency

cutterbl:
Memoize-one updated their function signature for their custom equality function, creating conflicts when using RBC with other components such as React-Select. This would cause serious issues when navigating around the calendar, especially when relying on events updates according to range change, primarily in the allDay events areas.

### DIFF
--- a/package.json
+++ b/package.json
@@ -134,7 +134,7 @@
     "invariant": "^2.2.4",
     "lodash": "^4.17.11",
     "lodash-es": "^4.17.11",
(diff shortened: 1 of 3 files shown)


cutterbl:
@TrySound I noticed that @aginsberg tried something similar with #1583 [slightly lower memoize-one version], that you approved, but that the build failed due to an issue with the build system? Any idea when we might get these changes?

TrySound:
Looks like setup had some problems

PR #1591, opened 2020-02-20
Fix an issue where <Show more +> is not displayed when the calendar first loads

zheyujie:
Fix an issue where <Show more +> is not displayed when the calendar first loads

### DIFF
--- a/src/Month.js
+++ b/src/Month.js
@@ -37,7 +37,7 @@ class MonthView extends React.Component {
 
   componentWillReceiveProps({ date }) {
     this.setState({
-      needLimitMeasure: true,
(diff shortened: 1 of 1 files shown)


src/Month.js
```diff
@@ -37,7 +37,7 @@ class MonthView extends React.Component {
 
   componentWillReceiveProps({ date }) {
     this.setState({
-      needLimitMeasure: !dates.eq(date, this.props.date, 'month'),
+      needLimitMeasure: true,
```
jquense: this not correct, now any component update will trigger a measure, which is expensive.

PR #1624, opened 2020-03-27
Fix duplicate time slots because of DST

milicamitk:
DST issue:
Duplicate time slot (3AM)

march-2020


hamez0r:
This is starting to become a problem again, can you please merge this? @jquense

hamez0r (approved):

PR #1633, opened 2020-04-08
Date-fns localizer - display dayFormat correctly

CraigMorton:
Previous format string displayed a day as '13 013' when using date-fns, rather than '13 Fri' as
RBC would with the moment localizer.


jquense:
thanks!

PR #1639, opened 2020-04-16
Hide indicator when current time is not in the interval

kvbrg:


src/DayColumn.js
```diff
@@ -163,7 +164,7 @@ class DayColumn extends React.Component {
             <span>{localizer.format(selectDates, 'selectRangeFormat')}</span>
           </div>
         )}
-        {isNow && (
```
jquense, line 166: should this check if the position is `!= null` instead of a new flag?

kvbrg: Sorry for delay! It's not new flag, i just added a checking for this flag. Based on flag name, i think this checking was just forgotten. If you use a position, you need to reset it (and add a new code), otherwise the indicator will get stuck at the bottom over time. So, I think these approaches are equivalent and the current edits are quite sufficient.

jquense:
thanks!

PR #1657, opened 2020-05-11
Update the agenda view to not show date color.

rwilson504:
Currently when you run the agenda view and utilize background colors for events the Date column for the Agenda view will display in the color of the first event for that day.  This happens because the Date field is a TD within the first TR and the TR is what is set with a color.  

Instead to alleviate this confusion I would move the styling to the Time and Event TD instead of putting it on the TR.  This way it will end up with the Date field having no color since setting any color to this on a day which could have multiple events of different colors is just confusing to users.

Final Result is that the Date column has no color associated to it unless defined in the CSS style sheet.

image

### DIFF
--- a/src/Agenda.js
+++ b/src/Agenda.js
@@ -113,14 +113,19 @@ class Agenda extends React.Component {
       return (
         <tr
           key={dayKey + '_' + idx}
-          className={userProps.className}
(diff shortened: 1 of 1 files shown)


rwilson504:
Ended up solving using css instead.

PR #1677, opened 2020-06-08
fixed return type for View parameter

catruzz:
The types declaration of onRangeChange function are wrong: onRangeChange function always returns also the view parameter (undefined when clicking on next/prev buttons but valued when changing view).
https://codesandbox.io/s/goofy-noether-6kgst

I've already asked for an update also on related typescript library here:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/45358

### DIFF
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -250,7 +250,7 @@ class Calendar extends React.Component {
     /**
      *
      * ```js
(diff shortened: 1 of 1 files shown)


catruzz:
> seems like more of a style change? these types are functionally the same no?

Well I noticed this error when typescript warned me about the view parameter that is totally missing on the DefinitelyTyped library and decided to better specify the different type here in the src ;)

jquense:
fair, thanks!

Issue #1620, opened 2020-03-25
refactor to hooks?


sibelius:
#### Do you want to request a _feature_ or report a _bug_?
feature

#### What's the current behavior?
it shows many warnings in strict mode

related to https://github.com/jquense/react-big-calendar/issues/1570
and https://github.com/jquense/react-big-calendar/issues/1481

#### What's the expected behavior?

it should work well without warnings in strict mode

would you accept a migration to hooks to fix all this warnings?

jquense:
sure, always open to PRs. They need to be thorough tho, i don't have much time to review for this lib!

GingerJams206:
I'm using this in a project of my own...has this gone anywhere since March of 2020?

jquense:
nope!

DanielUche:
Is this still an issue, I want to use this component in a new project but I fear this warning is an issue.

mdestafadilah:
Any alternative here?

jbcortez:
Well this is a huge bummer. Any plans on ever fixing this? Also, any alternative packages out there? I'd hate to have to build a calendar from scratch.

mdestafadilah:
I think someone created a hook version and done. nice job @AugustoCalaca  #1687 , please close this issued @sibelius. thx

andresmurillo13:
react_devtools_backend.js:4012 Warning: Using UNSAFE_componentWillReceiveProps in strict mode is not recommended and may indicate bugs in your code. See https://reactjs.org/link/unsafe-component-lifecycles for details.

* Move data fetching code or side effects to componentDidUpdate.
* If you're updating state whenever props change, refactor your code to use memoization techniques or move it to static getDerivedStateFromProps. Learn more at: https://reactjs.org/link/derived-state

Please update the following components: DayColumn2

royscript:
I had these error. Please, upgrade the code.

![image](https://user-images.githubusercontent.com/27907731/216846127-e12d6249-2d09-4374-9719-ba5d4e10a75a.png)

bu6n:
Don't want to sound harsh but this is an open source project and most people fixing bugs are doing so for free. We have a Slack space where we are discussing the next version of react-big-calendar. It's not as easy as "upgrading the code". If you want to help: https://join.slack.com/t/bigcalendar/shared_invite/zt-1ml1j99af-qIvqOfosMog1W7WxM0~j2Q

MichaelKleyman:
[...]

PR #1687, opened 2020-06-23
Feat/migrate react hooks

AugustoCalaca:
I started the migration to hooks api

* update react and react-dom to v16.13.1
* add eslint plugin enforces the rules of hooks
* refactor the fist file Agenda.js to hooks

The main idea of ​​this PR is to be the starting point for migrate the entire repo and finally close this issue #1620


.eslintrc
```diff
@@ -2,6 +2,9 @@ parser: babel-eslint
 extends:
   - jason/react
   - prettier
+  - plugin:react-hooks/recommended
```
jquense, line 5: you've got this over specified here. If we add the rules manually like you have below the extends isn't needed here

src/Agenda.js
```diff
@@ -10,75 +10,25 @@ import { navigate } from './utils/constants'
...
-  renderDay = (day, events, dayKey) => {
+  const renderDay = (day, events, dayKey) => {
     let {
       selected,
       getters,
       accessors,
       localizer,
       components: { event: Event, date: AgendaDate },
-    } = this.props
+    } = props
```
jquense, line 31: props should be spread out at the top level (the component function body), so each function doesn't have to respread out the ones they need

AugustoCalaca: great

hiroki-shiotsuka:
Still, neither `jquense/` nor `/intljusticemission` repo hadn't removed UNSAFE_componentWillReceiveProps() function yet. It won't work on strict mode.

PR #1689, opened 2020-06-24
feat(hooks): migrate TimeGutter.js to react hooks

AugustoCalaca:


jquense:
might be worth bunching these into a a single feature branch so we can release at once. You can still do individual PR's if that's helpful but make them to your feature branch instead of master.

AugustoCalaca:
great, I will do it this way in next PRs

PR #1696, opened 2020-07-06
DnD plugin: fully support custom event objects

algrant:
While trying to use the drag and drop plugin with ImmutableJS
events I was getting a number of errors which were caused by
my custom accessors getting some close but not quite JS objects
rather than my glorious ImmutableJS Maps /s

Eventually I narrowed it down to the `update` function in
EventContainerWrapper - which in essence is creating a new
version of the event object in state with modified start & end
times.

Unfortunately it was simply doing this by treating the event as
a regular JS object with start & end attributes.  This is then
consumed by the TimeGridEvent component, which tries to use the
accessors again and then we're in trouble...

The fix involves creating customisable `updateEvent` function which
follows the same path as the `accessors` -- not 100% sure it's the
best, but it works for my purposes!

Also it could be that I'm just abusing the small amount of power
I was given with the accessors, but it felt a little odd to run into
an issue after having it work so well elsewhere!

### DIFF
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -493,16 +493,16 @@ class Calendar extends React.Component {
      * (date: Date, resourceId: (number|string)) => { className?: string, style?: Object }
      * ```
      */
-	slotPropGetter: PropTypes.func,
-	
-	/**
-	 * Optionally provide a function that returns an object of props to be applied 
-	 * to the time-slot group node. Useful to dynamically change the sizing of time nodes.
-	 * ```js
-	 * () => { style?: Object }
-	 * ```
-	 */
-	slotGroupPropGetter: PropTypes.func,
+    slotPropGetter: PropTypes.func,
+
+    /**
+     * Optionally provide a function that returns an object of props to be applied
+     * to the time-slot group node. Useful to dynamically change the sizing of time nodes.
+     * ```js
+     * () => { style?: Object }
+     * ```
+     */
+    slotGroupPropGetter: PropTypes.func,
 
     /**
      * Optionally provide a function that returns an object of className or style props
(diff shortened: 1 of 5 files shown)


algrant:
Heya!

This fixes an issue I encountered -- but is also my first pull request on GH -- so if I'm missing anything please let me know.

I actually haven't tested that this works with a regular repo (but had a dream that maybe some of the automation testing would catch anything too buggy?) but it does work with mine.

Thought it'd be fun to save someone the hassle should they run into a similar issue.  Happy to add in any relevant docs / or tests if you can think of 'em.

Cheers & thanks for the maintaining this project.

Al

src/addons/dragAndDrop/WeekWrapper.js
```diff
@@ -61,8 +61,12 @@ class WeekWrapper extends React.Component {
   }
 
   update(event, start, end) {
+    const { accessors } = this.props
+    const updatedEvent = accessors.eventUpdater(event, { start, end })
```
jquense, line 65: ideally this shouldn't be an accessor, since it's not accessing anything. it makes more sense as a separate prop. It's also not updating the event it's creating a new pseudo event

src/addons/dragAndDrop/WeekWrapper.js
```diff
@@ -61,8 +61,12 @@ class WeekWrapper extends React.Component {
...
+    updatedEvent.__isPreview = true
```
jquense, line 66: It's a bit weird that we then mutate the object anyway to add this flag.

src/Calendar.js
```diff
@@ -754,6 +754,7 @@ class Calendar extends React.Component {
     allDayAccessor: 'allDay',
     startAccessor: 'start',
     endAccessor: 'end',
+    eventUpdater: (event, eventUpdates) => ({ ...event, ...eventUpdates }),
```
jquense, line 757: this isn't part of the core Calendar API it should go in the DND wrapper

PR #1700, opened 2020-07-11
Documentation: clarify the time unit for 'step'

th0rgall:
The `step` property accepts minutes but this is not explicitly mentioned. Although the default value of `30` suggests that minutes should be given, this is not immediately obvious.


jquense:
thanks!

PR #1727, opened 2020-08-13
Add background events feature

jmauzy:
Picking this up from #1161. We've forked this and believe it would still make a good feature. The previous PR was pretty stale, but conflicts should be resolved and updated examples are added.

Background events behave like standard events with respect to event handlers and titles/labels. The opacity of background events was lowered to .75 for more distinction.

image

A couple potential outstanding UI/UX issues:
- There is currently no display for background events in monthly or agenda view. Our current use case doesn't require this and we can't think of a better option for display than listing it in a similar manner as a standard event, which may be confusing.
- If a standard event overlaps the start time of a background event, the background event's label will be hidden.

If these are immediate needs for this feature, I'm happy to discuss potential solutions.


src/TimeGridEvent.js
```diff
@@ -39,24 +40,36 @@ function TimeGridEvent(props) {
...
   ]
 
+  const eventStyle = isBackgroundEvent
+    ? {
+        ...userProps.style,
+        top: `${top}%`,
+        height: stringifyPercent(height),
+        width: `calc(${width}% + 10px)`,
+        [rtl ? 'right' : 'left']: `${Math.max(0, xOffset)}%`,
+        opacity: 0.75,
```
jquense, line 50: this should be done via CSS so it's easy to override

src/TimeGridEvent.js
```diff
@@ -39,24 +40,36 @@ function TimeGridEvent(props) {
...
-          height: stringifyPercent(height),
-        }}
+        style={eventStyle}
         title={
           tooltip
             ? (typeof label === 'string' ? label + ': ' : '') + tooltip
             : undefined
         }
         className={clsx('rbc-event', className, userProps.className, {
+          'rbc-background-event': isBackgroundEvent,
```
jquense, line 72: hmm maybe it should be `isBackgroundEvent ? 'rbc-background-event' : 'rbc-event'` should give a bit more flexibility for styling, wdt?

src/Calendar.js
```diff
@@ -112,6 +112,27 @@ class Calendar extends React.Component {
      */
     events: PropTypes.arrayOf(PropTypes.object),
 
+    /**
+     * An array of background event objects to display on the calendar. B events objects
```
jquense, line 116: can we describe _what_ a background event is and how it's distinct from any other type of event?

src/TimeGridEvent.js
```diff
@@ -39,24 +40,36 @@ function TimeGridEvent(props) {
     </div>,
   ]
 
+  const eventStyle = isBackgroundEvent
+    ? {
+        ...userProps.style,
+        top: `${top}%`,
```
jquense, line 46: reuse `stringifyPercent`

jmauzy:
Thanks for taking the time to look this over! In addition to updating the prop description, I went ahead and updated the Background Events example to something more demonstrative of the behavior:

image

Technical differences: Background Events are not factored in when Events calculate their collision behavior and do not affect their layout.

Practical differences: Background Events allow the user to add a second 'class' of events that are more informational.

Common applications and use cases: Appointment availability vs scheduling, employee shifts vs tasks.

Currently background events that collide will just sit on top of each other. I can't really think of a more clever way to handle this from a UI perspective, and this would be a rare edge case in most (all?) of the use cases I can come up with. I'm happy to discuss solutions, though.

ronmara (approved):

vAskitsa:
This would be very nice feature.
In my case, whole day is filled by events and i have constantly overlapping situations whose make my view very ugly (i have lot of content in cell's so overlapping make those cell's unreadable). Background event's would fit for my case perfectly.

KhaledMohamedP:
This looks great, thanks @jmauzy for all the work. Can we test this somewhere?

threepears:
Was halfway through building something like this myself when I found this ticket ... would love to have this feature!  Appreciate all the work @jmauzy has put in!  Any word on when this might be approved and merged in?  Once I get a work thing or two off my plate, I can try to give this some review at some point ...

vAskitsa:
Is this going anywhere? This is very needed feature.

threepears:
@jmauzy I have some time this week to look this over, and would like to help get this over the finish line.  I'm pulled this branch down locally and am looking through your code ... 

To address your earlier points ... I think that adding background events to the month and agenda views is a good idea.  I know, at least in our business's use case, we would want that ... people here want to know they have availability slots that day, in addition to the actual scheduled events themselves.  I think as long as we leave them open to be colored as the user wants, they could distinguish them from other events.  I thought about maybe fading them like the ones in the week or day view, but then that's something the user can't easily override.

The "standard event overlaps the start time of a background event" thing is a little more difficult, admittedly.  I'm sort of poking around for some solutions currently to see what looks good.  Did you have any further thoughts on this?  Any other issues that should be addressed before getting this PR accepted?

jmauzy:
I'd be happy to catch this back up over the weekend if maintainers have the intention of merging it in. I've been using the branched version in production for some time now with no issues and basically abandoned this PR due to lack of response.

threepears:
@jmauzy Glad to hear you're willing to circle back around to this!  I went ahead and put in some code here locally to have background events show up in both month and agenda views, as I know that's something we need in our use case ... but if you feel it's not something everyone would want, we could put a boolean prop in or something that would give folks the option?

Also, I tried to solve the issue of events and background events starting at the same time, but just wasn't able to come up with a good solution.  I thought maybe about having the tops of events as tabs, but that would take some graphic work.  I also tried to push events in with padding a little bit (as shown below), but I'm not really pleased with that either.  Still thinking on it, though.  Another question: should we let people control the opacity of background events with a prop?  Some people may not want them faded, or maybe would want them faded more?

Screen Shot 2020-11-25 at 9 51 39 AM

JCardenete:
As others said above, it looks great and is a much needed feature

danmurphycp:
This feature would be super useful for something i'm working on and I would love to see it merged.

Issue #1646, opened 2020-04-30
Top part of a 24 hour event in Week/Day View is hidden behind column header


Khanbhai1990:
#### Do you want to request a _feature_ or report a _bug_?
This is a bug I come across
#### What's the current behavior?
 top part of a 24 hour event in Week/Day View is hidden behind column header. Which in turn is hiding the 'start' and 'end' of the event.  I have "showMultiDayTimes" as true.
<img width="1383" alt="Screen Shot 2020-04-30 at 3 01 34 PM" src="https://user-images.githubusercontent.com/29368029/80721863-96159f80-8ab3-11ea-91ed-166f8836df0a.png">

#### What's the expected behavior?
top in css should be 0% instead of being -2.08333%

Khanbhai1990:
if I remove deselect 'top' you can see the top half of the event is visible.
<img width="1373" alt="Screen Shot 2020-04-30 at 3 29 57 PM" src="https://user-images.githubusercontent.com/29368029/80723079-01ac3c80-8ab5-11ea-8f0c-84163909f7ac.png">

Khanbhai1990:
After further investigation I found that events that end after 11:00pm have the 'top' percentage calculated incorrectly 
<img width="1373" alt="Screen Shot 2020-04-30 at 3 29 57 PM" src="https://user-images.githubusercontent.com/29368029/80819378-b879ec80-8b89-11ea-84e0-69008831bded.png">
<img width="1383" alt="Screen Shot 2020-04-30 at 3 01 34 PM" src="https://user-images.githubusercontent.com/29368029/80819397-c0d22780-8b89-11ea-8530-989c5d55991e.png">

Khanbhai1990:
After further investigation I found that events that end after 11:00pm have the 'top' percentage calculated incorrectly
event ending before 11:00pm
<img width="1340" alt="Screen Shot 2020-05-01 at 4 26 03 PM" src="https://user-images.githubusercontent.com/29368029/80819500-f0812f80-8b89-11ea-8ba3-36bfa248f105.png">
same event ending after 11:00pm. Even though the event should start at 8:00pm according to the label. The top calculation is showing the event as if its starting 30 mins earlier.

Khanbhai1990:
@jquense 
hey bro, 
would love to know if this is an actual bug or something wrong with my setup. I have exhausted all my options to fix this issue related to events being bumped up 30 mins (if they end after 11:30p) but label showing different time.

[...]

PR #1732, opened 2020-08-20
Fix top part of 24hour event in week/day view

jyunhanlin:
ref: #1646 

`rangeEndMin > step * (numSlots - 1)` will cause the event(`00:00 ~ 23:31(to 23:59)`) shift to wrong position.


jquense:
thanks!

Issue #1200, opened 2019-01-31
Work in Strict Mode to avoid hitting a wall in upcoming React versions


tdd:
#### Do you want to request a _feature_ or report a _bug_?

It's a feature of sorts, more like making sure we can still use this by the end of 2019…

#### What's the current behavior?

Rendering a calendar inside React’s `<StrictMode>` wrappers yields a ton of deprecation / unsafe pattern warnings in the console:

- [Deprecated `findDOMNode()` usage](https://reactjs.org/docs/strict-mode.html#warning-about-deprecated-finddomnode-usage) (the modern Refs API should be used instead)
- [Unsafe lifecyle methods](https://reactjs.org/docs/strict-mode.html#identifying-unsafe-lifecycles) (typically `componentWillMount()` and `componentWillUpdate()`)

#### What's the expected behavior?

The codebase should try and move to the newer, cleaner, safer accepted ways to achieve the same results, in order not to find itself stonewalled when some upcoming React 16.x version finally strips these, some of which have long been deprecated.

My OSS/PR queue is already quite full for the next few weeks but I might try and give it a go around late Feb / March if the regular maintainers can't.

justingrant:
Shouldn't be closed. This should be fixed before `findDOMNode` is removed from React.

bboyjacks:
Yea, is it possible to update? Otherwise it will only be usable for people with some version of react and below.

niccai:
Yes, please reopen this issue.

jquense:
Happy to take a PR!

avindra:
Please re-open this issue.

This will become a major blocker with a new React release. Probably 17.

https://reactjs.org/docs/concurrent-mode-adoption.html#feature-comparison

justingrant:
FWIW, I've moved away from using react-big-calendar, partially for this reason.

avindra:
@justingrant is there another open source project that you've moved towards? For those of us who also want to be proactive instead of waiting indefinitely for removal of `findDOMNode` and other hacks that were clearly documented as things to avoid.

justingrant:
Hi @avindra - Nope, I ended up rolling my own plain-JSX solution.  My own implementation is limited and lacks most RBC features, but my app has pretty limited needs so it's good enough for now. I may consider switching back to RBC later but for now I've moved on.

PR #1739, opened 2020-09-01
<React.StrictMode>-compliance

erlendfh:
In reference to #1200 I've started on removing references to `findDOMNode()`. 

I've gotten ridd of all of them except one in `EventContainerWrapper`, which needs a larger rethink due to refs not working quite the same as `findDOMNode()` with components that use `React.cloneElement()` in `render()`.

I've also converted the draggable context from the legacy format to using `React.createContext()`. A larger rewrite would probably be necessary to bring this part of the code base more in line with modern React, but this works for now.

Finally, I've replaced `UNSAFE_componentWillReceiveProps()` with `componentDidUpdate()` and `getDerivedStateFromProps()` equivalents.

### DIFF
--- a/examples/App.js
+++ b/examples/App.js
@@ -168,5 +168,10 @@ class Example extends React.Component {
 }
 
 document.addEventListener('DOMContentLoaded', () => {
-  render(<Example />, document.getElementById('app'))
+  render(
+    <React.StrictMode>
+      <Example />
+    </React.StrictMode>,
+    document.getElementById('app')
+  )
 })
--- a/src/TimeGrid.js
+++ b/src/TimeGrid.js
@@ -2,7 +2,6 @@ import PropTypes from 'prop-types'
 import clsx from 'clsx'
 import * as animationFrame from 'dom-helpers/animationFrame'
 import React, { Component } from 'react'
-import { findDOMNode } from 'react-dom'
 import memoize from 'memoize-one'
 
 import * as dates from './utils/dates'
@@ -21,17 +20,14 @@ export default class TimeGrid extends Component {
     super(props)
 
     this.state = { gutterWidth: undefined, isOverflowing: null }
-
     this.scrollRef = React.createRef()
     this.contentRef = React.createRef()
+    this.gutterRef = React.createRef()
     this._scrollRatio = null
   }
 
-  UNSAFE_componentWillMount() {
-    this.calculateScroll()
-  }
-
   componentDidMount() {
+    this.calculateScroll()
     this.checkOverflow()
 
     if (this.props.width == null) {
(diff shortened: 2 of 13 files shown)


src/addons/dragAndDrop/WeekWrapper.js
```diff
@@ -268,27 +258,34 @@ class WeekWrapper extends React.Component {
...
-            className="rbc-addons-dnd-drag-row"
-            segments={[segment]}
-            accessors={{
-              ...accessors,
-              ...dragAccessors,
-            }}
-          />
-        )}
-      </div>
+      <DragAndDropContext.Consumer>
```
jquense, line 262: this should use `static contextType` where possible to avoid the extra component and instance assigning in the render fn

src/TimeGrid.js
```diff
@@ -64,28 +60,22 @@ export default class TimeGrid extends Component {
...
-    if (this.props.width == null) {
-      this.measureGutter()
-    }
-
-    this.applyScroll()
-    //this.checkOverflow()
-  }
-
-  UNSAFE_componentWillReceiveProps(nextProps) {
+  componentDidUpdate(prevProps) {
```
jquense, line 63: some of this should be in a getSnapshotBeforeUpdate hook, it's important that the measuring code run _before_ render

src/TimeGrid.js
```diff
@@ -21,17 +20,14 @@ export default class TimeGrid extends Component {
...
+    this.gutterRef = React.createRef()
     this._scrollRatio = null
   }
 
-  UNSAFE_componentWillMount() {
-    this.calculateScroll()
-  }
-
   componentDidMount() {
+    this.calculateScroll()
```
jquense, line 30: see comment below

erlendfh:
@jquense Thanks for the feedback, even learned a few new tricks in the process! Please take a look at the new commits when you have some time available. Any thoughts on the remaining findDOMNode() in EventContainerWrapper?

jquense:
> I've gotten ridd of all of them except one in EventContainerWrapper, which needs a larger rethink due to refs not working quite the same as findDOMNode() with components that use React.cloneElement() in render().

Not sure i understand specifically what you mean here. Adding a ref via cloneElement works well enough. Is the concern that it would override an existing ref, or that the ref may be passed to a function component? For the former case [`mergeRefs`](https://github.com/react-restart/hooks/blob/master/src/useMergedRefs.ts#L13) can handle it, and for the latter case, it's a fine restriction for users that their component _must_ accept a ref and pass it to the DOM node. As a migration path, I suggest something like: https://github.com/react-bootstrap/react-overlays/blob/master/src/safeFindDOMNode.ts which will fall back to findDOMNode if needed, which allows users to migrate at their own pace

erlendfh:
@jquense My mistake, I was of the impression that EventContainerWrapper expected an array of children, but it's a single child element, which made this a trivial fix. The `examples` project still has errors / warnings in the console due to an old version of Bootstrap being used, but I'm going to leave that for a separate PR as it seems at first glance to be a bit more involved than just changing the version number in package.json.

examples/App.js
```diff
@@ -168,5 +168,10 @@ class Example extends React.Component {
 }
 
 document.addEventListener('DOMContentLoaded', () => {
-  render(<Example />, document.getElementById('app'))
+  render(
+    <React.StrictMode>
+      <Example />
+    </React.StrictMode>,
```
jquense, line 174: lets not add this to the examples

jquense:
there are a few places where code was switched to `componentDidUpdate` that depended on running before render, e.g. slotMetric sets and the like.

luanraithz:
Couldn't we throw the `slotMetric` stuff inside the component state and use `getDerivedStateFromProps` instead of doing this manual re-initialization?

[Maybe a reference](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#what-about-memoization)

runtothefather:
any chances to get this merged?

ixnas:
Any updates on this? It's been a while, and it would be great to be able to use this library in StrictMode without errors.

cutterbl:
@erlendfh Thanks for your work here. I integrated your changes with current code, taking into account @jquense feedback. The only stuff I didn't use was your DnD context stuff, for now. This is in the v1.2.0 release.

erlendfh:
Wow, what a blast from the past! Awesome to finally get this merged though 👍🏻

Issue #1753, opened 2020-09-15
How can I make Events accessible to keyboard only users


SeandenDrijver98:
Hi there,
#### Do you want to request a _feature_ or report a _bug_?
I would like to request a feature as I'm currently making my implementation of this component more accessible to keyboard only users and need to implement a function when the user presses the Enter key on an event, similar to the ``onSelectEvent``.
#### What's the current behavior?
Currently Im unsure on how to implement this as the ``OnSelectEvent`` only fires for Click events.
I have tried using a custom component however the wrapper of the custom component with class ``rbc-event`` will first receive focus leading the user to believe they can select the event however the event is not selected when the user presses the Enter key. 

#### What's the expected behavior?
I have Forked this Repo and implemented an onKeyPressEvent prop which can be passed to the Calendar.
The implementation works exactly the same as how the ``onDoubleClickEvent`` was implemented and this prop will provide functionality to pass a function which should fire when the focus is on an event and the user presses any key. 

Im fairly new to working with this library so if there is another method to achieve this I will gladly give that a try.

I will create a PR with my code and attach it to this issue for review otherwise.

See screenshot of how this prop would be passed to the basic example.
![Screenshot 2020-09-15 at 10 57 49](https://user-images.githubusercontent.com/34751534/93189355-54278600-f742-11ea-89c2-dd3c29ba24f1.png)

PR #1754, opened 2020-09-15
Add onKeyPressEvent

SeandenDrijver98:
This PR implements an onKeyPressEvent prop which can be passed to the Calendar.
The implementation works exactly the same as how the onDoubleClickEvent was implemented and this prop will provide functionality to pass a function which should fire when the focus is on an event and the user presses any key.

see #1753 for more context

### DIFF
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -313,6 +313,15 @@ class Calendar extends React.Component {
     onDoubleClickEvent: PropTypes.func,
 
     /**
+     * Callback fired when a focused calendar event recieves a key press.
+     *
+     * ```js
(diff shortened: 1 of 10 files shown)


jquense (approved):

jquense:
thanks!

ehahn9:
I think onKeyPress was mistaken marked as a required prop?

```
index.js:1 Warning: Failed prop type: The prop `onKeyPressEvent` is marked as required in `DayColumn`, but its value is `undefined`.
    in DayColumn (created by TimeGrid)
    in TimeGrid (created by Week)
    in Week (created by Calendar)
    in div (created by Calendar)
```

PR #1760, opened 2020-09-25
Refactor of multiple components to functional components

cutterbl:
Refactored several components into functional components. One questionable item was the `renderRow` method of `TimeGridHeader`, where I found no references in code using the method. Also went through these and traded `let` for `const` in areas where the particular variable wasn't being later mutated.


cutterbl:
I can absolutely see where conversion of some of these other components to functional components can get problematic. Biggest issue being around measuring content for layout. I've switched most `findDOMNode` references in code to use a `ref`, either created from `createRef` or `useRef`, and hammer tested everything in Storybook while making my changes. The only one I couldn't do this with was the dragAndDrop `EventContainerWrapper`, as it has no true container to apply a ref towards (cloning `children`)

PR #1761, opened 2020-09-29
added component override for time gutter wrapper

zshannon:


src/TimeGutter.js
```diff
@@ -37,9 +39,9 @@ export default class TimeGutter extends Component {
 
   render() {
     const { resource, components, getters } = this.props
-
+    const { timeGutterWrapper: Wrapper = FallbackWrapper } = components
     return (
-      <div className="rbc-time-gutter rbc-time-column">
```
dhruvgoel92, line 42: I think we can keep this div as is and create a wrapper with default </> around it? If someone wants to override the class names, they will be able to.

src/TimeGutter.js
```diff
@@ -37,9 +39,9 @@ export default class TimeGutter extends Component {
...
+      <Wrapper className="rbc-time-gutter rbc-time-column">
         {this.slotMetrics.groups.map((grp, idx) => {
```
dhruvgoel92, line 45: We can pass this.slotMetrics as an argument to the wrapper to make it more powerful. A good use case for this might be for a wrapper to render "multiple" TimeSlotGroup to support multi-timezone time view.

dhruvgoel92 (commented):
A simple story using the wrapper might be useful. Let me know if I can help in anyway, happy to contribute.

dhruvgoel92:
@zshannon will it be okay with you if I collaborate on this PR and take it forward?

zshannon:
yup, i'm not going to add anymore here. good luck!

dhruvgoel92:
I have build on this here: https://github.com/jquense/react-big-calendar/pull/2236

I think we can close this PR.

PR #1774, opened 2020-10-14
update import in tutorial

fbigand:
I wanted to use React Big Calendar, but I was not able to make it work using [this tutorial](http://jquense.github.io/react-big-calendar/examples/index.html). The tutorial on github is up to date, but not this one. I make this update to avoid confusion for newcomers like me.


jquense:
thanks, we need to redeploy the docs, the file changed here is automatically generated.

PR #1780, opened 2020-10-21
Use main Reactiflux invite

vcarl:
The old invite was putting joiners into an error state due to the Code of Conduct gate! Not ideal


jquense (approved):
Thanks Carl!

PR #1789, opened 2020-11-02
Fixed a bug that the height of the column is broken when displayed in IE11

sawa-zen:
## What
I found a bug that shrinks the height of the TimeColumn when viewed in IE11.
Added `height: 100%;` as well as other parts, so that it won't collapse on IE11.

↓ Before
スクリーンショット 2020-11-02 14 09 08

↓ After
スクリーンショット 2020-11-02 14 09 50

### DIFF
--- a/src/sass/time-column.scss
+++ b/src/sass/time-column.scss
@@ -5,6 +5,8 @@
   flex-direction: column;
   min-height: 100%;
 
+  height: 100%; // ie-fix
+
   .rbc-timeslot-group {
     flex: 1;
   }


jquense (approved):

jquense:
thanks!

licia03:
I just updated my project to 0.28.2 and now I'm experiencing the same bug on Chrome.
Schermata 2020-11-02 alle 16 27 56

jason-krypton:
Also noticed this broke day view for us as well in `DAY` and `WEEK` view the absolute position gets messed up by the change to height.

Is there a planned fix?

PR #1795, opened 2020-11-05
Preserve time on horizontal resizing

jkhoang313:
When an event is resized horizontally (e.g. on the Month
View), keep the time of the event the same and only change
the date.

This also fixes the bug where you cannot resize an event to
a single day because the `dates.max` or `dates.min` would
force the event to occur at least 1 day after the
`start` or `end` respectively.

### DIFF
--- a/src/addons/dragAndDrop/WeekWrapper.js
+++ b/src/addons/dragAndDrop/WeekWrapper.js
@@ -138,6 +138,8 @@ class WeekWrapper extends React.Component {
     const { accessors, slotMetrics: metrics } = this.props
 
     let { start, end } = eventTimes(event, accessors)
+    let originalStart = start
+    let originalEnd = end
 
     let rowBox = getBoundsForNode(node)
     let cursorInRow = pointInBox(rowBox, point)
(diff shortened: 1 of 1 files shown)


jkhoang313:
In the `withDragAndDrop` file, it looks like resizing is supposed to preserve the times and only change the date, but that isn't the case (https://github.com/jquense/react-big-calendar/blob/master/src/addons/dragAndDrop/withDragAndDrop.js#L41). 

Right now, resizing an event automatically sets the changed value (start or end) to midnight of the new date. This also caused this bug (https://github.com/jquense/react-big-calendar/issues/1629) where you couldn't resize an event to a single day.

Current Behavior: Notice that the start date is set to midnight when resizing and that I can't make the event last only on Thursday.
current

New Behavior: Resizing will preserve the time of the event, and only change the date. I can make the event last only a single day.
new_behavior

jquense (approved):

jquense:
thanks!

nelsieborja:
It is not preserving the time on [this demo](http://jquense.github.io/react-big-calendar/examples/index.html#dnd).

PR #1796, opened 2020-11-05
Enforce `resizable` prop

jkhoang313:
Update the `resizable` prop to determine whether events
could be resized. Before, the prop could be supplied, but it
would not do anything.


jquense (approved):
Thanks!

PR #1800, opened 2020-11-06
Fix bug with resize segments not being removed

jkhoang313:
When an event is resized to span multiple rows, the preview
segments in the non-ending row would not be removed because
the `!pointInBox` would cause an early return.

This commit removes segments when the resize action finishes.

`onEventResize` is set to an empty function so we would expect the event to revert back to normal when the resize action ends.
**Bug**:
The preview segment in the 3rd and 4th rows persist after the resizing is finished.
Bug

**Updated**:
The preview segment in the 3rd and 4th rows are removed after resizing is finished. The event state is the same as the original state since `onEventResize` is an empty function.
Updated


jquense (approved):

jquense:
thanks!

Issue #1776, opened 2020-10-15
popper.js


noctua84:
I got a deprecated warning for popper.js@1.16.1 while installing react-big-calendar via npm.
Possibly a dependency that needs to be updated in the future.

ehahn9:
react-big-calendar uses an old version of react-overlays, which in turn uses popper 1.x. The current react-overlays uses @popper.js 2.x.

Alas, it doesn't appear trivial to upgrade to a newer react-overlays (I tried, but don't have the skills).

This turns out to be a more serious problem than one might think because it appears that using react-big-calendar in an app which also uses popper 2.0 breaks r-b-c drag & drop (!). So, for example, react-semantic-ui can't be used with r-b-c at the moment because of the popper version conflict (methinks, not 100% sure).

Which is a long way of saying +1 for the popper (react-overlays) dependency update ;-).

ehahn9:
fixed with 0.29 (thanks @jquense)

Also, my comment above is *incorrect* - turns out the problem with DnD is the addon's use of the legacy context API (see #1795). Working on that now.

PR #1819, opened 2020-11-29
fix: switch DnD to modern context API (was legacy)

ehahn9:
DnD addon was relying on legacy context api which finally broke with React 17 (couldn't resize events, etc.).

Fixes #1795. Related to #1776.


ehahn9:
Hmmm... might still have remaining v17 compatibility issues due to DnD's use of findDOMNode (deprecated). Working on that too.

jquense:
let me know when you want to merge 👍

ehahn9:
Thanks, @jquense -- the context switch was easy. The use of deprecated findDOMNode is harder (am working on that now).

The original DnD addon is looking like it need some serious love (very non-DRY, odd js idioms throughout) -- I might take a crack some rejuvenation throughout, but I also want to see it working with React 17 asap.

I'll ping you when I have something useful to review!

ehahn9:
Hi @jquense... I just realized that the [examples build](http://jquense.github.io/react-big-calendar/examples/) on github seem to be of a different vintage than those on master? (for example, the page's dropdown is missing more recently-added examples).

I've been working on finishing up the DnD refresh and thought I had introduced a regression but noticed that the examples built from master have the same probs (!).

I'll dive into this but wondering if you had thoughts on why the examples page is out-of-date? (or maybe I'm missing something...)

jquense:
@ehahn9 i just always forget to deploy new docs which is why they are out of date. we should make that automatic with the version release tho

ehahn9:
HI @jquense  - I [finally, sorry] had a chance to work on this. Many changes:

1. Use modern react context API (was legacy)
2. Uses refs rather than `findDOMNode` throughout the addin (although alas there is a bunch in the main rbc code, thus preventing react strict mode...)
3. Moved the DnD readme to the top-level of the addin (was buried in code comment)
4. Storybook: updated to v6, respect allDay in events
5. Fixed event propagation bugs (seen only in react 17)
6. Fixed a number of bounding box and fencepost errors in WeekView
7. Changed the HOC to extend, rather than repeat `propTypes` and `defaultProps`

I had hoped to do a LOT more (css modules for scoped addon classnames, and major cleanup of original implementation which is difficult to maintain) -- maybe I'll get back to that at some point!

I'm still a newbie, so apologies if this PR needs more work (happy to try!).

yann-combarnous:
@jquense, any chance you could review and possibly merge this PR? With DND resize broken in react 17, would be great to have it fixed.

ehahn9:
Hi @jquense  - oh apologies - I'm new to this and I wasn't even aware that this was waiting on me. Sorry for asking for a lesson in OSS, but what's the protocol when one is reviewing their own PR? Do I just click squash + merge or should others be reviewing first/also? thx.

jquense:
@ehahn9 generally you still want someone else to review it, but i've not been prompt on that sorry!

jquense:
There is a lot of different fixes and changes here that make it a bit hard to review all at once. Generally its better to split these things into separate PR's, e.g. one for event propagation bugs, one for context changes, etc.

examples/demos/dnd.js
```diff
@@ -55,7 +55,7 @@ class Dnd extends React.Component {
 
     const nextEvents = events.map(existingEvent => {
       return existingEvent.id == event.id
-        ? { ...existingEvent, start, end }
+        ? { ...existingEvent, start, end, allDay }
```
jquense, line 58: doesn't the `existingEvent` already have this?

src/addons/dragAndDrop/EventContainerWrapper.js
```diff
@@ -73,43 +59,31 @@ class EventContainerWrapper extends React.Component {
...
-    let end = dates.add(
-      currentSlot,
-      dates.diff(eventStart, eventEnd, 'minutes'),
-      'minutes'
-    )
-
-    this.update(event, slotMetrics.getRange(currentSlot, end, false, true))
+    const { duration } = eventTimes(event, accessors)
+    let newEnd = dates.add(newSlot, duration, 'milliseconds')
+    this.update(event, slotMetrics.getRange(newSlot, newEnd, false, true))
```
jquense, line 74: I'm not sure i understand what the changes in this file are fixing? can you walk me through it?

src/addons/dragAndDrop/EventContainerWrapper.js
```diff
@@ -271,8 +246,10 @@ class EventContainerWrapper extends React.Component {
       ),
     })
   }
-}
 
-EventContainerWrapper.propTypes = propTypes
+  render() {
+    return <div ref={this.ref}>{this.renderContent()}</div>
```
jquense, line 251: i don't think spliting renderContent out as a separate function adds much? maybe just have `render()`.

On the another note, I think it might more more sense to pass the ref to `children` instead of adding a new DOM element

src/addons/dragAndDrop/EventWrapper.js
```diff
@@ -32,28 +23,28 @@ class EventWrapper extends React.Component {
...
     this.context.draggable.onBeginAction(this.props.event, 'resize', 'UP')
   }
   handleResizeDown = e => {
     if (e.button !== 0) return
-    e.stopPropagation()
     this.context.draggable.onBeginAction(this.props.event, 'resize', 'DOWN')
   }
   handleResizeLeft = e => {
     if (e.button !== 0) return
-    e.stopPropagation()
```
jquense, line 45: It seems like removing these would create bugs, not fix them? Can you walk me through why this is necessary?

src/addons/dragAndDrop/WeekWrapper.js
```diff
@@ -79,42 +61,34 @@ class WeekWrapper extends React.Component {
     this.setState({ segment })
   }
 
-  handleMove = ({ x, y }, node, draggedEvent) => {
+  handleMove = (point, bounds, draggedEvent) => {
```
jquense, line 64: All the changes in this file are hard to follow (though i'm sure correct and needed). What is is going on here, and how should test that it works?

jquense (commented):
lots of great work, sorry i'm having a hard time following it!

examples/demos/dnd.js
ehahn9: The old story's `moveEvent` handler did't honor changes affecting `allDay`. This change purports to make it possible to demo moving events in and out of the allDay gutter - supported by DnD but previously not supported by the story as it ignored the new value for allDay.

src/addons/dragAndDrop/EventContainerWrapper.js
ehahn9: Absolutely right. I was having trouble forwarding the ref and took the cowards way out. I'll try to work on this.

src/addons/dragAndDrop/EventWrapper.js
ehahn9: Ok, this one is tricky. the Dnd addon uses rbc's selectable() mechanism to monitor mouse events before/during drags. selectable needs to see those events in order to work. The event wrapper stopped the propagation and so the selection logic was basically broken.

So first, I removed the stop propagation - that helped a lot.

But because the resize anchors are on top of the drag (move) anchor, clicking on a resize would also start a drag/move, which would abort the resize, etc.

So I added a check (see :42) to ignore this case:

```
    // hack: because of the way the anchors are arranged in the DOM, resize
    // anchor events will bubble up to the move anchor listener. Don't start
    // move operations when we're on a resize anchor.
    const isResizeHandle = e.target.className.includes('rbc-addons-dnd-resize')
    if (!isResizeHandle)
      this.context.draggable.onBeginAction(this.props.event, 'move')
```

Now the million dollar question is _why did this ever work?_ My only theory is that react 17 changed where the event listeners were placed and _maybe_ that subtle changed the even propagation?

I would love to back all of this out, but it seemed to be required both in reading the code and trying it.

src/addons/dragAndDrop/WeekWrapper.js
ehahn9: Purely idiomatic: all the other event handlers had the `(point, bounds)` signature (as does the similar code in `EventContainerWrapper`) so I was just conforming the idiom.

BTW, if you look closely, you'll see that `WeekWrapper` and `EventContainerWrapper` are in DIRE need of being DRY'ed out -- they basically do the same exact thing, with the main difference being the orientation of the event (horizontal or vertical). I spent a few days on that, but had to abandon due to time.

src/addons/dragAndDrop/EventContainerWrapper.js
ehahn9: Good catch. The problem I was addressing was that the addon would sometimes fail to update the event `end` on a move. This change always sets `end = start + duration` where start = `newSlot` (e.g. the target time slot at the end of the move).

examples/demos/dnd.js
ehahn9: I should have done this as a separate PR for sure! Newbie/learning.

ehahn9:
I'll wait to hear back from you (@jquense)  before working further on this PR. Agreed, I really should have done this in more atomic pieces (and will next time!). Alas, I find the DnD addon to be super hard to work on (maybe even a rewrite) and so my energy was flagging during this... apologies.

src/addons/dragAndDrop/WeekWrapper.js
ehahn9: ... oh, and the other changes (sorry, that's probably what you were asking about) - were all about making the date handling correct during drags - which had a number of probs, some of which might have been introduce with the (required) upgrade to date-arithmetic (which was out-of-date and causing some probs).

Happy to answer in more detail about these edits although my memory is fading a bit on this.

ehahn9: testing: I think the only way is via storybook/examples -- the code seems too intertwined IMHO for proper testing (definitely a good should a rewrite occur!)

src/addons/dragAndDrop/EventWrapper.js
jquense: sounds like you got it all handled 👍

src/addons/dragAndDrop/WeekWrapper.js
jquense: fine to defer the testing if there isn't a good way to write a unit test. I was thinking of the common-ish case where we fix one of these date bounds issues and introduce another one, partly because the calculations have no tests asserting their current behavior so easy to break stuff!

jquense (approved):
I would rename the PR so semantic release doesn't call all of this a "fix", might be worth hand rebasing the commits into different messages and then doing a "merge" instead of a "squash" but up to you.

ehahn9:
Thanks @jquense  - just to be sure I'm doing the right thing: are you suggesting I replace this PR with smaller, more focussed ones? (I presume there's a way to cancel this PR?) I'm happy to work on that! I know there's some time pressure for this due to v17 so I hope I'm not holding things up.

jquense:
>  just to be sure I'm doing the right thing: are you suggesting I replace this PR with smaller, more focussed ones? (I presume there's a way to cancel this PR?)

not quite, the suggestion was to rebase the commits in the current PR to something like
- fix: event propagation bugs
- fix: drag and drop bounds issues
- fix: update to modern context

etc, etc, and then merging instead of squash and merge so the commits stay intact on master. That said it all involves a bunch of Git-fu, that you may or may not be comfortable with, so fine to merge as is if you want

ehahn9:
Just rebased all the commits and reworded - lmk if this looks okay to you -- then I'm happy to merge (merge commit, no squash)

jquense:
very nice 👍

jquense:
🎉  very nice

nlevchuk:
@ehahn9 @jquense Thank you guys for your work

yann-combarnous:
@ehahn9 , @jquense , thx for this! Any chance dnd will be made to work wit touch events at some stage? 

Dragging works but the initial event selection is very hard to get to work, takes generally a few seconds to manage selection.

ehahn9:
@yann-combarnous - although more knowledgeable folks might disagree, I suspect the addon might be at the limits of maintainability at this point :-(. 

ps: I took a stab at [rewriting it from scratch as a completely separate project using react-dnd,](https://github.com/ehahn9/react-big-calendar-dnd) but ran into some architectural problems with resize animation.

PR #1850, opened 2021-01-27
feat: Add new Calendar prop: `doShowMoreDrillDown`

### LINKED ISSUE #1147: show more popup customize
How can we customize show more popup

### LINKED ISSUE #1527: popup = {false} brings up day view
Report a bug

When you set popup = {false}, it goes to the day view.  I don't want that, I have my own pop up and I want the view to remain in the same place.  

Expected behavior:
popup = {false}, simply turns off the baked-in '+more' modal.  Going to the day view should be independent, with its own boolean. Why are these two things intertwined?

teryngull:
Setting popup={false} doesn't let you control the onShowMore function properly as described in #1147  and #1527 , because it only goes to the drill down view. In order to fix this, I've added another prop called `doShowMoreDrillDown` that is defaulted to `true`. If true, it will do the same thing it currently does, and go to the drill down view. However, if you set this prop to false, it will instead just trigger the `onShowMore` function where you can then show your own customized popup as desired.
Fixes #1147 
Fixes #1527

### DIFF
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -388,6 +388,13 @@ class Calendar extends React.Component {
     views: componentViews,
 
     /**
+     * Determines whether the drill down should occur when clicking on the "+_x_ more" link.
+     * If `popup` is false, and `doShowMoreDrillDown` is true, the drill down will occur as usual.
+     * If `popup` is false, and `doShowMoreDrillDown` is false, the drill down will not occur and the `onShowMore` function will trigger.
+     */
+    doShowMoreDrillDown: PropTypes.bool,
+
+    /**
      * The string name of the destination view for drill-down actions, such
      * as clicking a date header, or the truncated events links. If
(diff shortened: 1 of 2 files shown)


teryngull:
@jquense can you take a look at this when you get a chance? we'd love to get this in so we can properly control and do our own 'show more' popup without also performing the drill down to the day view.

thomasmi (approved):
I looked at this PR and it seems like an unintrusive change, by default the code works the same and the new behavior is opt-in. My team would also greatly benefit from this change as we need to use a custom dialog. How can we get this merged?

pietrofxq (approved):

pietrofxq:
@teryngull please update commit to follow conventional commit standard and we can merge

teryngull:
@pietrofxq I had adjusted the commit message accordingly. Please let me know if I need do anything else. Thank you!

PR #1851, opened 2021-01-28
Add background events feature

kamildebicki:
The feature of handling background events was proposed in #1161 and then improved in #1727. 

The differences between this PR and #1727:
- c9ed1e8 Resolved conflicts between the master branch and the feature branch
- 16e2100 Reverted changes not related to the background events feature
- 0df77cb Small (-1 +1) fix of the Storybook story of this feature
- 559663e Small fix of background event width calculation

I haven't noticed any particular element of these changes that should require adding unit tests, but I'm happy to discuss and open to any suggestions.

**Checklist 🗒️**

<details>
  <summary>API documentation ✔️</summary>

Screenshot 2021-01-28 at 13 49 49

</details>
<details>
  <summary>Storybook story ✔️</summary>
Screenshot 2021-01-28 at 13 55 17

</details>
<details>
  <summary>Example ✔️</summary>

Screenshot 2021-01-29 at 12 55 14
Screenshot 2021-01-29 at 12 55 22
Screenshot 2021-01-29 at 12 55 30

</details>


mishelashala (approved):

src/DayColumn.js
```diff
@@ -174,9 +178,8 @@ class DayColumn extends React.Component {
     )
   }
 
-  renderEvents = () => {
+  renderEvents = ({ events, isBackgroundEvent }) => {
```
danmurphycp, line 181: Observation/Suggestion: It feels like the `isBackgroundEvent` property here may be confusing since the name is singular, but it applies to all events passed in.

How do we feel about keeping the current state where `renderEvents` doesn't accept arguments, and instead we get both `events` and `backgroundEvents` from the props within `renderEvents` and merge them into a single collection of events such that background events have a property `isBackgroundEvent` set to true?

src/TimeGridEvent.js
```diff
@@ -40,29 +41,44 @@ function TimeGridEvent(props) {
     </div>,
   ]
 
+  const eventStyle = isBackgroundEvent
+    ? {
+        ...userProps.style,
+        top: stringifyPercent(top),
+        height: stringifyPercent(height),
+        width: `calc(${width}% + 10px)`,
```
danmurphycp, line 49: Is the additional 10px here to compensate for the right margin on events?  It might be worth adding a comment to explain why we need this.

danmurphycp (approved):
LGTM overall!  A couple of suggestions/questions, but nothing major.

src/DayColumn.js
kamildebicki: Merging `events` and `backgroundEvents` will cause [`no-overlap`](https://github.com/kamildebicki/react-big-calendar/blob/841a43612e4e108cd5a33bcb6dbcd0e9cd46aac7/src/utils/layout-algorithms/no-overlap.js) algorithm to treat them as colliding events, and place them next to each other instead of one on top of each other.

src/TimeGridEvent.js
kamildebicki: Good point! Also, there was a small inconsistency causing these events to not render at 100% width of the column. I'll add a comment and fix this behavior.

Also, I'll update the examples screenshots.

jquense (approved):
LGTM to me.

jquense:
thanks folks looks great

PR #1859, opened 2021-02-06
Pass a className to selection

ThomasJamesCrawford:


bartvanandel:
What is the use care for this change? There is already [`eventPropGetter`](https://github.com/jquense/react-big-calendar/blob/2431026b5c8e983a1306c06af2f55ef38f1bca07/src/Calendar.js#L519) that takes the start and end date, as well as an `isSelected` boolean as properties. You can add additional class names using that function. Isn't that enough?

PR #1863, opened 2021-02-12
feat: add accessibility to month view

Nmarinsiruela:
While we're happy in general of being able to use this component for our work, the lack of accessibility has been quite the problem lately. This is not a full solution by _any_ means, but it can enable a screen-reader user to actually navigate through the table of events. Notice that this PR basically just adds roles and a few minor things; it should not crash anything!

One thing that I was unable to fix was the fact that the user is still unable to read the events per se; in our custom code we had to resort to move the focus once the situation was found from the front-end, but that's definitely no good!

Note: This mainly address the **Month** view. From what I've tested, the Week view will be basically invisible for the time being to screen-reader users!

### DIFF
--- a/src/DateContentRow.js
+++ b/src/DateContentRow.js
@@ -153,7 +153,7 @@ class DateContentRow extends React.Component {
     }
 
     return (
-      <div className={className}>
+      <div className={className} role="rowgroup">
         <BackgroundCells
           date={date}
           getNow={getNow}
@@ -175,6 +175,7 @@ class DateContentRow extends React.Component {
             'rbc-row-content',
             showAllEvents && 'rbc-row-content-scrollable'
           )}
+          role="row"
         >
           {renderHeader && (
             <div className="rbc-row " ref={this.createHeadingRef}>
--- a/src/DateHeader.js
+++ b/src/DateHeader.js
@@ -7,7 +7,7 @@ const DateHeader = ({ label, drilldownView, onDrillDown }) => {
   }
 
   return (
-    <a href="#" onClick={onDrillDown}>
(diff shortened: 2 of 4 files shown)


jquense (approved):

jquense:
thanks!

oc-jgm:
@Nmarinsiruela I know it's been a while since this was opened, but could you explain what you did to move focus onto the events? I was able to at least get the screen reader to read the events in the month view, but they are read after the row of dates it's in are read, and it adds columns which isn't ideal. Ideally the screen reader can move to the event(s) right after it's respective day is read.

Nmarinsiruela:
@jgarciamccausland it has been a _while_ since I did this. What we did was rely on a useEffect react hook and do it manually. I do not know if the library has changed since then and you can do it in another way, keep in mind we are still on an outdated version. But we did something like:

```      // Used to move the focus in case we deal with a date that contains events.
  useEffect(() => {
    if (currentDate && datesWithEvents.length) {
      const elementToFocus = document.getElementById(constants.AGENDA_MONTH);
      const dateContainsEvents = datesWithEvents.includes(
        currentDate.toISOString()
      );
      if (elementToFocus && dateContainsEvents) {
        elementToFocus.focus();
      }
    } else {
      // Used to handle the focus when tabbing.
      const dateToFocus = document.getElementById(currentDate.toDateString());
      if (dateToFocus) {
        dateToFocus.focus();
      }
    }
  }, [currentDate]);```

oc-jgm:
@Nmarinsiruela Took a long shot not expecting a anything so I greatly appreciate your response! Hopefully the rebuild of this package (if/when its ever released) takes into consideration accessibility and screen readers from the beginning.

PR #1882, opened 2021-03-10
revise some styles

zhaoce:
revise some styles, make it more compatible


minajiale (approved):

bartvanandel:
More compatible with what? Could you elaborate on why this change is needed?

zhaoce:
https://stackoverflow.com/questions/63475073/css-flex-basis-0-has-different-behaviour-to-flex-basis-0px

PR #1892, opened 2021-03-19
Dnd: Offset is not needed

tiwatson:
https://github.com/jquense/react-big-calendar/issues/1886

Drag and drop is currently using the very left edge of the calendar as an offset from the mousedown position. 
Resulting in unusable behavior. 

`getSlotAtX` is being sent an `x` value of `point.x - point.x - bounds.left`

It appears with the last few merged commits that `eventOffsetLeft` is no longer needed and removing it fixes all issues.

### DIFF
--- a/src/addons/dragAndDrop/WeekWrapper.js
+++ b/src/addons/dragAndDrop/WeekWrapper.js
@@ -65,7 +65,7 @@ class WeekWrapper extends React.Component {
 
     const slot = getSlotAtX(
       bounds,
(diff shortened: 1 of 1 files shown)


norbertdalecker:
+1

ckarsan:
Will this be merged soon I really need the draggable handles to work and want to upgrade to latest version

MattBlack01:
+1

Omkar-Jawalkar:
It's still giving the same problem

Issue #1909, opened 2021-04-10
Incorrectly shortening the multi-day event time-slot span by 1 day


croraf:
#### Do you want to request a _feature_ or report a _bug_?
bug

#### What's the current behavior?

- Check the demo's page http://jquense.github.io/react-big-calendar/examples/index.html#selectable
- select month view
- click+drag through 2 days to create an event 2 days long. The library will create an event 1 day long
- click+drag through 3 days to create an event 3 days long. The library will create an event 2 days long
- it works correctly with click and click+drag through 1 day only

react-big-calendar: v0.33.2 (I also tried locally using this git tag)
browser: Chrome, Version 89.0.4389.114 (Official Build) (64-bit)

#### What's the expected behavior?

- when click+dragging through more than 1 day timespan in month view it should correctly create events spanning through the days selected

croraf:
I think when selecting one date on month view the `start` should be on the 00:00 of that date and `end` on 00:00 on the next date. Same with the multiday select.

croraf:
(the same issue appears in day range select in Week view)

mackenziekira:
Thanks for fixing this @croraf  and @pietrofxq, I was running into the same issue.

Unfortunately the merged [pull request](https://github.com/jquense/react-big-calendar/pull/1910) breaks our use case, as we use a different calendar library for our react-native mobile app. With the new change, events created from the web app (using `react-big-calendar`) display 1 day too long on our mobile app, e.g.
WEB:
<img width="967" alt="Screen Shot 2021-07-06 at 5 38 00 PM" src="https://user-images.githubusercontent.com/7525998/124683056-14708d00-de81-11eb-82fe-8f7eb056f655.png">

MOBILE:
<img width="360" alt="Screen Shot 2021-07-06 at 5 39 44 PM" src="https://user-images.githubusercontent.com/7525998/124683117-2d793e00-de81-11eb-9788-8f2b2ae885fe.png">

Could we revert this change and instead solve the problem by providing a custom `endAccessor` prop? This is what I did in our app:

```
endAccessor={(event) => {
            if (isStartOfDay(event.end)) {
              return new Date(moment(event.end).add(1, "minute").toISOString());
            } else if (event.end) {
              return event.end;
            } else {
              throw new Error("Must provide an event end time");
            }
          }}
```

[...]

PR #1910, opened 2021-04-10
[Create event] fix end date on slot selection

croraf:
For end date in month-view slot-selection increase day by 1.

Fixes: #1909

EDIT:  The fix is applied for Month View, whole days Week view, whole day in Day view.

### DIFF
--- a/src/Month.js
+++ b/src/Month.js
@@ -303,10 +303,14 @@ class MonthView extends React.Component {
 
     slots.sort((a, b) => +a - +b)
 
+    const start = new Date(slots[0])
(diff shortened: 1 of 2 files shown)


pietrofxq (approved):

pietrofxq:
@croraf you mentioned in the issue you also applied the fix for Week view but this PR has only changes for month, do you want to push more changes?

croraf:
@pietrofxq The change I did in `src/TimeGrid.js` file handles that case. The problem for week was only when the selection of entire day (in the topmost row) happens. 

image

The same is in Day view for the whole day selection:
image

PR #1932, opened 2021-05-14
Improve the date-fns example in readme

fturmel:
Per date-fns/date-fns#2474, the `date-fns` example in the readme might've been wrong/broken. I am not familiar with your project, so please double-check that this suggestion is OK!

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,23 +78,23 @@ const MyCalendar = props => (
 )
 ```
 
-#### date-fns 2.0
+#### date-fns v2
 
 ```js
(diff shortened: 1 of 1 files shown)


pietrofxq (approved):

jquense (commented):
this example works but demostrates sub optimal setup. This fully depends on tree shacking to not include all of date-fns, which may or may not work. We should fix the example but continue to include cherry-picking

fturmel:
Sure, we can keep the cherry-picked imports if you prefer. Give me a minute and I'll update the PR.

bartvanandel (approved):

PR #1938, opened 2021-05-21
Adds support for using localizer math operators for timezone

lenehan2:
Issue: https://github.com/jquense/react-big-calendar/issues/1939


lenehan2:
This PR allows users to additionally request react-big-calendar uses the localizer when making date calculations in several key locations.

This currently is only supported with the moment localizer.

nitzanmo:
Could you merge this? I desperately need this solution.

nitzanmo (approved):

lenehan2:
@nitzanmo this is waiting on approval from a maintainer.  In the meantime I've added additional timezone support for the DnD package in this PR.  I am currently using my fork to get around these issues.

jquense:
Hey this is awesome, thanks for the PR. Ideally we would like a solution to this that is more universal, such as adding tz offset support to the underlying date math library. That might not be so hard? I am honestly not sure. Otherwise, we would at least want to update the existing localizers to support this before merging.

src/utils/dates.js
```diff
@@ -158,13 +180,77 @@ export function week(date) {
...
-  return dates.add(dates.startOf(new Date(), 'day'), -1, 'day')
+  return dates.add(startOf(new Date(), 'day'), -1, 'day')
 }
 
 export function tomorrow() {
-  return dates.add(dates.startOf(new Date(), 'day'), 1, 'day')
+  return dates.add(startOf(new Date(), 'day'), 1, 'day')
+}
+
+export function endOf(date, unit, localizer) {
```
jquense, line 194: Hmm I am not really a fan of overriding basically all of the date math library like this. There are too many forks in behavior that may change depending on the localizer and no explicit contract for how these methods should behave.

jquense: If we think it makes sense to put the logic in the localizer then we should probably design that directly. e.g. expose these methods on the localizer class and remove the helper. I am confident though that's the best approach though

PR #1963, opened 2021-06-23
fix: numGroups calculation

mrclickbits:
s. https://github.com/jquense/react-big-calendar/issues/1685

### DIFF
--- a/src/utils/TimeSlots.js
+++ b/src/utils/TimeSlots.js
@@ -19,7 +19,7 @@ export function getSlotMetrics({ min: start, max: end, step, timeslots }) {
     1 + dates.diff(start, end, 'minutes') + getDstOffset(start, end)
   const minutesFromMidnight =
     dates.diff(daystart, start, 'minutes') + daystartdstoffset
-  const numGroups = Math.ceil(totalMin / (step * timeslots))
+  const numGroups = Math.ceil((totalMin - 1) / (step * timeslots))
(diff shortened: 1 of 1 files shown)


pietrofxq (approved):

src/utils/TimeSlots.js
```diff
@@ -19,7 +19,7 @@ export function getSlotMetrics({ min: start, max: end, step, timeslots }) {
     1 + dates.diff(start, end, 'minutes') + getDstOffset(start, end)
```
pietrofxq, line 19: Wouldn't it be better to handle the +1 here? Won't removing it cause any problems?

mrclickbits: Although I'm not completely certain, I think the `totalMin` value is used for positioning items in the calendar as well while the `numGroups` is only used for figuring out the amount of slot groups. So the impact of changing the `numGroups` seems less risky to me.
Plus, my PR is inspired by the changes suggested in https://github.com/jquense/react-big-calendar/issues/1685 and an internal patch a colleague applied for fixing this behavior. And they both have the `- 1 ` expression in the `numGroups` value and its effect is working for us quite nicely.

jhnns:
@jquense What are your thoughts on this change? 😀

jquense:
i don't have any! happy to merge tho since someone approved it

Issue #1809, opened 2020-11-18
List "resourceHeader" in components props documentation


guillaumegaluz:
It's [listed](https://github.com/jquense/react-big-calendar/blob/master/src/Calendar.js#L697) as a prop of `components` but the [docs](https://github.com/jquense/react-big-calendar/blob/master/src/Calendar.js#L660) don't have it

PR #2014, opened 2021-08-26
Docs/fix all custom components documentation

khendrikse:
Me and my colleagues had a hard time figuring out what was possible with the library. I dove into the source code and realised that there was a bunch of documentation, but it was not rendering.

The docs that I moved from the top of the `Calendar.js` file: I don't know if these pieces of information are still relevant, but I added them to the intro. They were not rendering at all before.

I also added a comma so the custom component docs would render properly, and I added more info on the messages.

Might help solve: 

- #1809

Before:

image

image

After:

image

image

### DIFF
--- a/examples/Intro.md
+++ b/examples/Intro.md
@@ -6,11 +6,12 @@ _yarn:_ `yarn add react-big-calendar`
 
 _npm:_ `npm install --save react-big-calendar`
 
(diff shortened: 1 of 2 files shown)


ssusanto2706:
Could you make an example for the the custom components?

PR #2019, opened 2021-09-03
Update variables.scss to fix breaking change

DanielBailey-web:
Breaking change caused by deprecation of node-sass. 

New version sass (dart-sass) requires division be used differently [math.div()]. So updated this to use vanilla css.

I guess it is important to note that you only get this error if you upgrade to dart-sass from node-sass. However, this fix provides backwards compatibility.

### DIFF
--- a/src/sass/variables.scss
+++ b/src/sass/variables.scss
@@ -7,7 +7,7 @@ $cell-border: #DDD !default;
 
 $border-color: #CCC !default;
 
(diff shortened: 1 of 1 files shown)


pietrofxq (approved):

whl83111 (approved):

DanielBailey-web:
@cutterbl changed it to Math.div in #2079. Rip.

Issue #2011, opened 2021-08-22
Bug: Resizing 0h event with same start and end time


BlackRockSoul:
Resizing an event that has the same `start` and `end` dates cause the start date to become 'Invalid date'.

![image](https://user-images.githubusercontent.com/4317830/130369304-f8d7b138-c76d-46c4-a9f2-180dd41bd5a9.png)

This happens because this line of code adds 1 day to 0h events for some reason.
https://github.com/jquense/react-big-calendar/blob/d7c14f68616dce3072921250b644becca37a4f53/src/addons/dragAndDrop/common.js#L42-L43

And after that, the `closestSlotFromDate` function is not able to find the closest slot because the closest to the end is already the next day.
https://github.com/jquense/react-big-calendar/blob/d7c14f68616dce3072921250b644becca37a4f53/src/utils/TimeSlots.js#L107-L108

It just returns `undefined` instead of a correct slot - offset `-1` doesn't help at all.

![image](https://user-images.githubusercontent.com/4317830/130369254-5d0f642d-4711-49bb-839b-634d2a4e4c78.png)

In this case `offset` should be at least `-49`
I'm not sure if adding 1 day to an event is necessary in some particular cases, but I'd remove this - without this line, I'm not getting any errors in this case.

riazguzman:
Has this been resolved yet?

Issue #1199, opened 2019-01-30
Unable to resize start or end time of event spanning multiple days


kelsey93:
It would be nice to be able to have the ability to resize the start and end time of events that span over multiple days on the calendar. For example: an event that starts at 3pm on the 27th and ends at 5am on the 28th. Currently if you try to resize these times the resizeEvent returns either an invalid start date or an invalid end date and the event disappears from the calendar. Behaves similarly for events that span more than 2 days.

ZakRabe:
bumping so this issue doesn't get closed

anlijiu:
why won't fix ?

PR #2024, opened 2021-09-13
Resize overflow event cause `Invalid Date`

### LINKED ISSUE #1598: Cannot correctly resize or move multi-day events with showMultiDayTimes
Report a bug

Mutli-day events disappear when trying to resize on the week view. with `showMultiDayTimes` set to true. Viewable by trying to resize in the example repro:
https://codesandbox.io/s/rbc-c3vbm
From debugging so far it seems to be caused by invalid dates

Screenshot 2020-02-26 at 18 34 38
See in the above `start` is invalid

Screenshot 2020-02-26 at 18 30 49
Because it's calculating that `diffMins = 88`

Screenshot 2020-02-26 at 18 30 30
But there are only `49` slots for the day, so it returns `undefined`, which `dates.min` tries to convert to a date

I need to investigate further but I'm not sure if it should be using the end of the first day rather than the last to calculate `diffMins`?

The event is resized, with only the start or end adjusted to the new time

BlackRockSoul:
Resizing `startDate` of multi-day events or events that overlaps `max` of the calendar, make the `startDate` 'Invalid Date'

`dates.min` of `[Date, undfined]` returns `Invalid Date`.
Add fallback date from `closestSlotFromPoint` if no `closestSlotFromDate` found.

This should fix #1598, #1199, and partially the #2011

It could be reproduced on the 'draggable and resizable with showMultiDayTimes' page of the storybook, but for custom `min` and `max` calendar props I added 'draggable and resizable with event overflows the custom time period'

### DIFF
--- a/src/addons/dragAndDrop/EventContainerWrapper.js
+++ b/src/addons/dragAndDrop/EventContainerWrapper.js
@@ -81,9 +81,15 @@ class EventContainerWrapper extends React.Component {
 
     let { start, end } = eventTimes(event, accessors)
     if (direction === 'UP') {
-      start = dates.min(newTime, slotMetrics.closestSlotFromDate(end, -1))
+      start = dates.min(
+        newTime,
+        slotMetrics.closestSlotFromDate(end, -1) || newTime
+      )
     } else if (direction === 'DOWN') {
-      end = dates.max(newTime, slotMetrics.closestSlotFromDate(start))
+      end = dates.max(
+        newTime,
+        slotMetrics.closestSlotFromDate(start) || newTime
+      )
     }
 
     this.update(event, slotMetrics.getRange(start, end))
--- a/stories/DragAndDrop.js
+++ b/stories/DragAndDrop.js
@@ -1,8 +1,9 @@
+import moment from 'moment'
 import React from 'react'
 import { storiesOf } from '@storybook/react'
 import { action } from '@storybook/addon-actions'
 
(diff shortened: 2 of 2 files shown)


BlackRockSoul:
The event still shrinks to the end of the day when resizing, I can also add my workaround if necessary.

cutterbl:
@BlackRockSoul This may be part of why you're running into file conflicts, as the localizers now handle all date math. So `dates.min()` becomes `localizer.min` and `dates.eq()` becomes `localizer.eq()`. You seem to be working with an older version of Big Calendar...

BlackRockSoul:
@cutterbl Huh yes, I switched to my own fork and made some changes a while ago. And still using this old version. I'll switch to the original repo after you merge your fix 😃 
So yeah, your variant should fix both issues probably so I'm gonna try!

dhruvgoel92:
@BlackRockSoul @cutterbl so do you folks think https://github.com/jquense/react-big-calendar/pull/2138 is the right way forward? We have already done extensive testing on it and we are up to date with the latest version. Would love to get it merged and fixed for everyone!

BlackRockSoul:
@dhruvgoel92 I'm pretty sure it should work since it's 80% similar to my old hotfix and it working fine for a while already (not production yet, but anyway). So yeah, 100% I will switch from my fork to the original repo after this fix will be merged.

cutterbl:
Closing, as @dhruvgoel92 's change is now in 0.39.2

PR #2054, opened 2021-11-05
Trim Allday events on the TimeGrid based views

### LINKED ISSUE #2053: Trim Allday events on the TimeGrid based views
On the month view, the events are auto trimmed to fit the available cell space, and a "+{count} more" legend is appended. There's also a `showAllEvents` prop to display all events in a scrollable fashion.

In the TimeGrid based views, the allDay events (top row) are shown unlimited, no trimming is implemented, so eventually they'll push regular events down, and make them completely out of view.

Screenshot (9)
.

decimoseptimo:
Fixes https://github.com/jquense/react-big-calendar/issues/2053
also bump up node-sass version due to some error being thrown at former during compilation

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,25 +1,25 @@
 {
   "./dist/react-big-calendar.js": {
-    "bundled": 535473,
-    "minified": 165496,
-    "gzipped": 51033
+    "bundled": 537597,
+    "minified": 166210,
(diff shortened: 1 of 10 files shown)


decimoseptimo:
Adding the counterpart `showAllDayEventsMin` is also a good idea, because it is currently harcoded to 2.

Issue #2072, opened 2021-12-08
Bug with listeners in Selection.teardown method


vmstarchenko:
#### Bug report
I am not sure but visually there is a bug with `Selection.teardown` method. Everywhere in this file variable `_listeners` is used but in `teardown` method it doesn't starts with underscore:
`this._listeners = Object.create(null)`
vs
`this.listeners = Object.create(null)`

(https://github.com/jquense/react-big-calendar/blob/master/src/Selection.js#L47 vs https://github.com/jquense/react-big-calendar/blob/master/src/Selection.js#L102)

It might cause a problem with event listeners unbinding.

cutterbl:
@vmstarchenko Thanks again for pointing this out. It has now been resolved.

PR #2075, opened 2021-12-08
Correct listeners reference in the Selections.teardown() method

cutterbl:
This corrects an issue where the `teardown` was not properly referencing the internal `_listeners` property of the class.

This addresses Issue #2072


cutterbl:
@jquense Can you review this one please? Think this minor fix could repair some performance issues.

jquense (approved):

PR #2076, opened 2021-12-08
Add compressed CSS to package output

cutterbl:
Addressing Issue #2065 , this will add additional minified css files to the distribution package


jquense:
this feels a bit unnecessary to me. I think it's fine to leave minification up to the consumer, in general they are going to want that integrated with their bundler anyway

cutterbl:
@jquense While I agree that modern dev would bundle code, you do provide a umd version of the package, and there are developers who do not bundle code (unfortunately).

popuguytheparrot:
Always developers exclude all node_modules parse from build and use min files from package

jquense:
@cutterbl I don't feel that strongly but it's an added maintenance burden. The umd build exists solely to power usage in like codepens, tho code sandbox now makes that mostly unnecessary.

Issue #2079, opened 2021-12-12
node-sass


jwhite88:
Older version of node-sass will not will not allow full install of dependencies to run examples.  Installation eventually quits and displays errors.  I'm using NodeJS version 16.13.1.

Regards

jquense:
I don't think we support node-sass anymore, it's been deprecated and most people should switch to using dart sass

cutterbl:
@jquense It's the RBC current dependency for generating css from the sass files. I'll replace it and submit a PR.

cutterbl:
@jwhite88 This is now fixed in release 0.38.2.

PR #2082, opened 2021-12-20
fix: Replace deprecated dependency

cutterbl:
Replace node-sass with Dart Sass and correct autoprefixer script

#2079


cutterbl:
@TrySound Looks like I've duplicated this. Still trying to figure out the contribution protocols.

PR #2083, opened 2021-12-20
ci: Add browserslist config

cutterbl:
Moved from former `autoprefixer` script line for project level support

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,11 +68,6 @@
     "trailingComma": "es5",
     "bracketSpacing": true
   },
-  "browserslist": [
-    "ie >= 10",
-    "safari >= 8",
-    "last 2 versions"
-  ],
   "peerDependencies": {
     "react": "^16.6.1 || ^17",
     "react-dom": "^16.6.1 || ^17"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -68,11 +68,6 @@
     "trailingComma": "es5",
     "bracketSpacing": true
   },
-  "browserslist": [
-    "ie >= 10",
```
jquense, line 72: i don't actually think it works with ie10, even tho flex theoretically does

cutterbl: Good question. I took the list from the old `autoprefixer` call in the scripts. I did note, in the `.browserslistrc`, that we should review browser support. MS officially ends all support for IE 11 on all desktop OS versions on June 15th, 2022.

jquense: yeah, personally i don't support ie11 on any projects anymore. I'd be ok with officially killing support, I don't think it actually works well in those browsers anyway since we don't test (and there are open issues about it). It'd also open the door to using even better layout patterns like CSS grid, position sticky for fixed headers, etc.

jquense: another big win is we could use Intl APIs as the default localizer (did something similar in react-widgets) which would remove the requirement for using moment/globalize/etc

PR #2090, opened 2022-01-02
fix: Update deprecated-to-be syntax on less variable

andrevitalb:
Following [Sass' docs](https://sass-lang.com/documentation/breaking-changes/slash-div), using slash as a division will no longer be supported, this addresses said breaking change.


cutterbl:
@andrevitalb This was already resolved in Release 0.38.2

Issue #2089, opened 2021-12-30
Allday event display in week-view


mitsol:
Is there a way to define, when react-big-calendar shows up an event in the "allday" cell above the timeline in week-view? For my purpose it is a bit confusing, that allday implicitly means from [date] - 00:00:00 to [date+1 day] - 00:00:00 . 

 I know there is the showMultiDayTimes property, but this is just a boolean. It would be nice if I could dynamically define this property based on the event. Then i could use a allDay flag in each event to either show it in the allday cell or not.

decimoseptimo:
take a look https://github.com/jquense/react-big-calendar/issues/2021

PR #2097, opened 2022-01-08
2089 alldayevents

mitsol:
According to my #2089 issue I modified `showMultiDayTimes` property of the Calendar Component to allow also a function

So with my changes it is also possible to dynamically decide if an event should be displayed in allDay Row or as a normal event:

```javascript
showMultiDayTimes={event => {
    //if return false, the event is shown in allday-Row, if true it is shown according to start/end

    const startMoment = moment(event.start)
    const endMoment = moment(event.end)

    //if start and end is same day and start is 00:00 and end is > 23:00
    if (
        startMoment.isSame(endMoment, 'day') &&
        startMoment.hours() == 0 &&
        startMoment.minutes() == 0 &&
        endMoment.hours() == 23 &&
        endMoment.minutes() >= 0
    ) {
        return false; //show as allAay
    }
    return true;
}}
```

(In this example, the event is shown as `allday` if start equals to 00:00 and end is after 23:00)

for backwards compatibility it is also possible to leave the property as a boolean.

I used the `dndresource.js` for this example, so my pull request modified this example.

Unfortunately there is a small issue i recognized when i tested this: On the initial render, the allDay row is not shown. - However when I trigger a re-render (e.g. just by resizing the window), all is shown as expected. Maybe one of you has an idea on this, might be some render before initialize problem.


cutterbl:
@mitsol If you can update your branch from master, and move your examples into a story to update documentation, I'll review what you've done here.

PR #2099, opened 2022-01-12
Fix monthly display showing nothing with no events

omarwaleed:
Displaying the monthly calendar with no events opens up calendar entries with 0px height which shows nothing. Adding a minimum height to the row gives it a more (unbroken) look

### DIFF
--- a/src/sass/month.scss
+++ b/src/sass/month.scss
@@ -53,6 +53,7 @@
   flex: 1 0 0; // postcss will remove the 0px here hence the duplication below
   flex-basis: 0px;
   overflow: hidden;
+  min-height: 75px;
 
   height: 100%; // ie-fix
 


cutterbl:
@omarwaleed Can you provide a screenshot of the before and after of this change? I'm not following what it is you are trying to do here. What it is you are trying to resolve.

omarwaleed:
Screen Shot 2022-01-14 at 5 12 26 PM


this is what I get if I don't add a minimum height value in the global css file for my project

this is what I have in my stylesheet

```css
.rbc-month-row {
  min-height: 100px;
}
```

I added a smaller value but anyone can redefine it if the size doesn't suit them. Or we can instead introduce a custom variable passed to the component

cutterbl:
You did apply a `height` value to your container, right? It does state that in the documentation. I think that would resolve your issue, without applying code changes to the library.

omarwaleed:
You are actually right. Thanks

PR #2114, opened 2022-01-27
Removed unnecessary import

programandoconro:
A component was imported but never used


bartvanandel:
No longer applies. Files have been shuffled around and cleaned up.

PR #2108, opened 2022-01-22
docs: Complete Documentation Rewrite

cutterbl:
This is an exhaustive overhaul of documentation of Big Calendar. Here's the basic highlights:

- All documentation has been moved to Storybook
- Storybook has been upgraded
- All previous Examples have been updated and included
- Stories, with mdx docs and specific examples, have been created for every prop
- Addon stories, to include examples and prop stories, have been created for DnD
- Guide stories have been created to talk about Custom Styling, Custom Views, information about Big Calendar and Dates, and a guide on dealing with Timezones
- General documentation contribution guidelines have been included
- A GitHub workflow is included that will automatically build and deploy documentation to GitHub pages, after successful Release workflow (requires review and discussion with @jquense)

@jquense There is an issue with testing, after some of the dependency upgrades, that appear directly related to the `babel-preset-jason` preset in conjunction with Babel and Jest.

Submitting this now for review/help/feedback. 

**Note:** This does require a `yarn install` for fresh dependencies

PR #2121, opened 2022-02-05
fix: Correct variable name that gets passed on to EventWrapper so dragndrop ha…

dhruvgoel92:
Fix variable name that gets passed on to EventWrapper so dragndrop has access tocontinuesPrior and continuesAfter flags.
Based on these flags, drag n drop determine whether to show north-south anchors
for multi day events.

Before Fix:
* Multi day event shows an anchor on the top even if it was continued from before.
* Multi day event shows an anchor on the bottom even if it is continued after.

After Fix:
* Multi day event doesn't show north anchor and only south anchor if it was continued from before.
* Multi day event doesn't show south anchor and only north anchor if it is continued after.

### DIFF
--- a/examples/demos/dnd.js
+++ b/examples/demos/dnd.js
@@ -106,6 +106,7 @@ class Dnd extends React.Component {
         onEventDrop={this.moveEvent}
         resizable
         onEventResize={this.resizeEvent}
+        showMultiDayTimes={true}
         onSelectSlot={this.newEvent}
         onDragStart={console.log}
         defaultView={Views.MONTH}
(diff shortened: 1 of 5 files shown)


dhruvgoel92:
@cutterbl @jquense there were no contributing guidelines, let me know if there is any process that I am missing. We are using RBC extensively and you will likely see more PRs from my end.

Love how extendable the library has been so far!

cutterbl:
@dhruvgoel92 Thanks for taking care of this. Yes, we need contribution guidelines. Hoping to work on that soon. When you get a chance, checkout PR #2108 and give some feedback. Still a work in progress.

PR #2124, opened 2022-02-10
Pull public JS packages from npmjs.org instead of nexus

mqchau:
Nexus is deprecated


mqchau:
Sorry. I was trying to make PR on the fork

PR #2138, opened 2022-03-02
fix: Correct resize for multi-day event.

dhruvgoel92:
Bug

#### What's the current behavior
* When users try to resize a multi-day event, the resize just changes the event start/end as per that day's start/end.
 
* Resizing a multi day event should change just the start or end of the event and the start/end across the other days should remain consistent.


cutterbl:
Thanks @dhruvgoel92

PR #2140, opened 2022-03-07
Fix resizing event bug in Week & Day

OldDream:
The bug can be reproduce in this site: http://jquense.github.io/react-big-calendar/examples/index.html#dnd

### What's the current behavior:
In Day or Week view,  resize a event, keep the mouse pressed and move up, release the mouse after the cursor cross the top border of 12:00 AM, then the resize icon of this event will never show up untill you move or resize another event.

### What's the expected behavior?
The resize icon can show up in every condition.

### DIFF
--- a/examples/demos/dnd.js
+++ b/examples/demos/dnd.js
@@ -68,7 +68,7 @@ class Dnd extends React.Component {
 
   resizeEvent = ({ event, start, end }) => {
     const { events } = this.state
-
+    console.log('resizeEvent')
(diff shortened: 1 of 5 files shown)


cutterbl:
@OldDream The current documentation site is out of date, and does not use the latest version of Big Calendar. Can you recreate this bug in a CodeSandbox to review the issue you're trying to resolve here? [Use this as a base](https://codesandbox.io/s/react-big-calendar-example-v9wdyd).

PR #2162, opened 2022-03-25
fix: In day view with multiple resource，the header remain "selected" after mouse up.

OldDream:
The bug can be reproduce in this site (local dev server): http://localhost:9002/?path=/story/addons-drag-and-drop--example-2

What's the current behavior:
In Day  view with multiple resources, click the 'header' , hold the mouse down,  move the cursor out side the column, then mouse up. You would see one of the 'header' remain selected.

What's the expected behavior?
The 'header' should be unselected after mouse up.

I have some screen record here:
https://github.com/OldDream/bug-demos/blob/main/bug2.png
https://github.com/OldDream/bug-demos/blob/main/bug2.gif


cutterbl:
@OldDream Can you please update your branch with the latest changes from 'master' for testing please?

OldDream:
> @OldDream Can you please update your branch with the latest changes from 'master' for testing please?

OK, it's up to date.

cutterbl:
@OldDream In reviewing reported behavior, from current master and then after you updating to current master, it appears the issue was already resolved in an update. Can you pull latest master locally and verify?

Issue #2147, opened 2022-03-16
Unable to resize an event to include the last time slot (23:30 to 23:59)


ricky-ho:
### Reproduction link

https://codesandbox.io/s/react-big-calendar-example-forked-8xsd8f?file=/src/App.js

### Bug description

When resizing an event, the event cannot be resized to the last time slot (23:30 to 23:59) because the resize step is 30 minutes. The resize is prevented because this would essentially resize the event to the time slot (23:30 to 00:00) of the same day which is impossible. 

However, there is a workaround for this when only for dragging and dropping events specified in issue #1625
The workaround described in that issue works because the onEventDrop handler still fires for invalid event drops, which we can then modify the event start and end times.

I wanted to do something similar using the onEventResize handler, but the event handler does not fire on the invalid resize. 

### Expected Behavior

When resizing events to the last time slot, the end time should be auto-corrected or changed to 23:59, similar to how it is handled when creating an event in the 23:30 to 23:59 time slot. If creating events in the 23:30 to 23:59 time slot works, there should be a way for resize to work as well.

### Actual Behavior

Events cannot be resized to include the last time slot at 23:30 to 23:59

### react-big-calendar version

0.38.9

### React version

17.0.2

### Platform/Target and Browser Versions

Chrome

### Validations

- [X] Read the [docs](http://jquense.github.io/react-big-calendar/examples).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a react-big-calendar issue and not an implementation issue

dykim-sfive:
I am suffering same issue.
even I set max with ("2022-06-15T20:00:00"), there are no way to resize to last time slot 20:00:00.

PR #2192, opened 2022-06-16
2147 Fixed resetting event when resize action was done outside of container

LiniovasDovydas:
Fixes the issue #2147  when resizing an event past the root container, also solves drag creating an event that had the same issue.

You can observe the issues here:
[Resizable story](https://jquense.github.io/react-big-calendar/examples/index.html?path=/story/addons-drag-and-drop-props--resizable) - resizing event past the container causes the event to reset.
[Create events story](https://jquense.github.io/react-big-calendar/examples/index.html?path=/story/examples--example-2) - creating an event while dragging causes the event to reset

### DIFF
--- a/src/Selection.js
+++ b/src/Selection.js
@@ -323,16 +323,14 @@ class Selection {
       return this.emit('reset')
     }
 
-    if (!inRoot) {
-      return this.emit('reset')
-    }
-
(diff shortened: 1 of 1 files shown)


cutterbl:
@LiniovasDovydas Thanks for this.

Issue #2200, opened 2022-07-04
Display an error in console: TypeError: Cannot use 'in' operator to search for 'window' in undefined


ninhva:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

https://jquense.github.io/react-big-calendar/examples/index.html?path=/docs/examples--example-1

### Bug description

I got an error when running the application with "Day" view. (also basic demo: https://jquense.github.io/react-big-calendar/examples/index.html?path=/docs/examples--example-1) 

TypeError: Cannot use 'in' operator to search for 'window' in undefined

Please help me to check this. Thanks all.

### react-big-calendar version

0.40.4

### React version

17.0.2

### Platform/Target and Browser Versions

Chrome

### Validations

- [X] Read the [docs](http://jquense.github.io/react-big-calendar/examples).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a react-big-calendar issue and not an implementation issue

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

PR #2202, opened 2022-07-04
#2201 Fix gutter dom reference when measuring gutter width

### LINKED ISSUE #2201: Uncaught TypeError: Cannot use 'in' operator to search for 'window' in undefined

- [X] I confirm

### Reproduction link

https://jquense.github.io/react-big-calendar/examples/index.html?path=/story/addons-drag-and-drop--example-2

### Bug description

TimeGrid fails to measure gutter width due to gutter being undefined. 

### Expected Behavior

Gutter is measured

### Actual Behavior

Fails to measure the gutter due to invalid ref to the gutter dom element

0.40.4

### React version

^16.14.0 || ^17 || ^18

Chrome, Edge, Firefox

### Validations

LiniovasDovydas:
Fixes #2201 #2200 issues.

Seen in https://jquense.github.io/react-big-calendar/examples/index.html?path=/story/addons-drag-and-drop--example-2 day/week views.

### DIFF
--- a/src/TimeGrid.js
+++ b/src/TimeGrid.js
@@ -283,7 +283,7 @@ export default class TimeGrid extends Component {
     }
     this.measureGutterAnimationFrameRequest = window.requestAnimationFrame(
       () => {
-        const width = getWidth(this.gutter)
+        const width = getWidth(this.gutterRef.current.containerRef.current)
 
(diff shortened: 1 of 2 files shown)


cutterbl:
There were some minor issues with this, but it has now been resolved in 0.40.7

PR #2230, opened 2022-07-13
fix: calendar auto scroll while dragging event at top/bottom edge

### LINKED ISSUE #2231: [Request for Feedback] Calendar does not scroll if we drag event at top or bottom edge of container

When the user wants to move the event to a time that is currently outside of the view port of the calendar then the user needs to drag and drop the event multiple times to move it to the desired time.

### Suggested solution

When an event is being dragged at the top or bottom edge of the calendar then it will auto-scroll the calendar by adjusting the scrollTop property of the calendar container 

### Alternative

_No response_

### Additional context

_No response_

### Validations

- [x] I'm willing to open a PR

AmitRDesai:
Closes #2231 

Problem:
When the user wants to move the event to a time that is currently outside of the viewport of the calendar then the user needs to drag and drop the event multiple times to move it to desire time.

Solution:
For `WEEK` view of `DragAndDropCalendar`,
There is a new option added called `enableDragAutoScroll`. when it is enabled and an event is being dragged at the top or bottom edge of the calendar then it will auto-scroll the calendar by adjusting the `scrollTop` property of the calendar container 

default value of `enableDragAutoScroll` is `false`


cutterbl:
Thanks for doing this. I need to pull this code down and test it local. However, I don't think this should be a prop, but rather the default behavior. We should get feedback on that though. Why don't you write up an 'Issue' (more visible to more eyes) laying out the problem and possible solution, then edit your initial comment here by adding `Closes #{issuenumber}`.

Again, thanks for the PR. I think this would be really helpful.

src/addons/dragAndDrop/EventContainerWrapper.js
```diff
@@ -157,7 +184,11 @@ class EventContainerWrapper extends React.Component {
       const bounds = getBoundsForNode(node)
       const { dragAndDropAction } = this.context.draggable
 
-      if (dragAndDropAction.action === 'move') this.handleMove(box, bounds)
+      if (dragAndDropAction.action === 'move') {
+        if (this.context.draggable.enableDragAutoScroll)
+          this.updateParentScroll(parent, node)
+        this.handleMove(box, bounds)
+      }
```
dhruvgoel92, line 191: Though this logic seems reasonable here, if we were to emit a onDragging event as per: https://github.com/jquense/react-big-calendar/pull/2061/files it would be good to call updateParentScroll inside handleMove function. If onDragging returns false, we can choose not to call updateParentScroll.

src/addons/dragAndDrop/EventContainerWrapper.js
```diff
@@ -157,7 +184,11 @@ class EventContainerWrapper extends React.Component {
...
       if (dragAndDropAction.action === 'resize') this.handleResize(box, bounds)
```
dhruvgoel92, line 192: Can you check the experience for scrolling on resize too? When I am resizing and the event is on boundary of my scrollable window, it might make sense to scroll.

cutterbl:
@AmitRDesai Consensus seems to be making this the default behavior, rather than tying it to a prop. You might want to reach out to @dhruvgoel92 for any clarification on his feedback here.

src/addons/dragAndDrop/EventContainerWrapper.js
AmitRDesai: It is a good idea, I have checked it completely makes sense

src/addons/dragAndDrop/EventContainerWrapper.js
AmitRDesai: As of now, it will be fine since we need `parent` and `node` params for `updateParentScroll`

AmitRDesai:
I have made auto-scroll the default behavior for drag and resize. @cutterbl @dhruvgoel92

dhruvgoel92 (approved):

dhruvgoel92:
Looks good to me.

dhruvgoel92:
@cutterbl I have tested this, possible for you to take a look as well?

danielcosteirawella (approved):

cutterbl:
@AmitRDesai Can you update your branch with most recent changes from 'master' so we can retest this for inclusion please?

AmitRDesai:
I have rebase it master @cutterbl

PR #2232, opened 2022-07-14
fix(dnd): dont use classname

dhruvgoel92:
## Description
Use getAttribute('class') instead of className. className can be an object
 as per web standards.

## Bug Fix
* This fixes a bug when there is className is a SVGAnimatedString inside the event card.
Web standards for className : https://developer.mozilla.org/en-US/docs/Web/API/Element/className#notes

## Testing
* I have not added a story for this bug, as such it is outside the scope of this project. We are just adhering to the right standards.
* The existing story for DnD is functional with this bug including resize not triggering movement.


cutterbl (commented):
Think it better to just use the `classList` here

```
const isResizeHandle = e.target?.classList?.contains?.('rbc-addons-dnd-resize')
```

cutterbl:
@dhruvgoel92 I follow now. Had to dig into the full code to see what you were talking about.

PR #2239, opened 2022-07-23
fix: no-overlap layout algorithm

### LINKED ISSUE #2240: no-overlap layout algorithm issue

- [X] I confirm

### Bug description

`no-overlap` algorithm does not work correctly.

### Steps to reproduce:

1. Open https://jquense.github.io/react-big-calendar/examples/index.html?path=/story/examples--example-3
2. Select Day view.
3. Create event from 12:00 AM to 1:00 AM.
4. Create another event from 1:00 AM to 1:30 AM.

### Current effect
e1

### Expected effect

Both slots from the example should be full wide.

1.5.0

### React version

17.0.2

### Validations

lukbak:
Closes #2240 

### We have 3 cases to handle:
cases

### Current effect
e1
e3

### Effect after fix
e2
e4

### DIFF
--- a/src/utils/layout-algorithms/no-overlap.js
+++ b/src/utils/layout-algorithms/no-overlap.js
@@ -50,8 +50,11 @@ export default function ({
       const y3 = se2.style.top
       const y4 = se2.style.top + se2.style.height
 
-      // be friends when overlapped
-      if ((y3 <= y1 && y1 <= y4) || (y1 <= y3 && y3 <= y2)) {
+      if (
+        (y3 >= y1 && y4 <= y2) ||
+        (y4 > y1 && y4 <= y2) ||
+        (y3 >= y1 && y3 < y2)
+      ) {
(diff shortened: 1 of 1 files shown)


blugnomeXD:
I am testing this, and doesn't work. Could you provide thecode that your used for this test ?

lukbak:
@blugnomeXD just reproduce the issue from #2240. Please remember that this code is related to no-overlap algorithm.

blugnomeXD:
@lukbak  Thanks for your answer. Yes i am using no-overlap because i have the same issue that you reported.  Is enought to pass the prop dayLayoutAlgorithm='no-overlap' or i need something else?
Thanks

lukbak:
Hello @cutterbl. I would like to ask when we can expect this change? Do you have any comments?

lukbak:
@cutterbl @dhruvgoel92 Please take a look.

PR #2245, opened 2022-07-26
feat: use css custom properties on top of sass variables #2214

### LINKED ISSUE #2214: Usage of css custom properties on top of sass variables

Some of us don't want to use scss/sass, but still want to be able to customize the looks of the calendar. We might use other toolchains that make it hard to incorporate scss, such as tailwind css. It might also be desirable to change the style during runtime, if there is a dark theme for the website for example. 

### Suggested solution

I suggest you add custom properties to the `rbc-calendar` css class for all the scss variables, and make use of those everywhere instead of the scss variables. That means it would look something like this:

```
.rbc-calendar {
  box-sizing: border-box;
  height: 100%;
  display: flex;
  flex-direction: column;
  align-items: stretch;
  
  --rbc-out-of-range-color: #{$out-of-range-color};
  --rbc-out-of-range-bg-color: #{$out-of-range-bg-color};
  ...
}
```

And then, you'd use it like this:

```
.rbc-off-range {
    color: var(--rbc-out-of-range-color);
}
```

This way, you don't lose any backwards compatability for those using scss in their builds, while allowing people to use rbc without scss as well as changing themes easily in runtime. 

### Alternative

_No response_

### Additional context

_No response_

### Validations

PooSham:
Fixes #2214 

This adds the possibility to use [css custom properties](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties) to create custom styling instead of using sass variables. By using fallback values, there is still support for scss.*

*This differs from my proposal in #2214, but I realized this would be a cleaner approach that would require less work when adding new variables.


gnapse:
This is awesome! Any updates on whether this can be merged soon? It would be really helpful. Thanks!

sethmcleod:
Any hope of getting this merged and released? Our team would like to customize styles at runtime to support dark mode.

PooSham:
@cutterbl what do you say? There seems to be a demand for this. I can rebase master and clean up the formatting to minimize the diffs if you want.

fernando-deka:
We really need this!

cutterbl:
@PooSham I see a lot of potential here. I am concerned that this will be a breaking change for some. Think it would be very beneficial to

1. Get latest code into branch
2. Create some form of documentation, in storybook, that outlines the changes, and lists the available vars

Issue #1831, opened 2020-12-20
Warning: React.createFactory() is deprecated and will be removed in a future major release


ktmn:
Getting this warning with React 16.13.1: `Warning: React.createFactory() is deprecated and will be removed in a future major release. Consider using JSX or use React.createElement() directly instead.`

This seems to be the only place where it's used:

https://github.com/jquense/react-big-calendar/blob/master/src/addons/dragAndDrop/common.js#L2

teamzz111:
+1 awaiting response

gianlucca:
this issue has been fixed in 1.19.3 and can be closed

PR #2258, opened 2022-08-16
fix: remove deprecated React.createFactory. #1831

zhangmq:
checked with story "draggable and resizable with custom eventWrapper"


cutterbl:
@zhangmq Can you update your branch with the latest changes from 'master' so that we can test this please?

Issue #2261, opened 2022-08-23
Add onContextMenu


dandresfg:
### Clear and concise description of the problem

I just wanna have control over right clicks on slots.

### Suggested solution

`onContextMenu={(e) => onSelect && onSelect(event, e)}`

### Alternative

Also, there is another option to add a custom event handler **onContextMenu** at the top of the components

### Validations

- [X] Read the [docs](http://jquense.github.io/react-big-calendar/examples/index.html).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Would you like to open a PR for this feature?

- [X] I'm willing to open a PR

dandresfg:
> i was able to use onContext menu by customising dateCellWrapper and timeSlotWrapper

Can you add an example how?

meriemMat:
timeSlotWrapper: (timeSlotWrapperProps) => (
                        <SchedularCell
                            WrapperProps={timeSlotWrapperProps}
                            selected={selected}
                            scRef={scRef}
                            additionalContextMenu={additionalContextMenu}
                            useStandardAdd={useStandardAdd}
                            onEventAdd={onEventAdd}
                            view={view}
                            setView={setView}
                        />        

and SchedularCell looks loke this :
<div
            // style={style}
            onContextMenu={(e) => {
                if (
                    scRef.current?.props?.className !==
                    localRef.current?.props?.className
                )
                    scRef.current?.hide();
                scRef.current = localRef.current;
                localRef.current?.show(e);
            }}
            className={
                selected.start <= WrapperProps?.value &&
                WrapperProps?.value < selected.end
                    ? 'schedular-cell selected-event'
                    : 'schedular-cell'
            }
            onClick={(e) => {}}
        >
            <ContextMenu
                className={' overflow-hidden ' + WrapperProps?.value}
                model={scModel}
                ref={localRef}
                style={{ minWidth: '300px' }}
            />
            {WrapperProps.children}
        </div>
i used <schedularcell/> with dateCellWrapper too ,
its working like a charm,hope this guide you a little

PR #2262, opened 2022-08-23
add onContextMenu handler

dandresfg:
#2261


cutterbl:
I like the idea of an `onContextMenu`, but feel this needs to be it's own prop and method rather than a reuse of `onSelect`. Also need an a prop story with example in the documentation.

PR #2273, opened 2022-09-10
feat(add colors): add colors to cards (#2236)

ktechila:
Add custom colors to the event cards


cutterbl:
@ktechila This requires one to include styling with each event, rather than conditional styling based upon an event's properties. The [eventPropGetter](http://jquense.github.io/react-big-calendar/examples/index.html?path=/docs/props--event-prop-getter) approach already exists to handle this kind of scenario.

ktechila:
@cutterbl Thanks. I can use the same approach.

PR #2275, opened 2022-09-13
Feature/add custom formats

smarzism:
I found an issue with date formats in react-big-calendar when i tried to use jalali calendar.
I defined the last _formats_ as a default value for **momentLocalizer** (_defaultStringFormats_ and _defaultFunctionFormats_). 
so when develper wants to change for example _dayHeaderFormat_ from 'dddd MMM DD' to ' @#MMM DD - dddd #@' can pass the customFormats parameter to **momentLocalizer** and it will be used in calendar instead of _defaultStringFormats_


cutterbl:
@smarzism I'm confused. Are you saying our current way of [overriding the default formats](http://jquense.github.io/react-big-calendar/examples/index.html?path=/docs/props--formats) doesn't work?

PR #2293, opened 2022-11-03
feat: add custom timezone support for luxon localizer

hdoan-codaio:
Issue: In our codebase, we may have multiple calendar with different custom timezone settings. We want to be able to use the Luxon Localizer with a custom timezone, instead of using a global timezone setting for all Luxon DateTime instances. 

In this PR, I add a new option for the `luxonLocalizer` for timezone. Whenever we create a DateTime object, we pass the `zone` in as a parameter. 

Test Plan:

Open the luxon localizer stories, change the timezone from `America/Los_Angeles` to `America/Detroit` and observe that the calendar display the events in new timezone.

https://github.com/jquense/react-big-calendar/issues/2292

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,10 +29,10 @@ element has a height, or the calendar won't be visible. To provide your own cust
 $ git clone git@github.com:jquense/react-big-calendar.git
 $ cd react-big-calendar
 $ yarn
-$ yarn examples
+$ yarn storybook
 ```
 
-- Open [localhost:3000/examples/index.html](http://localhost:3000/examples/index.html).
+- Open [localhost:9002/](http://localhost:9002/).
 
 ### Localization and Date Formatting
 
--- a/src/localizers/luxon.js
+++ b/src/localizers/luxon.js
@@ -57,24 +57,29 @@ function fixUnit(unit) {
 // Luxon does not currently have weekInfo by culture
 // Luxon uses 1 based values for month and weekday
 // So we default to Sunday (7)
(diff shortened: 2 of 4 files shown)


cutterbl:
@hdoan-codaio I understand your use case, though think it is also the edge case. The default scenario we typically see is one calendar instance in a page, and typically those developers would want to set the timezone on the global instance, so it would apply to any instance of Luxon in same interface (for, say, an 'edit' modal, or date manipulations during selection event handlers). While what you've done is good, it would break those scenarios where someone needed that global setting. Better to have an optional, un defaulted, prop to the localizer and only implement if the option is present.

jhnns:
We have an application where you need to view calendar instances in different timezones (depending on the user you're currently viewing). This change would make it a lot easier for us. Our current hacky approach is this:

```ts
const CustomDateTime = Object.create(DateTime);

CustomDateTime.fromJSDate = (date) =>
  DateTime.fromJSDate(date, { zone });

const localizer = luxonLocalizer(CustomDateTime);
```

Afaict there's no cleaner or easier way to bring multiple TZ support to react-big-calendar.

@cutterbl I think both requirements could be fulfilled by just omitting the default parameter `zone = "local"` in this PR. This way, you could still configure it globally. What do you think?

qisaw:
hi @cutterbl I also have this same requirement. Are you still open to merging this PR? If so I can remove the default zone and create a new PR?

PR #2319, opened 2022-12-22
Update WeekWrapper => handleMove | fix bug rtl dragging

eeefg650:
Fix bug dragging an event in the RTL month view calendar gets confused to the wrong side

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,3 @@
-lib/
-dist/
 
 # Logs
 logs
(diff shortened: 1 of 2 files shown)


eeefg650:
Files that were changed:

https://github.com/jquense/react-big-calendar/blob/master/src/addons/dragAndDrop/WeekWrapper.js#L65

cutterbl (commented):
Do not remove these directories from .gitignore. These directories are created as part of the build process, and are placed in the package. We do not want them included in the repository, as it is an unnecessary duplication of code.

PR #2323, opened 2022-12-28
Feature: End time after midnight

jgullstr:
This pull requests allows day columns to wrap over to the following day if the Calendar's _max_ prop is set to a time before _min_. This is useful if you got events that commonly span past midnight. Another edge case this resolves is that "min = max = 00:00" allows you to create events that end at 00:00 the following day instead of 23:59, which seems to be the current limit.

image


ViljarVoidula:
Wonder if it gets merged at some point ?

Nikolian77:
It's just what I need.
In my company the day start at 6am and finish at 6am next day.
I hope this pull can be merged a day.... 
@jquense @cutterbl

PR #2366, opened 2023-02-14
fix: event overlap other event #2348

### LINKED ISSUE #2348: In the week view, events are completely covered by other schedules and are not visible.

- [X] I confirm

### Reproduction link

https://jquense.github.io/react-big-calendar/examples/index.html?path=/docs/examples--example-2

### Bug description

image

I have attached a screenshot of the reproduction I made on the demo page. 

As you can see in the image, I have created 1 to 6 events on the demo page library. 
The 5th event is scheduled from 12 to 2 PM, but it is completely invisible due to being covered by the 4th and 6th events. 

Additionally, the 2nd event is scheduled from 10 AM to 12 PM, but the part from 11 AM to 12 PM is not visible as it is covered by the 4th event.

### Expected Behavior

the 5th event should be visible and the 2nd event should display its full time range. To make it easier for you to understand, I've attached a screenshot of a similar schedule created in Google Calendar.

image

### Actual Behavior

_No response_

1.5.2

### React version

17.0.2

macOS Chrome 109.0.5414.87

### Validations

handjunyoung:
This is a PR to fix [issue #2348.](https://github.com/jquense/react-big-calendar/issues/2348)

1. Added the case where B's start slot covers A's end slot to the onSameRow(in src/utils/layout-algorithms/overlap.js) function.
2. Added ordered z-index values to leaves belonging to a row.

I'm currently using that commit version in production.
However, I am concerned that the addition of the z-index will have a negative impact on other parts.

Thank you.

### DIFF
--- a/src/TimeGridEvent.js
+++ b/src/TimeGridEvent.js
@@ -31,7 +31,7 @@ function TimeGridEvent(props) {
 
   let userProps = getters.eventProp(event, start, end, selected)
 
-  let { height, top, width, xOffset } = style
+  let { height, top, width, xOffset, zIndex } = style
   const inner = [
     <div key="1" className="rbc-event-label">
       {label}
@@ -52,6 +52,7 @@ function TimeGridEvent(props) {
       }
     : {
         ...userProps.style,
+        zIndex: zIndex,
         top: stringifyPercent(top),
         width: stringifyPercent(width),
         height: stringifyPercent(height),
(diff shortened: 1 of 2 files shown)


cutterbl:
@handjunyoung We've had some discussion about this in the Slack groups. While we agree that this is an issue, we also agree that using `z-index` to correct might have some adverse effects.

Ultimately the issue appears to be the order in which events are being rendered, as your Event 5 is being rendered before Event 4. Event 4, being the next in the DOM tree, is be default at a higher layer level. My suggestion is that you look at repairing the `sortByRender` method of the `overlap` algorithm, so that these events are properly sorted for this scenario. I can't look at this myself right now, but suggest this is where you start.

PR #2368, opened 2023-02-15
fix: correct treatment of boolean view in 'views'

rajeshmn47:

### DIFF
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -26,6 +26,11 @@ import mapValues from 'lodash/mapValues'
 import { wrapAccessor } from './utils/accessors'
 
 function viewNames(_views) {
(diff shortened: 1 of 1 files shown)


rajeshmn47:
fixing custom views by filtering out those with false values
Screenshot from 2023-02-15 16-33-50

cutterbl:
@rajeshmn47 Your code would mutate the original `views` variable passed in to the method. Maybe something like this would be better?

```js
function viewNames(_views) {
  if (Array.isArray(_views)) {
    return _views
  }
  const views = []
  for (const [key, value] of Object.entries(_views)) {
    if (value) {
      views.push(key)
    }
  }
  return views
}
```

PR #2371, opened 2023-02-20
fix: consistent multi-day event time ranges

### LINKED ISSUE #2370: Multi-day event date time is formatted differently from single-day events with date-fns localizer

- [X] I confirm

### Reproduction link

https://codesandbox.io/s/react-big-calendar-example-forked-w2sbo0?file=/src/App.js

### Bug description

The datetime on multi-day events seems to be formatted differently from single-day events when using the `dateFnsLocalizer`.

As an example, in the locale that is relevant for my use case (`nb-NO`), the datetime in multi-day events is formatted along the lines of `7:30pm.-` on the starting day and something like `- 2:00pm.` on the end day, even though I would expect it to be `07:30. -` and `- 14:00.` respectively.

On single-day events the formatted seems correct according to my expectations

See attached screenshots for examples from the linked code sandbox

| Multi-day (unexpected) | Single-day (as expected |
| ------ | ----- |
| Screenshot 2023-02-20 at 11 08 43    | Screenshot 2023-02-20 at 11 08 25  |

### Expected Behavior

Multi-day event datetimes are formatted differently from single-day ones.

### Actual Behavior

Multi-day event datetimes should be formatted exactly as single-day ones.

1.5.0

### React version

18.0.0

macOS Chrome 109.0.5414.119

### Validations

hnerdrum:
Fixes: https://github.com/jquense/react-big-calendar/issues/2370

### DIFF
--- a/src/localizers/date-fns.js
+++ b/src/localizers/date-fns.js
@@ -8,10 +8,10 @@ let timeRangeFormat = ({ start, end }, culture, local) =>
   `${local.format(start, 'p', culture)} – ${local.format(end, 'p', culture)}`
 
 let timeRangeStartFormat = ({ start }, culture, local) =>
-  `${local.format(start, 'h:mma', culture)} – `
+  `${local.format(start, 'p', culture)} – `
 
 let timeRangeEndFormat = ({ end }, culture, local) =>
-  ` – ${local.format(end, 'h:mma', culture)}`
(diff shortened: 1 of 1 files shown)


hnerdrum:
Anything in particular this is missing in order to get merged @cutterbl?

Issue #2314, opened 2022-12-14
When dragging an event with same start and end date, an extra day is added to the end date


nsds26:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

https://codesandbox.io/s/react-big-calendar-dnd-error-wwk1wf?file=/src/App.js

### Bug description

Hey there.

So we were using RBC for a while, but only now we added the DND feature. 
We were able to implement it, but ran into a problem.
When there is an event with the exact same date and time, moving it will add an extra day to the end date. I tried adding 59 seconds to the end date to see if I could mitigate this, but the error persists. Adding a minute to the end date seems to work, but its not ideal since it will show up on the event description.
Am I doing something wrong, or RBC does not allow the start and end dates to be the same?

Thanks a lot in advance!

### react-big-calendar version

1.5.0

### React version

17.0.2

### Platform/Target and Browser Versions

Chrome

### Validations

- [X] Read the [docs](http://jquense.github.io/react-big-calendar/examples).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a react-big-calendar issue and not an implementation issue

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

bu6n:
Not sure why you would have events with the same start and end dates.  

The problem is this line:
https://github.com/jquense/react-big-calendar/blob/master/src/addons/dragAndDrop/common.js#L39

What you could do is replace it with:
```js
var isZeroDuration = localizer.eq(start, end, 'minutes') && localizer.diff(start, end, 'seconds') === 0;
```
and make your event's duration some seconds.

PR #2387, opened 2023-03-29
Moving events lasting less than a minute

MetaFactoK:
Fix issue #2314 
When working with events that can have only a few seconds difference between end and start, dragging is visually broken because the end date is set the next day.


NiharR27 (approved):

cutterbl (changes requested):
Is there an example, in Storybook, where I can test this functionality?

teamzz111:
+1 it's important

Ziobzee:
@cutterbl add this correction please, it's very important

cutterbl:
Someone will have to update the branch with the latest updates from master, and provide working examples in Storybook for proper testing, before this can be approved.

Issue #2386, opened 2023-03-24
Support "show more" in "All day" in week view


gnapse:
### Clear and concise description of the problem

When you have too many all-day events, it can become cumbersome in the week view:

<img width="978" alt="CleanShot 2023-03-24 at 10 44 40@2x" src="https://user-images.githubusercontent.com/15199/227537876-992835ae-407f-49d8-b55e-86491e2693d1.png">

### Suggested solution

Add a "show more" that would show a popup, just as with day cells in month view

### Alternative

I added this CSS rule:

```css
.rbc-allday-row {
  max-height: 110px;
  overflow: auto;
}
```

And it mitigates the issue, but not a great UX:

![CleanShot 2023-03-24 at 10 47 11](https://user-images.githubusercontent.com/15199/227538423-83a1e448-adce-4bbc-9e28-57566d26ddc1.gif)

### Additional context

Full Calendar seems to support it:

<img width="1123" alt="image" src="https://user-images.githubusercontent.com/15199/227539171-21613490-2967-4882-955b-bb0e4a6576ba.png">

### Validations

- [X] Read the [docs](http://jquense.github.io/react-big-calendar/examples/index.html).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Would you like to open a PR for this feature?

- [ ] I'm willing to open a PR

fornesarturo:
Please review the changes proposed, I've also been looking for this feature before

PR #2393, opened 2023-04-13
feat(time-grid): allow consumer to set allDayMaxRows (#2386)

fornesarturo:
* Addresses #2386 
* Currently react-big-calendar is not limiting the amount of events in the All day row displayed in Week and Day view, if this row grows it may also cover the events and prevent the user from seeing their normal events.

### In this commit:
* Add support for displaying Show more text in TimeGrid views via allDayMaxRows prop.
* Prevent Background cells from capturing click on show more button.

### DIFF
--- a/src/BackgroundCells.js
+++ b/src/BackgroundCells.js
@@ -4,7 +4,7 @@ import clsx from 'clsx'
 
 import { notify } from './utils/helpers'
 import { dateCellSelection, getSlotAtX, pointInBox } from './utils/selection'
-import Selection, { getBoundsForNode, isEvent } from './Selection'
+import Selection, { getBoundsForNode, isEvent, isShowMore } from './Selection'
 
 class BackgroundCells extends React.Component {
   constructor(props, context) {
@@ -77,7 +77,7 @@ class BackgroundCells extends React.Component {
     }))
 
     let selectorClicksHandler = (point, actionType) => {
-      if (!isEvent(node, point)) {
+      if (!isEvent(node, point) && !isShowMore(node, point)) {
         let rowBox = getBoundsForNode(node)
         let { range, rtl } = this.props
 
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -604,6 +604,15 @@ class Calendar extends React.Component {
     showMultiDayTimes: PropTypes.bool,
 
     /**
+     * Determines a maximum amount of rows of events to display in the all day
+     * section for Week and Day views, will display `showMore` button if
+     * events excede this number.
+     *
+     * Defaults to `Infinity`
+     */
+    allDayMaxRows: PropTypes.number,
+
+    /**
      * Constrains the minimum _time_ of the Day and Week views.
      */
     min: PropTypes.instanceOf(Date),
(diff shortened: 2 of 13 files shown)


cutterbl:
@fornesarturo First pass looks like it might be good, but we would need a story in storybook for the new prop, and showcasing how it works.

fornesarturo:
@cutterbl added story for the new prop and updated the API mdx to include it as well

cutterbl:
@fornesarturo Thanks. I'll try to load this up and take a look over the weekend maybe.

fornesarturo:
Anything missing to have this merged?

cutterbl:
@fornesarturo A deeper look at the failing 'show more' functionality shows that the `onShowMore` prop isn't getting passed in to the `TimeGrid` in the `Day`, `Week` or `WorkWeek` components, so that prop isn't available to call. In our next version we'll handle these things with context, but for now these props are manually passed down the component tree.

fornesarturo:
@cutterbl I see that first part as the usual behavior for the `maxRows` prop in the `DateContentRow` component and the `getSlotMetrics` function where `maxRows` includes the show more button row, I could make it so when passing the prop down to `DateContentRow` it is handled as `allDayMaxRows - 1`, maybe change the name of the prop to `allDayMaxEvents` to make it clear. For null I can add the if else with `Infinity`, to handle null as undefined.

On an implementation using react-big-calendar with these changes I am getting a call to the onShowMore function, since I'm working off of that call it worked for my case. I see that what the Month view currently does is it defines the overlay there, it's not something central, I could add the functionality following Month view's approach, per the commits that seems like an intentional exception:

```
handleShowMore = (events, date, cell, slot, target) => {
    const {
      popup,
      onDrillDown,
      onShowMore,
      getDrilldownView,
      doShowMoreDrillDown,
    } = this.props
    //cancel any pending selections so only the event click goes through.
    this.clearSelection()

    if (popup) {
      let position = getPosition(cell, this.containerRef.current)

      this.setState({
        overlay: { date, events, position, target },
      })
    } else if (doShowMoreDrillDown) {
      notify(onDrillDown, [date, getDrilldownView(date) || views.DAY])
    }

    notify(onShowMore, [events, date, slot])
  }
```

I see the differences in notifying the onShowMore prop here where it sends the parameters as an array, so that I can change quickly.

Let me know your thoughts.

fornesarturo:
Result of adding the Month view code to TimeGrid:

image

fornesarturo:
@cutterbl  Added changes to handle overlay and update the `onShowMore` prop `notify` call. Waiting on any comment to change handling of the limited rows in `allDayMaxRows` (subtract one before sending to `maxRows` to not include the Show more button row in the count) or keep as is (include Show more button row in the count).

cutterbl:
@fornesarturo I loaded your changes and brought up storybook. Still not getting the popup with the `popup` prop in the `allDayMaxRows` story...

cutterbl:
@fornesarturo Still not getting the popup, but I do see multiple console warnings
> Warning: Failed prop type: The prop `min` is marked as required in `Week`, but its value is `undefined`.
> Warning: Failed prop type: The prop `max` is marked as required in `Week`, but its value is `undefined`.
> Warning: Failed prop type: The prop `scrollToTime` is marked as required in `Week`, but its value is `undefined`.

fornesarturo:
@cutterbl Hi, just updated to correctly displayed the popup, and fixed propTypes to make sense with settings across the repo.

fornesarturo:
@cutterbl row count is fixed now and displays the number of rows for events specified.

PR #2405, opened 2023-05-22
2405 (Copy of 1649) multiple resources

### LINKED ISSUE #1649: Multiple resources per event
Hi

This is an awesome component, thanks! I'd like to be able to assign multiple resources to a single event and then render this in the resource view.

In my events object, I have an array of ids for each resource linked. This seems to prevent the event from rendering. If I change this to a single ID string it works fine.

Has anyone managed to make this work?

Cheers
Alex

basstager:
Closes #1649 . This PR is a continuation of https://github.com/jquense/react-big-calendar/pull/2165.


stories/demos/exampleCode/resource.js
cutterbl, line 1: You forgot to import `LinkTo` in this file, causing the story to error. Rendering also is missing some spacing.

stories/demos/exampleCode/dndresource.js
cutterbl, line 1: When using the 'copy event' feature, of this documentation, the user would expect that when they drag an event from one resource to another that it would then 'copy' that event, displaying it in both resource columns. This isn't currently the case. It does move the event, but it isn't showing on both columns.

cutterbl (changes requested):
If you can fix the few things noted, and update the branch with latest code from master, I think this will be a nice addition to Big Calendar

basstager:
@cutterbl Hi, thanks for the review. I updated it based on your comments and it should be up to date with master already.

PR #2406, opened 2023-05-31
Add custom timezone support to moment localizer

ariel22411:
Add a custom timezone to moment localizer

https://github.com/jquense/react-big-calendar/issues/2400

### DIFF
--- a/src/localizers/moment.js
+++ b/src/localizers/moment.js
@@ -13,7 +13,7 @@ const timeRangeFormat = ({ start, end }, culture, local) =>
   local.format(start, 'LT', culture) + ' – ' + local.format(end, 'LT', culture)
 
 const timeRangeStartFormat = ({ start }, culture, local) =>
-    local.format(start, 'LT', culture) + ' – '
+  local.format(start, 'LT', culture) + ' – '
 
 const timeRangeEndFormat = ({ end }, culture, local) =>
   ' – ' + local.format(end, 'LT', culture)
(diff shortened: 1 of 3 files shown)


cutterbl (changes requested):
I like the track you're on, but unfortunately this will not work. It assumes that all users of Big Calendar, using `moment`, are also using `moment-timezone`. Many implementations only use `moment` without timezone support, so initializing a date with `moment.tz()` will fail for them, as they don't have the necessary dependencies to extend `moment`.

A better option might be to
- If `timezone` is passed as an argument, then
- test if `moment.tz` is a thing, and
- if not, log a console error, but gracefully ignore, and use `moment()` on it's own

ariel22411:
I changed according to your review

cutterbl:
@ariel22411 One final request here is to add an example to the documentation, in using the `momentLocalizer` in this per instance timezone matter, with an included brief explanation on possible use case for doing so.

ariel22411:
sorry for the delay, now its okay?

tinsaeopus1io:
We've been using this for months, and it has been working great. But now we are having issues on the calendar week view for dates on the week of dst, where the time gutter is off by 1 hour. Please let us know if anyone experienced this issue, and has a fix or workaround @samopus1io @canopus1io

canopus1io:
here are a couple code sandbox examples:
[America/Los_Angeles DST example]( https://codesandbox.io/s/react-big-calendar-demo-2023-forked-4rg2dv?file=/index.js)
[(Europe/London DST example]( https://codesandbox.io/s/react-big-calendar-demo-2023-europe-london-nj3rw3?file=/index.js:4191-4206)

there are two localizer versions to compare here (moment-tz is the localizer in this PR, moment-tz modified is with one tweak that seems to fix the gutter time labels)

here is the tweak diff for the modified version:
image

canopus1io:
@ariel22411 thoughts on this change? (only seems to be different when DST shift is not included in displayed time range for the week)

ariel22411:
@tinsaeopus1io @canopus1io I have a solution, search and try to change this line 
`const tzName = st?._z?.name ?? moment.tz.guess()`
to
`const tzName = timezone ?? moment.tz.guess()`

ariel22411:
The next problem is the DST slots time: (see if you can help me with that) @cutterbl 
image

PR #2409, opened 2023-06-03
Fix popup position when there's no enough space to its right

zheyujie:
# Problem:
When an event title is too long, the `show more` popup may be positioned off screen with the `left` offset being set to a negative value.
Ideally the event title should be concise but I think it's still worth fixing the underlying issue here.

# When there's enough space to its right:
image

# When there's no enough space to its right:
image

# The fix:
Add another condition check to make sure `left` is never negative.

# How it looks like after the fix when there's no enough space to its right:
image

### DIFF
--- a/src/Popup.js
+++ b/src/Popup.js
@@ -26,7 +26,10 @@ function getPosition({ target, offset, container, box }) {
   const right = left + bWidth
   const { x, y } = offset
   const topOffset = bottom > viewBottom ? top - bHeight - y : top + y + height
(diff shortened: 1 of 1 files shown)


zheyujie:
Hey @cutterbl , any chance you can review this PR？
Thanks!

zheyujie:
Hi @cutterbl , are you able to take a look this PR when you have time? Cheers!

MessyComposer:
This seems to be an issue even to this day, would be nice to get this merged.

Issue #2310, opened 2022-12-08
Dragging an event in the RTL month view calendar gets confused to the wrong side


eeefg650:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

https://drive.google.com/file/d/1kKL80YCJLrz6q0Y1GJeWzlwNgkccJ-OA/view?usp=share_link

### Bug description

Calendar Drag and Drop month view always drags event from most left column.

It is important to note that in the "week" & "day" view it works properly.

### Expected Behavior

While dragging, I expect the event to hover over the date which I am selecting.

### react-big-calendar version

 1.5.2

### React version

18.2.0

### Platform/Target and Browser Versions

chrome 107.0.5304.107 microsoft edge

### Validations

- [X] Read the [docs](http://jquense.github.io/react-big-calendar/examples).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a react-big-calendar issue and not an implementation issue

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

butermann:
Can you provide a reproduction link, your code or a video?

eeefg650:
> Can you provide a reproduction link, your code or a video?

Yes, hope that will be helpful:
https://www.veed.io/view/79e6c327-f18f-402c-bb22-874473027393?sharingWidget=true&panel=share

eeefg650:
> Can you provide a reproduction link, your code or a video?

I would greatly appreciate your help, I tried every option but unfortunately without any success.

butermann:
That's probably a bug in the `dragAndDrop` add-on. It's probably just that no-one ever used it that way.
Maybe you can try to debug it. It should be somewhere near here: https://github.com/jquense/react-big-calendar/blob/master/src/addons/dragAndDrop/WeekWrapper.js#L62

eeefg650:
> That's probably a bug in the `dragAndDrop` add-on. It's probably just that no-one ever used it that way. Maybe you can try to debug it. It should be somewhere near here: https://github.com/jquense/react-big-calendar/blob/master/src/addons/dragAndDrop/WeekWrapper.js#L62

Thanks for you response!
how do i get access to that all events?

bu6n:
Don't understand your question. If you want to try to fix the bug, you could clone the repo locally, modify an existing Storybook story and then try to debug.

eeefg650:
[...]

Issue #1801, opened 2020-11-07
Dragging an event in the RTL month view calendar,


JonathanGilyadov:
#### Do you want to request a _feature_ or report a _bug_?
A bug
#### What's the current behavior?
When dragging an event in the RTL  month view calendar, the calendar gets confused and the event goes in the opposite direction

My browser is google chrome,
React big calendar version 0.28.5

I found this, which fully describes my issue, but for me, it keeps happening
https://github.com/jquense/react-big-calendar/pull/172

#### What's the expected behavior?

While dragging, I expect the event to hover over the date which I am selecting

spirituelle:
I released the same issue, I made a 2 days research to control it manually, but no solutions unfortunately, so I removed the drag and drop for the moment.

JonathanGilyadov:
> I released the same issue, I made a 2 days research to control it manually, but no solutions unfortunately, so I removed the drag and drop for the moment.

For me, the drag and drop is crucial, so I made the calendar LTR while my whole website is RTL, this so very unfortunate.

PR #2426, opened 2023-08-07
Fix bug - Dragging an event in the RTL month view calendar gets confused to the wrong side

amirbashan:
Fix issue #2310 and #1801

**Issue:** 
https://drive.google.com/file/d/1kKL80YCJLrz6q0Y1GJeWzlwNgkccJ-OA/view
When dragging an event in the RTL month view calendar, the calendar gets confused and the event goes in the opposite direction, this also happens in the week view at the allDay section.

**Solution:**
I have forwarded the `rtl` params to the `WeekWrapper`.
This is a fix specifically for dragAndDrop and doesn't solve all of the RTL problems.

### DIFF
--- a/src/DateContentRow.js
+++ b/src/DateContentRow.js
@@ -181,7 +181,7 @@ class DateContentRow extends React.Component {
             </div>
           )}
           <ScrollableWeekComponent>
-            <WeekWrapper isAllDay={isAllDay} {...eventRowProps}>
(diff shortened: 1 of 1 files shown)


cutterbl:
@amirbashan Thanks for this. I will take a look. You mentioned that this does not fix all of the RTL problems. We are in the process of planning the next major version of Big Calendar. If you could provide your feedback in the `#next` channel of our Slack community, it may be helpful in resolving other issues.

PR #2459, opened 2023-10-26
Fixing invalid ref with invalid scrollHeight

eduhdev12:
Hello, the issue is related with https://github.com/jquense/react-big-calendar/issues/725 and https://github.com/jquense/react-big-calendar/pull/734

The error is triggered if the calendar gets suspended by a parent and when the component should be re-rendered it's throwing the error because the content ref is not valid at that time.

image


cutterbl (changes requested):
Think you could get away with just `!content?.scrollHeight` here, and eliminate the `||` all together

PR #2460, opened 2023-10-28
fix(utils): correct a react lifecylce method in DayColumn.js component

MichaelKleyman:

In production and development, the top navigation buttons like "Today", "Next", "Back" and the "Month" "week" "day" "agenda" do not operate due to the following error below:

Error: Warning: Using UNSAFE_componentWillReceiveProps in strict mode is not recommended and may indicate bugs in your code. See https://reactjs.org/link/unsafe-component-lifecycles for details. * Move data fetching code or side effects to componentDidUpdate. * If you're updating state whenever props change, refactor your code to use memoization techniques or move it to static getDerivedStateFromProps. Learn more at: https://reactjs.org/link/derived-state

With some reluctant research the only alternative I sought after was to change the lifecycle method UNSAFE_componentWillReceiveProps which is deprecated, to the getDerivedStateFromProps method and see from there if the problem is fixed.


cutterbl:
What you describe above is not an error, but a JS warning. This is related to React's future deprecation of that particular class based life cycle method. I have seen your issue, when using uncontrolled 'date' and 'view' props in an SSR environment.

MichaelKleyman:
What do you suggest is a proper way to navigate the issue of the top navigation buttons not working on the calendar UI when it is living on a nextjs client component so its being rendered during SSR. I'm curious also if you have this problem as well in development or production.

cutterbl:
I know that using controlled setup (controlled 'date' and 'view') helps. Having never personally used nextjs, I really don't know what creates the issue, and don't know of anyone actively working a PR that would truly address the problem. We are discussing the next version of Big Calendar in our Slack channel, which will be a major refactor, but it's all still in it's infancy.

MichaelKleyman:
So I have two followup questions to your comment:
- Do you think the JS warning i'm getting contributing to my problem of not being able to have function on those specified buttons I mentioned earlier? It doesn't appear on other example uses of RBC that I viewed.
- Would I able able to join the slack channel?

cutterbl:
Anyone can join the Slack channel. The link is on the README. The warnings are not the cause of the problem (they are warnings, not errors). Moving to controlled vs uncontrolled has seemed to work for others using SSR, but I can't verify that.

MichaelKleyman:
Slack channel link doesn't work for some reason, do you have an invite link you can send over? I'd want to keep up to date on anything going on with the library since my application relies on calendar functionality.

MichaelKleyman:
Update, i ended up fixing the issue on my end, nothing wrong with the RBC code so this merge is unnecessary. For future nextjs13 and higher version users, simply going into the next.config.js and setting the config as shown below fixes the issue because by default strictmode is set to true, and that was causing my issue

const nextConfig = {
  reactStrictMode: false,
};

PR #2470, opened 2023-11-20
Fix UNSAFE_componentWillReceiveProps in strict mode is not recommended

justoverclockl:
This fix should address the warning for:

```
UNSAFE_componentWillReceiveProps in strict mode is not recommended and may indicate bugs in your code. See https://reactjs.org/link/unsafe-component-lifecycles for details.

* Move data fetching code or side effects to componentDidUpdate.
* If you're updating state whenev
```

### DIFF
--- a/src/DayColumn.js
+++ b/src/DayColumn.js
@@ -38,35 +38,33 @@ class DayColumn extends React.Component {
     this.clearTimeIndicatorInterval()
   }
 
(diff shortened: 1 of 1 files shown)


Ahiaz:
Hello, how do I apply these changes to the library, to correct this error, since 18.6 continues with the problem?

hyoloui (approved):
LGTM!!

PR #2472, opened 2023-11-23
feat: time gutter position features

bzbetty:
Allows the time gutter to be on the Left (default), Right, or Both

image

image

### DIFF
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -690,6 +690,11 @@ class Calendar extends React.Component {
       timeGutterFormat: dateFormat,
 
       /**
+       * The position the time gutter gets rendered (left, right, both)
+       */
+      timeGutterPosition: PropTypes.oneOf(['left', 'right', 'both']),
+
+      /**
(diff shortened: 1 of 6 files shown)


cutterbl:
This looks really good. Only thing I see missing is a story under `/stories/props`.

bzbetty:
Added in a story under /stories/props for TimeGutterPosition

cutterbl:
OK, finally had a little more time to look at this. First, Thank You for your PR, and adding the base level documentation, though you did miss adding the prop/link to the [Full Prop List](https://jquense.github.io/react-big-calendar/examples/index.html?path=/docs/props-full-prop-list--page).

A more detailed review shows one item that, while it doesn't appear to be currently causing an issue, may create one. The TimeGutter gets a ref. Currently you're assigning the same ref even when you have two gutters (left and right). This is used for measurement layout in the `measureGutter()` method. Might be nothing to worry about, but does set off some bells.

PR #2475, opened 2023-11-26
Fix dragAndDrop EventWrapper.js error: cannot add property 'X', object is not extensible

moonbeam-dev:
Fix for error: cannot add property 'X', object is not extensible

Above console error appears when using the Resource Drag and Drop feature of react big calendar to move events.  I narrowed the issue down to this code in the addons/dragandDrop/EventWrapper.js:
let extendedEvent = this.props.event

Per the [linked discussion on Stack Overflow](https://stackoverflow.com/questions/55567386/react-cannot-add-property-x-object-is-not-extensible), it is not possible to modify this.props directly. The solution is to create a copy of the props using JSON.parse() and JSON.stringify()

Thus the modified code that I have gotten working without error is:
let extendedEvent = JSON.parse(JSON.stringify(this.props.event))

### DIFF
--- a/src/addons/dragAndDrop/EventWrapper.js
+++ b/src/addons/dragAndDrop/EventWrapper.js
@@ -47,7 +47,7 @@ class EventWrapper extends React.Component {
       .getAttribute('class')
       ?.includes('rbc-addons-dnd-resize')
     if (!isResizeHandle) {
-      let extendedEvent = this.props.event
+      let extendedEvent = JSON.parse(JSON.stringify(this.props.event))
       extendedEvent.sourceResource = this.props.resource
       this.context.draggable.onBeginAction(this.props.event, 'move')
     }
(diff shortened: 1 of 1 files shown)


cutterbl:
The issue I see here is in the handling of dates. Event `start` and `end` dates must currently be true JS Date objects. By using the methodology to 'copy' an event, these values would be converted to strings. Perhaps a better way of handling this would be:
`let extendedEvent = {....this.props.event};`

moonbeam-dev:
Thank you so much Steve @cutterbl !  I tested your solution and can confirm it is working well, minus the one extra dot :wink:
`let extendedEvent = {...this.props.event};`

When I edited my previous PR to match your solution, it created a new commit. Still learning the ropes here, so not sure if this is the correct process.

Thank you again and let me know if you need me to make any changes for the approval.

Issue #2488, opened 2024-01-16
Screen Readers should read month & year


tylerlinquata:
### Clear and concise description of the problem

When using a Screen Reader, the user will never hear the month or year that are in the toolbar. When tabbing though date cells, only the day is read. This has been flagged as an accessibility issue by my company.

Current experience using VoiceOver:

https://github.com/jquense/react-big-calendar/assets/25713879/92689545-99da-452a-bda8-e8b71fc8bade

### Suggested solution

In `DateHeader.js`, we can add a label that allows the screen reader to read the full localized date. This would be visually hidden. 

Example js:
```
<button
  type="button"
  className="rbc-button-link"
  onClick={onDrillDown}
  role="cell"
>
  <span aria-hidden="true">{label}</span>
  <span className="rbc-sr-only">
    {localizer.format(date, 'ddd MMM DD YYYY')}
  </span>
</button>
```

Example css:
```
.rbc-sr-only {
  position: absolute;
  width: 1px;
  height: 1px;
  padding: 0;
  margin: -1px;
  overflow: hidden;
  clip: rect(0, 0, 0, 0);
  border: 0;
}
```
Example video:

https://github.com/jquense/react-big-calendar/assets/25713879/96adf020-0e43-44c0-837d-e030b306acf8

### Alternative

One alternative would be to change the `Toolbar` such that it is read by the screen reader. This may be considered compliant from an accessibility standpoint, but it doesn't feel like a great user experience.

### Additional context

I'm happy to contribute this fix if allowed.

### Validations

- [X] Read the [docs](http://jquense.github.io/react-big-calendar/examples/index.html).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Would you like to open a PR for this feature?

- [X] I'm willing to open a PR

cutterbl:
From an accessibility standpoint though, is that really the right way to do it? We do pass the `date` into the `DateHeader` component. Perhaps if we wrapped the `label` in an html `<time>` tag with a formatted `datetime` attribute?

Accessibility is a large problem in Big Calendar, and improvements are on our radar for the 'next' version. We'd be happy to look at any PR you might submit for this particular issue, and welcome your feedback in our Slack community's 'next' channel on how to better approach accessibility going forward.

tylerlinquata:
[...]

PR #2493, opened 2024-01-30
feat: add screen reader support to date headers

tylerlinquata:
Adds screen reader support to date headers, as discussed in #2488.

### DIFF
--- a/src/DateHeader.js
+++ b/src/DateHeader.js
@@ -1,18 +1,18 @@
 import PropTypes from 'prop-types'
 import React from 'react'
 
-const DateHeader = ({ label, localizer, date, drilldownView, onDrillDown }) => {
-  const renderLabel = () => (
-    <>
-      <span aria-hidden="true">{label}</span>
-      <span className="rbc-sr-only">
-        {localizer.format(date, 'ddd MMM DD YYYY')}
-      </span>
-    </>
-  )
+const DateHeaderLabel = (label, localizer, date) => (
+  <>
+    <span aria-hidden="true">{label}</span>
+    <span className="rbc-sr-only">
+      {localizer.format(date, 'ddd MMM DD YYYY')}
+    </span>
+  </>
+)
 
+const DateHeader = ({ label, localizer, date, drilldownView, onDrillDown }) => {
   if (!drilldownView) {
-    return renderLabel()
+    return <DateHeaderLabel label={label} date={date} localizer={localizer} />
   }
 
   return (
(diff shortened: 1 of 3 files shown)


src/DateHeader.js
```diff
@@ -1,9 +1,18 @@
 import PropTypes from 'prop-types'
 import React from 'react'
 
-const DateHeader = ({ label, drilldownView, onDrillDown }) => {
+const DateHeader = ({ label, localizer, date, drilldownView, onDrillDown }) => {
+  const renderLabel = () => (
```
cutterbl, line 5: Doing it this way recreates this function with each render of the DateHeader. Create a separate HeaderLabel component function above the DateHeader (you don't need to export it), passing it the `label`, `date` and `localizer`, and use the component within the DateHeader.

cutterbl (changes requested):
Looks good, if you address the one item in the DateHeader

src/DateHeader.js
tylerlinquata: Updated to move the label into a separate component.

src/DateHeader.js
```diff
@@ -1,18 +1,18 @@
...
-const DateHeader = ({ label, localizer, date, drilldownView, onDrillDown }) => {
-  const renderLabel = () => (
-    <>
-      <span aria-hidden="true">{label}</span>
-      <span className="rbc-sr-only">
-        {localizer.format(date, 'ddd MMM DD YYYY')}
-      </span>
-    </>
-  )
+const DateHeaderLabel = (label, localizer, date) => (
```
cutterbl, line 4: Not a true function component signature. The arguments are props, and should be destructured from the props object (see line 13 as an example).

When contributing changes like these, make sure that your running Storybook locally (`npm run storybook`). This will allow you to see your changes real time, and avoid errors.

Corrections, as described above, will apply to the 'day' display (the number in each day cell) of the Month view of the Calendar. For column headers (Month or TimeGrid views) you would update the Header component. In this case, it may be best to separate your DateHeaderLabel into it's own separate component, and import it into each file.

Further reflection on your 'format' shows that, while appropriate for the 'moment' localizer, it would break if someone was using the 'Luxon', 'dayjs' or the 'date-fns' localizers, all of which have different format tokens. I would suggest adding a new format type to each (`screenReaderDateFormat`?) with the appropriate format, and then reference that in your new `DateHeaderLabel` component `localizer.format(date, 'screenReaderDateFormat')`.

cutterbl (changes requested):

PR #2495, opened 2024-02-01
fix(dnd): move merge components to render to be able to receive state…

iwollmann:
Fix for bug: https://github.com/jquense/react-big-calendar/issues/2359

The components are being created in the constructor and overwrites in render with this it's not receives any state updates from any custom component.


leonardo-bittencourt-atria:
👍 I've been facing the same issue here! Thanks for your PR, Iuri!

leonardo-bittencourt-atria:
@jquense Could you please take a look here?

PR #2496, opened 2024-02-02
feat: showMore message add event info

zht826:
image
Sometimes, we need to incorporate additional event information into texts that display more items. Therefore, we have added the input of event parameters.


cutterbl:
Can you give me a practical use case for this?

zht826:
such as this
image
I want to show the title of the remaining items

cutterbl:
Your code changes look good, but you should update the [messages](https://jquense.github.io/react-big-calendar/examples/index.html?path=/docs/props--messages) documentation prior to approval

zht826:
I improved the parameters and added the remaining events and all-day event parameters, which correspond to the previous usage scenario.
I have updated the messages doc.

Issue #2498, opened 2024-02-09
[Accessibility] "Ensure elements with ARIA role that require child roles contain them"


vasanthkumar258:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

https://lmwkj3.csb.app/

### Bug description

Testing the app with Accessibility Insights for Web resulted in errors. Observed same with Playwright-axe as well.
![image](https://github.com/jquense/react-big-calendar/assets/12809041/0053e6b1-db61-4fdf-abb3-f33fe4b3ff35)

### Expected Behavior

No accessibility issues should be reported

### Actual Behavior

Accessibility issues are observed

### react-big-calendar version

1.9.1

### React version

18.2.0

### Platform/Target and Browser Versions

Edge - 121.0.2277.112

### Validations

- [X] Read the [docs](http://jquense.github.io/react-big-calendar/examples).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a react-big-calendar issue and not an implementation issue

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

vasanthkumar258:
![image](https://github.com/jquense/react-big-calendar/assets/12809041/492c3e0b-fb48-4372-aba3-c0e4bb3f1d75)

IN the below image, under role="row", we have divs without any role but have tabindex. usually role="row" expects role"cell" as child. Also I hope we dont need any tabindex there.. if needed, please add any appropriate role for that..
![image](https://github.com/jquense/react-big-calendar/assets/12809041/bad40c8e-74bf-4125-b9d4-f4f6e57805b4)

PR #2508, opened 2024-02-14
fix(ARIA): remove tabindex

cutterbl:
The tabindex, applied to the event div, creates invalid ARIA component nesting

#2498


genehallman:
I've noticed that this PR breaks the key press event on the month view.

PR #2510, opened 2024-02-15
fix overlapping background events, fixes #2452

### LINKED ISSUE #2452: Multiple background events are overlapping with next day

- [X] I confirm

### Reproduction link

https://codesandbox.io/s/react-big-calendar-example-forked-r7djnd?file=/src/App.js

### Bug description

When calendar has multiple background events for the same period then second and next events displays in next day  
  
image

### Expected Behavior

Maybe to trim event (like usual event) will be acceptable in this case.  
Smth like following:  

image
  

1.8.1

### React version

18.2.0

Any browser/platform

### Validations

chazzlabs:
This PR addresses two issues with how background events are rendered using both the `overlap` and `no-overlap` layout algorithms. I opted for the change I made due to its simplicity and the unlikelihood that it would have any adverse effects elsewhere in the calendar. It appears to me that the implementation of the `no-overlap` layout algorithm very deliberately returns different types for the `width` and `xOffset` style values than does the `overlap` algorithm; given this it seemed most reasonable to correct how those values are handled when rendering events in `TimeGridEvent`. 

I added two stories under "Additional Examples > Layout" to demonstrate the `overlap`/`no-overlap` behavior since that seemed like the most appropriate location for them.

### DIFF
--- a/src/TimeGridEvent.js
+++ b/src/TimeGridEvent.js
@@ -47,7 +47,7 @@ function TimeGridEvent(props) {
         top: stringifyPercent(top),
         height: stringifyPercent(height),
         // Adding 10px to take events container right margin into account
-        width: `calc(${width} + 10px)`,
+        width: `calc(${stringifyPercent(width)} + 10px)`,
         [rtl ? 'right' : 'left']: stringifyPercent(Math.max(0, xOffset)),
       }
     : {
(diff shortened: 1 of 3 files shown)


chazzlabs:
Thanks for the feedback! I'll go ahead and make the necessary updates to have `backgroundEvents` behave similarly to `events` with respect to their size and position. I agree that the other enhancements seem more appropriate for a future version given the potential breaking changes for some folks.

chazzlabs:
I'm not sure of the rationale behind why `backgroundEvents` were styled differently than `events` with regard to layout, but my latest commit makes the layout styling consistent for both kinds of events which seems like what we wanted to achieve.

cutterbl:
@chazzlabs Looks great. Nice work.

PR #2528, opened 2024-03-04
fixes #2424 - fix js error TypeError: Cannot read properties of undef…

### LINKED ISSUE #2424: When selectable is true, click into calendar and press any key on keyboard at same time, this error appears.

- [X] I confirm

### Reproduction link

https://codesandbox.io/s/gracious-panini-5jwfwq?file=/src/App.js

### Bug description

When selectable is true, click into calendar and press any key on keyboard at same time, this error appears. 
Screenshot 2023-07-22 171140

### Expected Behavior

_No response_

### Actual Behavior

_No response_

1.8.1

### React version

18.2.0

Chrome, Firefox, etc...

### Validations

LorienHW:
…ined (reading 'getDate')

see issue 2424 

you can repro the issue by clicking around a selectable month view of the calendar a lot and it helps to press keys at the same time (as is mentioned in the issue)

i think whats happening is that `handleSelectSlot` sets the `_pendingSelection` but then calls `selectDates` via setTimeout, and before `selectDates` executes something else calls `clearSelection` which empties out the `_pendingSelection` array. I've tested this fix and it seems to work - exiting early results in the slot not appearing selected and the onSelectSlot callback not being called

### DIFF
--- a/src/Month.js
+++ b/src/Month.js
@@ -336,6 +336,8 @@ class MonthView extends React.Component {
   }
 
   selectDates(slotInfo) {
+    if (!this._pendingSelection?.length) return;
+    
     let slots = this._pendingSelection.slice()
 
     this._pendingSelection = []


forceddd:
@LorienHW Hi,this bug seems to be because the `keydown` event triggers the `click` behavior as a `select` behavior. I think this should be a logical bug with the Selector.What do you think？

LorienHW:
the `handleSelectSlot` function sets the `_pendingSelection` and then uses a `setTimeout` to call `selectDates`. Because its not called synchronously there is an opportunity for `clearSelection` to be called by something else (for example a key press), and empties out the `_pendingSelection` array (which is what i see happening in testing it). So just based on this sequence of code, since it is possible to get into `selectDates` with an empty `_pendingSelection` I think it should verify that there is something in there rather than assuming it has at least one item (which is what the change in this PR does)

forceddd:
Yes, you have a point. But this modification only silences this error in Month View. You will find another bug when you first select an area and then hold down the left mouse button while pressing the keyboard again.This video shows this performance

https://private-user-images.githubusercontent.com/63057542/318682798-b1f9a1e7-c26d-4abc-be77-7ccb8ce759ca.mov?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTIxMTA3MDgsIm5iZiI6MTcxMjExMDQwOCwicGF0aCI6Ii82MzA1NzU0Mi8zMTg2ODI3OTgtYjFmOWExZTctYzI2ZC00YWJjLWJlNzctN2NjYjhjZTc1OWNhLm1vdj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA0MDMlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwNDAzVDAyMTMyOFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWRhZjAyYmY0ZTNlNzFmYzBkODhkZGQ2NTc5ZGRlYTRjOTA4YmIyNzU4OTUwNjA5OGQ1YzBlZmIzYTE4OTg0YzImWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.i3GAkmZM7oKjl9VJOT6BaMlnWFshDdL4VLuKOy7vS2A

PR #2537, opened 2024-03-13
feat(components): showMore component customization

### LINKED ISSUE #2391: Support for mounting a custom "Show More" popup at a particular cell

I am using big calendar in my application, and I need to show a popup which appears on clicking the "Show More" button. But, this popup is very different from the one currently rendered by big calendar, hence I need to implement my custom popup instead of using the built-in one.

The issue I am facing is that the current contract of the calendar component has this `onShowMore` callback and this `messages` configuration function, both of these do not provide enough information for me to **mount my popup at the cell in which the "Show More" was clicked**.

1. The issue with `onShowMore`:  `onShowMore` provided me with the `events` and the `date` of the cell in which the "Show More" button was clicked. But it doesn't provide me with the `target` HTML element of the "Show More" button. Hence, I am not able to mount my popup at that specific cell.

image

2. The issue with `messages`: The `showMore` property on `messages` allows me to render custom JSX (if I ignore the string output type). Here I could technically render my own button, and manage my own click handlers. This would provide me with my target element for mounting my popup. But, here the issue is that I don't have access to the `events` and `date` of that particular cell, hence I won't be able to show the relevant information in my popup.

image

Neither of the above two props provide me with enough information to render my popup while mounting it at the cell in which "Show More" was clicked.

### Suggested solution

I suggest that we add support for `target` argument as well in the `onShowMore` callback. The updated function would look like this:

```
function (events: Array[event: Object], date: Date, target: HTMLElement) => any
```

I could then just use the `events` and `date` information to render my custom popup, while being mounted at `target` HTML element.

### Alternative

_No response_

### Additional context

I have read this issue - https://github.com/jquense/react-big-calendar/issues/1147. But it doesn't solve my problem. The fix which was raised for this issue only allows me to render my own popup, but does not allow me to mount it at a specific cell.

### Validations

davidmh:
I'm trying to render a custom component for the `show more` button. I know there are a few recommended ways to do this through the `onShowMore` callback and the `messages.showMore` prop as suggested in #1147.

But that doesn't cover all use cases. For example, I want to access the events that are being hidden by the `show more` button. I also want to render a custom popup right next to the `show more` button.

This new component adds the following props:

- localizer
- slot
- slotDate
- count
- events
- remainingEvents

I can also update the types once this gets merged.

Resolves #2391

---

I tried reaching out through the Slack community first, but the invitation link has expired, could you post a new link?

### DIFF
--- a/src/EventEndingRow.js
+++ b/src/EventEndingRow.js
@@ -75,10 +75,30 @@ class EventEndingRow extends React.Component {
   }
 
   renderShowMore(segments, slot) {
-    let { localizer, slotMetrics } = this.props
+    let { localizer, slotMetrics, components } = this.props
     const events = slotMetrics.getEventsForSlot(slot)
     const remainingEvents = eventsInSlot(segments, slot)
     const count = remainingEvents.length
+
+    if (components?.showMore) {
+      const ShowMore = components.showMore
+      // The received slot seems to be 1-based, but the range we use to pull the date is 0-based
+      const slotDate = slotMetrics.getDateForSlot(slot - 1)
+
+      return count ? (
+        <ShowMore
+          localizer={localizer}
+          slotDate={slotDate}
+          slot={slot}
+          count={count}
+          events={events}
+          remainingEvents={remainingEvents}
+        />
+      ) : (
+        false
+      )
+    }
+
     return count ? (
       <button
         type="button"
--- a/stories/resources/customComponents.js
+++ b/stories/resources/customComponents.js
@@ -81,6 +81,27 @@ const customComponents = {
       </div>
     )
   },
+  showMore: (showMoreProps) => {
+    return (
+      <button
+        id="my-custom-show-more"
+        style={{ border: '4px solid red', cursor: 'pointer' }}
+        onClick={() => {
+          console.log('showMoreProps', showMoreProps)
+          window.alert(`
+            Clicked ${showMoreProps.slotDate
+              .toISOString()
+              .substr(0, 10)} with ${
+            showMoreProps.remainingEvents.length
+          } remaining events.
+            Open the console for the full set of props.
+          `)
+        }}
+      >
+        {showMoreProps.count} more!
+      </button>
+    )
+  },
 }
 
 export default customComponents
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -750,6 +750,7 @@ class Calendar extends React.Component {
      *   timeGutterHeader: MyTimeGutterWrapper,
      *   timeGutterWrapper: MyTimeGutterWrapper,
      *   resourceHeader: MyResourceHeader,
+     *   showMore: MyShowMoreEvent,
      *   toolbar: MyToolbar,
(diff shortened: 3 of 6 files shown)


src/EventEndingRow.js
```diff
@@ -75,10 +75,28 @@ class EventEndingRow extends React.Component {
...
     const remainingEvents = eventsInSlot(segments, slot)
     const count = remainingEvents.length
+
+    if (components && components.showMore) {
+      const ShowMore = components.showMore
+
+      return count ? (
+        <ShowMore
+          localizer={localizer}
+          slotDate={slotMetrics.getDateForSlot(slot - 1)}
```
davidmh, line 89: The `getDateForSlot` access the slots range, and array, using the slot as the index. But the slot received in this method are not zero-based, which means, we need to adjust for the expected zero-based index used to pull from the rage: https://github.com/davidmh/react-big-calendar/blob/770be8b87118a59f7be01f0ea44e9bead14b59fa/src/utils/DateSlotMetrics.js#L38-L40

FerShihab:
is there any update for this feature?

davidmh:
@FerShihab still waiting for a review. @cutterbl could you take a look?

If it helps, I've been using it through a fork: https://www.npmjs.com/package/@davidmh/react-big-calendar

cutterbl:
@davidmh Just seeing this (busy times). If you could resolve the branch conflicts, then I can take a look

stories/resources/customComponents.js
```diff
@@ -79,6 +79,25 @@ const customComponents = {
...
+        style={{ border: '4px solid red', cursor: 'pointer' }}
+        onClick={() => {
+          action(
+            `custom showMore component clicked with props: ${JSON.stringify(
+              showMoreProps,
+              null,
+              2
+            )}`
+          )
+        }}
```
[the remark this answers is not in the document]
cutterbl: @davidmh Our Storybook implementation is out of date with current Storybook. I just haven't had the time to work the upgrade.

src/EventEndingRow.js
cutterbl: @davidmh Looks like a good catch to me. This code was long before my time on the project, and I can see where it would have been easily missed.

davidmh:
Thanks for the review @cutterbl, I'll address the feedback later this week, I'll try to create a separate PR to fix the issue with `getDateForSlot` and rebase this branch on top of that fix. Just to avoid mixing fixes and features.

src/EventEndingRow.js
davidmh: I started looking into the index issue with the slots, but that ends up breaking a bunch of other features, so I'll just leave it as it is.

I added a few more events in the story to make sure the custom component behaves as expected on the first day of a row.

cutterbl:
@davidmh Thanks for pulling in latest changes. One last bit I see is you need to update the `components` story to include `showMore` override ability.

davidmh:
Done

PR #2545, opened 2024-03-25
fix: issue 2536

forceddd:
This is a commit to fix[ issue 2536](https://github.com/jquense/react-big-calendar/issues/2536) and [2529](https://github.com/jquense/react-big-calendar/issues/2529).

When I was fixing [2529](https://github.com/jquense/react-big-calendar/issues/2529), I declared `slotMetrics` as a global variable, which caused all `DayColumn` instances to use the same instance of `slotMetrics`. This is also the reason why bug [ issue 2536](https://github.com/jquense/react-big-calendar/issues/2536) appears.

In this commit, I only modified the update timing of slotMetrics.Now both issues are resolved!

https://github.com/jquense/react-big-calendar/assets/63057542/756fc1b7-972e-4ef3-bb72-516e068218ac

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## [1.11.1](https://github.com/jquense/react-big-calendar/compare/v1.11.0...v1.11.1) (2024-03-04)
+
+
+### Bug Fixes
+
+* replace deprecated onKeyPress by onKeyDown ([21f51f2](https://github.com/jquense/react-big-calendar/commit/21f51f2bc4e218542fb09bf0e7d22be99ed50028))
+
 # [1.11.0](https://github.com/jquense/react-big-calendar/compare/v1.10.3...v1.11.0) (2024-02-26)
 
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-big-calendar",
-  "version": "1.11.0",
+  "version": "1.11.1",
   "description": "Calendar! with events",
   "author": {
     "name": "Jason Quense",
--- a/src/utils/TimeSlots.js
+++ b/src/utils/TimeSlots.js
@@ -68,12 +68,19 @@ export function getSlotMetrics({
     },
 
     nextSlot(slot) {
-      let next = slots[Math.min(slots.indexOf(slot) + 1, slots.length - 1)]
+      // We cannot guarantee that the slot object must be in slots,
+      // because after each update, a new slots array will be created.
+      let next =
+        slots[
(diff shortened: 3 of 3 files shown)


cutterbl:
Loading your changes in Storybook, if I go to the onSelectSlot story and begin a selection, it immediately changes it to begin at the beginning of the day up to my selection point.

forceddd:
This is an issue that exists in version 1.11.1 and is not caused by this commit.

forceddd:
> Loading your changes in Storybook, if I go to the onSelectSlot story and begin a selection, it immediately changes it to begin at the beginning of the day up to my selection point.

This is an issue that exists in version 1.11.1 and is not caused by this commit.

cutterbl:
@forceddd You are correct, but it was your previous PR that appears to have introduced this issue in the first place.

forceddd:
This bug should have existed a long time ago. [The issue 2290](https://github.com/jquense/react-big-calendar/issues/2290) seems to be the bug described.
I will try to fix it next week, if I successfully solve this issue, I will create another PR.

cutterbl:
@forceddd Thanks for looking into this. An you are partly right. That issue you pointed out was for dragging up. Now it is also doing it even if dragging down (which was not an issue before your change).

forceddd:
@cutterbl After each `slotMetrics` update, a new `slots` array will be created, but the `_initialSlot` object in `DayColumn` remains unchanged, so `nextSlot` becomes 0:00.
I modified `nextSlot` to handle this case. This video shows the performance after the modification.

https://github.com/jquense/react-big-calendar/assets/63057542/49a17560-6e0f-47c6-bfbb-b93e92cc010e

forceddd:
Is the margin-right here intentionally designed this way? Because when I'm selecting, there is no such spacing, but the rendered result has spacing. This makes me a little confused.
image

forceddd:
@cutterbl Would you please check this PR again? I submitted some new file changes.I think now this PR should be able to solve these related problems.

cutterbl:
@forceddd Another defect (#2565 ) has come up around your changes. Not only is the onSelectSlot no longer returning a slotInfo.action of 'doubleClick', but the onDoubleClickEvent handler function no longer fires either.

forceddd:
> @forceddd Another defect (#2565 ) has come up around your changes. Not only is the onSelectSlot no longer returning a slotInfo.action of 'doubleClick', but the onDoubleClickEvent handler function no longer fires either.

@cutterbl Ok, today I will look at this issue and try to fix it.

Issue #2534, opened 2024-03-12
Incorrect Calendar display and missing dates in 'month' view when using dayjsLocalizer - March 2024


ncaraker:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

https://jquense.github.io/react-big-calendar/examples/index.html?path=/docs/examples--dayjs-localizer

### Bug description

When using the `dayjsLocalizer` the `month` view for March 2024 is incorrect. March 10th 2024 is missing entirely and the days of the week display do not align with the dates on the calendar. This can be seen in the storybook examples provided ([here](https://jquense.github.io/react-big-calendar/examples/index.html?path=/docs/examples--dayjs-localizer))

### Expected Behavior

All dates should show on the calendar. The days of the week should start on Sunday and should align with calendar dates as expected

![Screenshot 2024-03-12 at 4 39 13 PM](https://github.com/jquense/react-big-calendar/assets/33259819/e48090f4-5fd6-4329-8f69-d41aeb38b25d)

### Actual Behavior

March 10th 2024 is missing entirely from the calendar, and the days of the week start on incorrect day and do not line up with the corresponding days on the calendar.

![Screenshot 2024-03-12 at 4 41 56 PM](https://github.com/jquense/react-big-calendar/assets/33259819/8170b36d-56ad-4b6c-b276-253e60e354a9)

### react-big-calendar version

1.11.2

### React version

^17.0.2

### Platform/Target and Browser Versions

All platforms

### Validations

- [X] Read the [docs](http://jquense.github.io/react-big-calendar/examples).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a react-big-calendar issue and not an implementation issue

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

sethmcleod:
We're also seeing issues with the dayjs localizer. Not the same as you (March 10 isn't missing) but many of the days in March are incorrectly rendering as out of range. This happens for all the years we've tested and is reproducible in [the Storybook](https://jquense.github.io/react-big-calendar/examples/index.html?path=/docs/examples--dayjs-localizer). Switching to Moment localizer resolves the issue.

<img width="1028" alt="image" src="https://github.com/jquense/react-big-calendar/assets/7340187/5fd711db-2c7b-4cf9-82bd-2415f0ca4961">

prescarlton:
Looks like it's directly related to leap years.

[...]

PR #2566, opened 2024-04-16
fix: issue #2534

prescarlton:
this commit fixes the issues mentioned in #2534 

In cases of leap years, the month of March would often display unexpected off-range days / missing days entirely.
This includes special handling for leapyears, and does manual math to get the start of the week rather than Dayjs' `startOf`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,7 @@ const MyCalendar = (props) => (
 Note that the dayjsLocalizer extends Day.js with the following plugins:
 
 - [IsBetween](https://day.js.org/docs/en/plugin/is-between)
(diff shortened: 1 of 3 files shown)


dhmoon91:
hmm This is over a year old now... Could we get this merged?

prescarlton:
> @prescarlton @eddielee394 @dhmoon91 The reason this has not been merged in is because the fix currently only addresses the issue when using the dayjs localizer. Is the issue present with the moment localizer? Or the luxon localizer? Has it been verified?

It's admittedly been a while since I've looked at this but iirc it only occurred with the dayjs localizer

PR #2573, opened 2024-04-25
Recalculate "more" with ResizeObserver

### LINKED ISSUE #828: Recalculate the need of `more`
react-big-calendar calculates the layout of the events and the need to show the `more` link when the component mounts and when the window resizes, but i am using my component inside a resizeable component and the calendar events are not recalculated when the parent component resizes.

You can see a demo here: https://codesandbox.io/s/mzz0p2pr0y

The solution could be using a resize observer with a polyfill like https://github.com/que-etc/resize-observer-polyfill, or having some way to flag the calendar that it needs to update the layout.

IndProgo:
The problem was that in month view _"+n more"_ button didn't recalculate, when calendar's size changed, only on date navigation.
Updated that it recalculates using ResizeObserver
Fixes #828
Also removed dummy rows for measurements. Using only real rows if one exist, so eventWrapper with a different event height should calculate _"more"_ correctly, though assuming that the _"more"_ is the same height as event rows


cutterbl:
Think you're on the right track, but seeing a big difference right now, just from the landing page of the documentation
Before
Before
After
After

IndProgo:
Updated that finds first week with an event. Also disables oberve when showAllEvents is enabled

Issue #2581, opened 2024-04-26
Add Japanese Language Support For Localization


ikrammt:
### Clear and concise description of the problem

This issue proposes adding support for the Japanese language (ja) to the current language translations.

The provided code currently supports English (en), British English (en-GB), Spanish (es), French (fr), and Arabic (ar-AE). This update would allow users to display the interface and labels in Japanese.

### Suggested solution

- Include require('globalize/lib/cultures/globalize.culture.ja'); to import the Japanese culture library.

- Update the cultures array to include "ja".

- Add a new key ja to the lang object.

- Define Japanese translations for the  strings within the ja key

### Validations

- [X] Read the [docs](http://jquense.github.io/react-big-calendar/examples/index.html).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Would you like to open a PR for this feature?

- [X] I'm willing to open a PR

PR #2583, opened 2024-04-26
feat: add japanese language support

ikrammt:
This pull request introduces support for the Japanese language (ja) #2581 in the react-big-calendar library. This enhancement expands the user base by enabling users who prefer Japanese to interact with the calendar interface and labels in their native language.


cutterbl:
I appreciate your general thought process here. But, Big Calendar, on it's own, does not provide any speciific language localization support, but rather the tools and props necessary to allow a developer to localize as necessary. While you've improved this story a bit, showing how to do this in Japanese, perhaps a better option may be in providing a more in depth story in the 'Guides' section, detailing the ins and outs of properly providing localization support.

ikrammt:
Thanks for the feedback! You're absolutely right, Big Calendar itself doesn't handle localization directly.  However, showcasing a basic example like the Japanese one is a good start.I will definitely add it to the 'Guides' section.

PR #2600, opened 2024-05-11
fix: use LLL instead of MMM to format month in luxon localizer

### LINKED ISSUE #2599: Luxon localizer should use LLL to format month instead of MMM

- [X] I confirm

### Reproduction link

https://codesandbox.io/p/sandbox/friendly-margulis-n7mfrv?file=%2Fsrc%2Findex.js

### Bug description

Luxon localizer use MMM to format month, this might work fine with language like English, but it doesn't work with all the languages, like Chinese
the correct way is to use LLL 

### Expected Behavior

_No response_

### Actual Behavior

_No response_

1.12.1

### React version

18.2

any browser

### Validations

rinick:
Month should be formatted with LLL in Luxon instead of MMM
MMM might work with some of the languages, but not all language

fix #2599

### DIFF
--- a/src/localizers/luxon.js
+++ b/src/localizers/luxon.js
@@ -5,10 +5,10 @@ function pluralizeUnit(unit) {
 }
 
 const weekRangeFormat = ({ start, end }, culture, local) =>
-  local.format(start, 'MMMM dd', culture) +
+  local.format(start, 'LLLL dd', culture) +
   ' – ' +
   // updated to use this localizer 'eq()' method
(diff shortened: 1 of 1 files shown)


cutterbl:
This is why we have the format props, to allow a developer to override the defaults. Our 'next' version of Big Calendar will use the Intl.DateTimeFormat for formatters, which should be much more accurate.

cutterbl:
@rinick You were right, and I apologize for being so hasty. Please resubmit your branch with updated code from master and I will more thoroughly review.

PR #2609, opened 2024-06-13
fix: render resources under each day in header column

fs-nathan:

### DIFF
--- a/src/TimeGrid.js
+++ b/src/TimeGrid.js
@@ -139,46 +139,50 @@ export default class TimeGrid extends Component {
     const groupedEvents = resources.groupEvents(events)
     const groupedBackgroundEvents = resources.groupEvents(backgroundEvents)
 
-    return resources.map(([id, resource], i) =>
-      range.map((date, jj) => {
-        let daysEvents = (groupedEvents.get(id) || []).filter((event) =>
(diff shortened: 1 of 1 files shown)


cutterbl:
While this initially appears really good, it completely changes the current default functionality, making it a breaking change. This would be better served by a new prop, with refactor to use as an alternative view, providing both the default (current) and new.

almeidapaulooliveira:
To make it simple and be able to merge, what if we have a prop (exposed in `Calendar`) to render the week on default and this? Will avoid the work of refactoring to an alternative view and quickly release the new feature. 

Wdyt?

cutterbl:
@almeidapaulooliveira It's not about having the week on default (that's already a given). It's the difference between grouping days under resources (current display) as opposed to resources under days. I agree that a top level prop (`resourceGroupingLayout`?) allowing to switch is good, but that would require conditional logic in TimeGrid for both the grouping logic, and application of flexbox wrapper components.

almeidapaulooliveira:
@cutterbl I created a new [PR](https://github.com/jquense/react-big-calendar/pull/2652), that implements what we discussed. Looking forward for your review.

cutterbl:
This has been addressed in a new PR with a new prop

PR #2616, opened 2024-07-06
fix(dnd): add correct dragOverFromOutside behavior in week view

### LINKED ISSUE #2383: dragFromOutsideItem does not work in week view

- [X] I confirm

### Reproduction link

http://jquense.github.io/react-big-calendar/examples/index.html?path=/docs/addons-drag-and-drop--example-3

### Bug description

When using `dragFromOutsideItem` to give feedback as you drag over the calendar something from outside, the feedback item appears as expected when in month view, but not when in week view. In week view, it appears when dragging over the full-day events area. But it does not appear when dragging over the hourly planning area.

https://user-images.githubusercontent.com/15199/226417975-97067d27-5ccb-4fc4-91a0-055752dfcad3.mp4

### Expected Behavior

I'd expect to get an event-like element appearing in the hourly planning area as I drag over it.

### Actual Behavior

Nothing appears over the hourly planning are in a week or day view as I drag over an event from outside.

1.6.8

### React version

18.2.0

Any browser

### Validations

achrafl0:
Hey,

This is an attempt to fix https://github.com/jquense/react-big-calendar/issues/2383.

The 2 issues I noticed were: 
- The selectors were listening to the wrong event ( `dragOver` instead of `dragOverFromOutside` )
- The old dragOver code called the `handleDrop` which is not exactly true ! 

It's not breaking, as the old "dragOver" event was never emitted, so the code was never called (from my understanding)! 

I'd love to have your feedback, and I'd be happy to add tests or better documentation if deemed necessary ! 

This is how it looks on the storybooks 

https://github.com/jquense/react-big-calendar/assets/63306573/d8fbb4a2-80d9-4424-b344-950a6ec4c163

### DIFF
--- a/src/addons/dragAndDrop/EventContainerWrapper.js
+++ b/src/addons/dragAndDrop/EventContainerWrapper.js
@@ -126,6 +126,18 @@ class EventContainerWrapper extends React.Component {
     })
   }
 
+  handleDragOverFromOutside = (point, bounds) => {
+    const { slotMetrics } = this.props
+
+    const start = slotMetrics.closestSlotFromPoint(
+      { y: point.y, x: point.x },
+      bounds
+    )
+    const end = slotMetrics.nextSlot(start)
+    const event = this.context.draggable.dragFromOutsideItem()
+    this.update(event, slotMetrics.getRange(start, end, false, true))
+  }
+
   updateParentScroll = (parent, node) => {
     setTimeout(() => {
       const draggedEl = qsa(node, '.rbc-addons-dnd-drag-preview')[0]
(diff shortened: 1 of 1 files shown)


cutterbl:
@achrafl0 I have to pull this code down and run it myself, but a quick look at you changes looks good. Give me a day or two to make sure there are no adverse effects.

cutterbl:
@achrafl0 The issue I'm seeing right now occurs within that same Storybook example. There is an item there which you should not be able to drop on the calendar ("Draggable but not for calendar"). In current master it respects that in the week and day views. Attempting to drop on the timegrid does nothing. With your code it actually shows the slot preview and creates an event on drop.

tranphuongthao2405:
Is there an update on this issue? I also need this feature. Thank you very much.

cutterbl:
@achrafl0 If you could please pull in latest changes so that I can review this again

PR #2637, opened 2024-07-31
refactor(a11y): aria role cell is not allowed for button element

dmitrydyachenko:
**Summary:**

According to [a11y](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/cell_role#description) - ARIA role cell is not allowed for a button element in the current context.

**Solution:**

Remove attribute **role="cell"** from a button element for the DateHeader component.

### DIFF
--- a/src/DateHeader.js
+++ b/src/DateHeader.js
@@ -7,12 +7,7 @@ const DateHeader = ({ label, drilldownView, onDrillDown }) => {
   }
 
   return (
-    <button
-      type="button"
-      className="rbc-button-link"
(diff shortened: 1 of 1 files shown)


dmitrydyachenko:
According to [a11y](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/cell_role#description)  - ARIA role cell is not allowed for a button element in the current context.

beckettkev:
Would it be possible to get this accessibility fix pull request merged please?

PR #2679, opened 2024-11-20
Implement Conditional Resource Grouping

anhdd-kuro:
Complete the missing parts of this [PR](https://github.com/jquense/react-big-calendar/pull/2652) and fix the UI width. 
I’m not sure why, but it seems to work well with a 141px header, possibly due to how the browser calculates the flexbox layout

### DIFF
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -1,28 +1,28 @@
+import clsx from 'clsx'
 import PropTypes from 'prop-types'
 import React from 'react'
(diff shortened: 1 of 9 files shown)


cutterbl:
@anhdd-kuro This all looks really good. Only thing I suggest is including more events at the beginning of the week in the prop story so that it's more obvious.

anhdd-kuro:
@cutterbl 
I've added more events and even add dnd to make sure it worked
please take a look

PR #2681, opened 2024-11-21
fix #2680 - Customize drag from outside event's length

### LINKED ISSUE #2680: Customize drag from outside event's length

When dragging an event from outsize of Calendar, the default of the size is 1 slot and having no way to customize this.
Example, for an event that we already know it takes 4 hours, the preview size could be 4 hours and movable around.

### Suggested solution

The size of event when dragged from outsize can be customized.

### Alternative

_No response_

### Additional context

_No response_

### Validations

ducfilan:
When dragging an event from outsize of Calendar, the default of the size is 1 slot and having no way to customize this.
Example, for an event that we already know it takes 4 hours, the preview size could be 4 hours and movable around.

This PR was created to fix #2680

### DIFF
--- a/src/addons/dragAndDrop/EventContainerWrapper.js
+++ b/src/addons/dragAndDrop/EventContainerWrapper.js
@@ -118,9 +118,11 @@ class EventContainerWrapper extends React.Component {
       boundaryBox
     )
 
+    const end = this._calculateDnDEnd(start)
+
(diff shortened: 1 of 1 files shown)


cutterbl:
@ducfilan See the Details of the unit testing

ducfilan:
Fixed @cutterbl
Thank you!

Issue #2423, opened 2023-07-19
Problem with scroll when resizing event


endeavourC:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

https://jquense.github.io/react-big-calendar/examples/index.html?path=/docs/addons-drag-and-drop--example-2

### Bug description

As we can see in example we have problem with scroll, there are some "lags" and "shifting"/"jumps" (hard to explain) when we want to resize event to top or bottom. And due to this fact, we cannot smoothly with good experience resize an event.

### react-big-calendar version

1.8.1

### React version

18

### Platform/Target and Browser Versions

Google Chrome

### Validations

- [X] Read the [docs](http://jquense.github.io/react-big-calendar/examples).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a react-big-calendar issue and not an implementation issue

### Would you like to open a PR for this bug?

- [X] I'm willing to open a PR

winterberryice:
hey @cutterbl I have made some git bisecting to investigate this issue, here is what I find out:

last working version for this problem was v1.6.4
commit that introduced this issue was d1c5085b004bb3c606a682b488a92585e50b12b4 #2230 

issue only happens when event is larger than calendar height (or user tries to resize event to make it bigger than calendar height) 

I'm not familiar wit codebase but problem is in updateParentScroll function because it sets top edge of event to top of calendar view-port 

There is also a duplicate issue filed for this bug #2437 

hope this info will help to fix this problem

benasher44:
@cutterbl @AmitRDesai #2230 really should have been behind the enableAutoScroll flag. The resulting implementation is somewhat un-intuitive, seeing as the scroll to top is very sensitive. Most controls increase scroll to top/bottom speed as you drag closer to the edge, whereas the scroll speed here is static (unsure there is a way to realistically/correctly implement this in pure JS).

I'm having a bit of trouble threading the enableAutoScroll prop through from the Calendar component to this one, but it would be much appreciated if there was a way to disable this behavior. As it's currently implemented, the sensitivity/jumpiness means that the predictability is preferred over having the auto-scroll behavior zoom your even to the top (or worse, quickly adjusting the event's size during the scroll to the top).

For now, I'm relying on yarn patch to comment out the previous behavior.

PR #2691, opened 2024-12-04
Allow disabling scroll when dragging behaviour

fcheung:
This PR allows disabling the autoscroll introduced by #2230, by adding a new prop (which defaults to true for backwards compatibility)

As noted in #2423 this can result in somewhat jumpy scrolling. It also causes issues with some layouts. for example if I load the sandbox at https://codesandbox.io/p/sandbox/react-big-calendar-example-forked-djgjc9, scroll down to the bottom of the page & resize an event then as soon as I start resizing the page jumps to the top

### DIFF
--- a/src/addons/dragAndDrop/EventContainerWrapper.js
+++ b/src/addons/dragAndDrop/EventContainerWrapper.js
@@ -155,6 +155,10 @@ class EventContainerWrapper extends React.Component {
   }
 
   updateParentScroll = (parent, node) => {
+    if(!this.context.draggable.autoscrollWhenDragging){
+      return
+    }
+
     setTimeout(() => {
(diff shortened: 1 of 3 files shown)


chudson121:
any timelines as to when this will be merged in?

tjhart:
I'm interested in this PR. I have a use case that has events in the day and week view that are as long as the display window. Scrolling is unusable when dragging these events.

jaggman:
As workaround you can use version before this bug `1.4.2`
So it's  works for me
Btw don't forget disable regular autoscroll `enableAutoScroll={false}`
PS.: it'll be better to fix incorrect behaviour of `updateParentScroll` #2230

PR #2693, opened 2024-12-11
adds a new event id accessor that is used as a react key, falls bak to default

milkman4:

### DIFF
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -898,6 +898,8 @@ class Calendar extends React.Component {
     resourceIdAccessor: 'id',
     resourceTitleAccessor: 'title',
 
+    eventIdAccessor: 'id',
+
     longPressThreshold: 250,
     getNow: () => new Date(),
     dayLayoutAlgorithm: 'overlap',
@@ -923,6 +925,7 @@ class Calendar extends React.Component {
     resourceAccessor,
     resourceIdAccessor,
     resourceTitleAccessor,
+    eventIdAccessor,
     eventPropGetter,
     backgroundEventPropGetter,
     slotPropGetter,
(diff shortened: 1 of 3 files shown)


src/DayColumn.js
```diff
@@ -210,6 +210,7 @@ class DayColumn extends React.Component {
     return styledEvents.map(({ event, style }, idx) => {
       let end = accessors.end(event)
       let start = accessors.start(event)
+      let key = accessors.eventId(event) || 'evt_' + idx
```
cutterbl, line 213: Think, to be on the safe side, you should use `??` here, rather than `||`

stories/props/eventIdAccessor.mdx
```diff
@@ -0,0 +1,10 @@
+import { Canvas, Story } from '@storybook/addon-docs'
+
+# eventIdAccessor
+
+- type: `string | function (event: Object) => string`
+- default: 'id'
+
+The id of the event. Must resolve to a string. Used as the key for the event in the DOM. If not provided, the event will be given a key of 'evt\_<index>'.
```
cutterbl, line 8: Must it be a string? Many use a numeric as their `id`.

cutterbl (changes requested):
Few bits for you to look at, but otherwise looks good.

src/DayColumn.js
milkman4: fixed! thanks cutter

stories/props/eventIdAccessor.mdx
milkman4: good call, thank you

src/Calendar.js
```diff
@@ -898,6 +898,8 @@ class Calendar extends React.Component {
     resourceIdAccessor: 'id',
     resourceTitleAccessor: 'title',
 
+    eventIdAccessor: 'id',
```
cutterbl, line 901: One last bit is to add `eventIdAccessor` to the `propTypes` at the top of the file.

cutterbl (changes requested):
One last bit that I missed before.

src/Calendar.js
milkman4: done! thanks

Issue #2720, opened 2025-02-12
Calendar Navigation and View Controls Not Functioning


NikhilBhagoria:
### Check that this is really a bug

- [x] I confirm

### Reproduction link

https://jquense.github.io/react-big-calendar/examples/index.html?path=/docs/additional-examples-event-durations--daylight-savings-starts

### Bug description

**Description**
Both the navigation buttons (Today/Back/Next) and view controls (Month/Week/Day/Agenda) in the Calendar component are not working. Users cannot navigate through dates or switch between different calendar views.

### Expected Behavior

Navigation buttons:
- Today button: Should navigate to current date
- Back button: Should navigate to previous period based on current view
- Next button: Should navigate to next period based on current view

View controls:
- Month button: Should switch to month view
- Week button: Should switch to week view
- Day button: Should switch to day view
- Agenda button: Should switch to agenda view

### Actual Behavior

Navigation buttons:
- Today button: No response when clicked
- Back button: No response when clicked
- Next button: No response when clicked

View controls:
- Month button: No response when clicked
- Week button: No response when clicked
- Day button: No response when clicked
- Agenda button: No response when clicked

### react-big-calendar version

1.17.1

### React version

19.0.0

### Platform/Target and Browser Versions

Chrome

### Validations

- [x] Read the [docs](http://jquense.github.io/react-big-calendar/examples).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [x] Make sure this is a react-big-calendar issue and not an implementation issue

### Would you like to open a PR for this bug?

- [x] I'm willing to open a PR

ardenc:
I was unable to reproduce this on any of the calendars on your reproduction page.

Does this happen to you for every single calendar on the example site?
If so just to rule it out perhaps test it with a different browser you don't typically use.
I see you are using Chrome which certainly works but perhaps you have something setup that is interfering. Try it with Edge or Firefox just to rule that out.

Waleed-abdullah:
Wasn't working for me in my local setup as well, even though i had set it up according to the docs and examples,
Had to manage the state myself using the `onView` and `view` props then it started working

ericvanular:
I'm also experiencing this in React 19. Console error: 

> Your app (or one of its dependencies) is using an outdated JSX transform.

ardenc:
@NikhilBhagoria ah,
Try the new release for RBC that just got pushed out 2 days ago.
It mentions supporting React 19.

[...]

PR #2738, opened 2025-03-31
fix: updated usage of Uncontrollable to fix issues in react 19

mahn00b:
A solve for #2720.

A number of people have been complaining about issues related to the navigation, specifically in react 19. 

After doing some digging, I realize the issue revolves around the usage of another dependency from @jquense called [@jquense/uncontrollable](https://www.npmjs.com/package/uncontrollable). 

In that version of the library, the helper function designed to wrap `class components` is using`React.forwardRef`. Since it's been deprecated in React 19, I've opted to use the hook implementation. 

This PR is just demonstrative. I suggest we refactor the `Calendar` class component to be a functional component to be able to use the hook implementation of `@jquense/uncontrollable`.

**Temporary Solution:**
- Wrap Calendar in a functional component to utilize the more modern hook implementation included by the library.

**Proposed Solution:**
- Refactor Calendar to a functional component
- Use hook implementation of `@jquense/uncontrollable`

### DIFF
--- a/src/Calendar.js
+++ b/src/Calendar.js
@@ -1,7 +1,7 @@
 import clsx from 'clsx'
 import PropTypes from 'prop-types'
 import React from 'react'
-import { uncontrollable } from 'uncontrollable'
+import { useUncontrolled } from 'uncontrollable'
 import {
   accessor,
   views as componentViews,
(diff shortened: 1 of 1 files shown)


cutterbl:
An initial review of your changes, in Storybook, shows all toolbar navigation now broken

PR #2742, opened 2025-04-11
Documenting the resourceAccessor prop in Storybook

john-osullivan:
The `resourceAccessor` prop is required if your event objects don't use `resourceId`, and while it has a docstring in the source code, it's missing from the Storybook docs.  This PR sets up the Markdoc and component for it, as well as updating the `storyDefault` props to show it varying.  My chosen text was pulled from the existing docstring in Calendar.js, but links to related props like the docs for the other resource accessors do:  

https://github.com/jquense/react-big-calendar/blob/7e6ee161b516ed62bbce5ba6923f53eae8323c37/src/Calendar.js#L239-L249

The `resourceIdAccessor` and `resourceTitleAccessor` still won't allow you to change the `events` and therefore the resource key, so I disabled edits on that param for those stories.  However, the `resourceAccessor` story will let you inspect & adjust both the `events` and `resources` arrays.  Other than that, I followed existing patterns wherever I could.

### DIFF
--- a/stories/props/API.stories.mdx
+++ b/stories/props/API.stories.mdx
@@ -283,6 +283,15 @@ Resource {
 }
 ```
 
+### resourceAccessor
+
+- type: `string | function (resource: Object) => string | number // must be unique`
+- <LinkTo kind="props" story="resource-accessor">
+    Example
+  </LinkTo>
+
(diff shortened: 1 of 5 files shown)


michaelruhl:
Ah yes - excellent find thanks John!

john-osullivan:
Hey @cutterbl , anything else I can do to help get this pushed out to everyone?

PR #2746, opened 2025-05-05
Clean up preview elements after dropping from outside

TirlaP:
**Issue**
When dragging items from outside sources into the calendar and dropping them, ghost preview elements sometimes remain visible after the drop operation completes. This creates a confusing UI where users see duplicate events - both the actual placed event and a lingering preview.

**Solution**
This PR adds a minimal fix by:

1. Adding a _cleanupPreviewElements helper method to the EventContainerWrapper component
2. Calling this method only in the handleDropFromOutside function
3. Directly removing any .rbc-addons-dnd-drag-preview elements from the DOM after an outside drop completes

The changes are intentionally minimal, only addressing the specific case of ghost previews after dropping from outside sources, without modifying any other component behaviors.

**Testing**
Tested with various drag operations from outside sources:

- Dragging items from outside into different time slots
- Performing consecutive drag operations from outside
- Confirming no ghost previews remain after dropping

The fix successfully cleans up preview elements without any side effects on normal drag and drop functionality.

This is the photo with the problem on my app:
Screenshot 2025-05-05 at 16 43 38

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,6 +89,7 @@
     "eslint-plugin-react": "^7.32.0",
     "eslint-plugin-react-hooks": "^4.6.0",
     "eslint-plugin-storybook": "^0.6.10",
+    "get-tsconfig": "^4.10.0",
     "husky": "^8.0.3",
     "is-ci": "^3.0.1",
     "jest": "^29.3.1",
--- a/package.json
+++ b/package.json
@@ -89,7 +89,6 @@
     "eslint-plugin-react": "^7.32.0",
     "eslint-plugin-react-hooks": "^4.6.0",
     "eslint-plugin-storybook": "^0.6.10",
-    "get-tsconfig": "^4.10.0",
     "husky": "^8.0.3",
     "is-ci": "^3.0.1",
     "jest": "^29.3.1",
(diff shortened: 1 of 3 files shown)


TirlaP:
@cutterbl Aloha :D,

When you have time, could you take a look at this PR, please?
It's a bug that it's happening in my app, I tried to hide the`.rbc-addons-dnd-drag-preview` className by `display: none !important;`

But this didn't work so I have created a PR to solve this thing. I have put a thorough description for it, but for any other questions you might have, let me know.

Thanks!

j0nas:
Events that are dropped into the "all day" slot of a day seem to still not be cleared. 

I add the following code to my `handleDropFromOutside` to work around this for now:

```typescript
      // HACK: Workaround for a bug in react-big-calendar where the drag
      // preview element is not always cleaned up after dropping an external event,
      // particularly on all-day slots.
      // We manually remove the preview element after a short delay.
      // Ref: https://github.com/jquense/react-big-calendar/pull/2746
      setTimeout(() => {
        document
          .querySelectorAll('.rbc-addons-dnd-drag-preview')
          ?.forEach((node) => node.remove());
      }, 0);
    },
    [assignmentEvent, openModalForAssignmentEvent],
  );
```

PR #2753, opened 2025-05-31
Optimize month layout dispaying events

TirlaP:
Optimize Month view performance for large event datasets

Request by [ahmetcengz](https://github.com/ahmetcengz): https://github.com/jquense/react-big-calendar/issues/2752

  **Problem**

  Month view had severe performance issues with 1000+ events:
  - Slow initial rendering (3-5s for 1000 events)
  - Full re-renders when clicking "Show more" popups
  - Inefficient event sorting and distribution

  **Solution**

  1. Isolated popup rendering - New IsolatedOverlayManager uses React Portals to prevent Month re-renders when showing popups
  2. Optimized event processing - Added fast preprocessing that distributes events to weeks in O(n) time using hash maps, and separates single-day/multi-day events for efficient
  sorting
  3. Layout caching - Cache expensive DateSlotMetrics calculations to avoid redundant computations

  **Performance Results**

  - 1,000 events: ~0.5-1s (was 3-5s)
  - 10,000 events: ~3-5s (was 30+s)
  - Popup clicks: Instant (no re-renders)

  **Changes**

  - Added src/IsolatedOverlayManager.js for portal-based popups
  - Optimized src/Month.js with event preprocessing and caching
  - Added metrics caching to src/DateContentRow.js
  - Minor cleanups and import reordering

  No breaking changes - all functionalities remain the same.

@cutterbl can you take a look when you have some time? I tried to test everything to make sure there are no breaking changes.

But as performance goes, this one is pretty big one.

And idk if it's worth on updating this library to functional components and another React version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -130,5 +130,6 @@
   "bugs": {
     "url": "https://github.com/jquense/react-big-calendar/issues"
   },
-  "homepage": "https://github.com/jquense/react-big-calendar#readme"
+  "homepage": "https://github.com/jquense/react-big-calendar#readme",
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
 }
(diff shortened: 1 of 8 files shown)


TirlaP:
@ahmetcengz take a look over this

cutterbl:
@TirlaP I like what you've done here. Have you tested as far back as React 17?

PR #2759, opened 2025-06-16
fix(dayjs localizer): reverts previous changes

### LINKED ISSUE #2758: Regression: dayjsLocalizer merge() interprets min/max as UTC, breaking local time handling

- [x] I confirm

### Reproduction link

https://codesandbox.io/p/sandbox/react-big-calendar-example-forked-tlyfwm

### Bug description

I'm assuming it's related to this PR:https://github.com/jquense/react-big-calendar/pull/2712

Since version 1.19.2 there is a time shift for different time zones 
now with `Day` and `Week` view the time starts not from 12AM but shifts depending on the time zone 
unfortunately the use of `min` or `max` prop. does not solve the issue.

The recent change in dayjsLocalizer where merge() uses utc(true) breaks local timezone support:

```
return dayjsLib(mergedDateTime).utc(true).toDate();
```

This forces all merged dates (including min, max, event bounds) into UTC, even though all inputs are Date objects in local time.

In timezones with offset, this causes visible shifts:

min={new Date(2025, 0, 1, 0, 0)} renders as 3:00 AM in +3

Previous version of merge() respected local time and worked correctly. Please consider reverting or adding a toggle.

Image  
  

Image

### Expected Behavior

When using min and max props in react-big-calendar (with dayjsLocalizer), the values passed as Date objects (e.g. new Date(2025, 0, 1, 0, 0)) should be interpreted in the local time zone, so that:

min={new Date(2025, 0, 1, 0, 0)} renders the start of the day at 12:00 AM local time

max={new Date(2025, 0, 1, 23, 59)} renders the end of the day at 11:59 PM local time

### Actual Behavior

After the update that added .utc(true) to the merge() function of dayjsLocalizer, the min and max values are now interpreted as UTC, not local time. This leads to:

min={new Date(2025, 0, 1, 0, 0)} being displayed as 3:00 AM in a GMT+3 time zone

max and all event start/end times also shifting away from expected local positions

This breaks expected rendering in all views that depend on accurate local time (especially day and week views).

1.19.2

### React version

18.3.1

Ubuntu Chrome

### Validations

- [x] Read the [docs](http://jquense.github.io/react-big-calendar/examples).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [x] Make sure this is a react-big-calendar issue and not an implementation issue

cutterbl:
Previous change was casting a date as UTC instead of local time

Closes #2758

### DIFF
--- a/src/localizers/dayjs.js
+++ b/src/localizers/dayjs.js
@@ -192,8 +192,7 @@ export default function (dayjsLib) {
     const tm = dayjs(time).format('HH:mm:ss')
     const dt = dayjs(date).startOf('day').format('MM/DD/YYYY')
     // We do it this way to avoid issues when timezone switching
-    const mergedDateTime = dayjs(`${dt} ${tm}`).toDate()
-    return dayjsLib(mergedDateTime).utc(true).toDate()
+    return dayjs(`${dt} ${tm}`).toDate()
   }
 
   function add(date, adder, unit) {


lesha1201:
@cutterbl It doesn't revert the previous changes. Previously, we were using `dayjsLib` in return:

image

Now, we use `dayjs` which will use `dayjs.tz` if available and produces issues when working with time zones and min/max dates.

MyNameIsTakenOMG:
> @cutterbl It doesn't revert the previous changes. Previously, we were using `dayjsLib` in return:
> 
> image
> Now, we use `dayjs` which will use `dayjs.tz` if available and produces issues when working with time zones and min/max dates.

That is true in my case, where I just tested out v1.19.4 with dayjs localizer. And for the properties: `min` and `max`, it just casted them to UTC times. Thus if we wanna set proper time range for each day, we have to use method `.setUTC*`, or like this `new Date(new Date().setUTCHours(9, 0, 0, 0))`

32-git:
> > @cutterbl It doesn't revert the previous changes. Previously, we were using `dayjsLib` in return:
> > image
> > Now, we use `dayjs` which will use `dayjs.tz` if available and produces issues when working with time zones and min/max dates.
> 
> That is true in my case, where I just tested out v1.19.4 with dayjs localizer. And for the properties: `min` and `max`, it just casted them to UTC times. Thus if we wanna set proper time range for each day, we have to use method `.setUTC*`, or like this `new Date(new Date().setUTCHours(9, 0, 0, 0))`

I'm still seeing this bug. Everything is up to date for me (v1.19.4), but the issue with the `min`/`max` props persists. Ideally, the workaround of using `setUTCHours()` shouldn't be necessary.

Issue #2785, opened 2025-11-20
Using outdated JSX transform causes runtime warning in React 18/19


alessandro-arg:
### Issue Summary

When importing `react-big-calendar` in a Next.js + React 19 project that uses the modern JSX runtime (`"jsx": "react-jsx"`), React logs the following warning in the browser console:

> Your app (or one of its dependencies) is using an outdated JSX transform. Update to the modern JSX transform for faster performance: https://react.dev/link/new-jsx-transform

This warning only appears after installing and importing `react-big-calendar`, so it seems the package is shipped using the classic JSX transform instead of the modern automatic JSX runtime introduced in React 17.

### Environment

- React: 19.2.0
- Next.js: 16.0.1
- react-big-calendar: 1.19.4
- TypeScript: 5.x

Excerpt from `tsconfig.json`:

```json
{
  "compilerOptions": {
    "jsx": "react-jsx"
  }
}

AdrianRodU:
Same

KarimKabbara00:
Any update on this?

skrtz:
Hey! I looked into the warning issue as I was getting the same thing in my browser console and I have a fix ready. I can open a PR with the changes if that works for you.

alessandro-arg:
> Hey! I looked into the warning issue as I was getting the same thing in my browser console and I have a fix ready. I can open a PR with the changes if that works for you.

sure, go ahead

mlfahadalsadoon:
can someone review the pr and approve it

PR #2796, opened 2026-04-03
fix: outdated jsx transform

skrtz:
solution for #2785 

I updated babel to use automatic JSX transform, fixing the outdated transforms within the source code. The remaining deprecation warning comes from `react-overlays@^5.2.1` and `uncontrollable@^7.2.1`, which still use the classic JSX transform and should be updated separately.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -37,6 +37,12 @@ module.exports = function (api) {
           absoluteRuntime: false,
         },
       ],
+      [
+        '@babel/preset-react',
+        {
+          runtime: 'automatic',
+        },
+      ],
     ],
     plugins: [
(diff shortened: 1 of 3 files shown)


mlfahadalsadoon:
hi guys can you please approve this PR it is stopping our react upgrade.

huyct96:
Hi, could you please take a look at this PR and approve it if everything looks good? It’s currently holding up our React upgrade, and we’re trying to unblock the migration work. The issue seems to come from dependencies still using the classic JSX transform, so merging this would let us move ahead cleanly with the upgrade path. Thanks in advance for your time and support.

PR #2801, opened 2026-05-20
fix(luxon): use standalone month tokens (LLL/LLLL) instead of MMM/MMMM

### LINKED ISSUE #2599: Luxon localizer should use LLL to format month instead of MMM

mixelburg:
Fixes #2599

The Luxon format tokens `MMM`/`MMMM` produce contextual month names, which are incorrect in some locales (e.g. Russian genitive "января" instead of "Январь", or broken output for Chinese). The standalone tokens `LLL`/`LLLL` are the correct form for calendar headers, where the month name appears on its own rather than as part of a full date phrase.

Changed `monthHeaderFormat`, `dayHeaderFormat`, `agendaDateFormat`, and `weekRangeFormat` to use `LLLL`/`LLL` accordingly.

### DIFF
--- a/src/localizers/luxon.js
+++ b/src/localizers/luxon.js
@@ -5,10 +5,10 @@ function pluralizeUnit(unit) {
 }
 
 const weekRangeFormat = ({ start, end }, culture, local) =>
-  local.format(start, 'MMMM dd', culture) +
+  local.format(start, 'LLLL dd', culture) +
   ' – ' +
   // updated to use this localizer 'eq()' method
-  local.format(end, local.eq(start, end, 'month') ? 'dd' : 'MMMM dd', culture)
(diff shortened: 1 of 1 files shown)


mixelburg:
Closing due to inactivity — over 40 days with no engagement. Happy to reopen if there's interest.

PR #2808, opened 2026-06-23
fix: guard dragFromOutsideItem in _calculateDnDEnd when not provided

### LINKED ISSUE #2780: dragFromOutsideItem issue when props is undefined

- [x] I confirm

### Reproduction link

https://codesandbox.io/p/devbox/react-big-calendar-example-forked-tpnyq8

### Bug description

When using the latest version of react-big-calendar, the prop dragFromOutsideItem seems to be expected even if it’s not defined.
If I don’t provide it, I get the following error:
`TypeError: _this.context.draggable.dragFromOutsideItem is not a function`

### Expected Behavior

if dragFromOutsideItem is not defined, the calendar should handle it gracefully instead of throwing an exception.

### Actual Behavior

When dragFromOutsideItem is not defined, after dropping an event, an error is thrown

1.19.4

### React version

19.0.0

Chrome

### Validations

- [x] Read the [docs](http://jquense.github.io/react-big-calendar/examples).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [x] Make sure this is a react-big-calendar issue and not an implementation issue

- [x] I'm willing to open a PR

mixelburg:
Fixes #2780

When using `onDropFromOutside` without providing `dragFromOutsideItem`, dropping an external item crashes with `TypeError: dragFromOutsideItem is not a function`.

This happens because `_calculateDnDEnd` unconditionally calls `this.context.draggable.dragFromOutsideItem()`. The `dragOverFromOutside` handler already guards against this (line 235 in EventContainerWrapper), but the `dropFromOutside` -> `handleDropFromOutside` -> `_calculateDnDEnd` path does not.

Fix: early-return `slotMetrics.nextSlot(start)` when `dragFromOutsideItem` is not provided, matching the guard pattern already used in WeekWrapper.js.

### DIFF
--- a/src/addons/dragAndDrop/EventContainerWrapper.js
+++ b/src/addons/dragAndDrop/EventContainerWrapper.js
@@ -145,6 +145,9 @@ class EventContainerWrapper extends React.Component {
 
   _calculateDnDEnd = (start) => {
     const { accessors, slotMetrics, localizer } = this.props
+    if (!this.context.draggable.dragFromOutsideItem) {
+      return slotMetrics.nextSlot(start)
+    }
     const event = this.context.draggable.dragFromOutsideItem()
     const { duration: eventDuration } = eventTimes(event, accessors, localizer)
 


mixelburg:
Closing due to inactivity — over a month with no review or feedback. Happy to reopen if there's interest.

